PR #13, opened 2017-12-07
update docker image

markuman:
update docker image. 
using now default alpine 3.7 image (comes with nodejs 8.9.1). It's 15 MB smaller now.


o1lab:
Thank you for this. 

Will it be a good thing to get docker as part of travis - I haven't dealt with docker and travis before.

markuman:
... and I'm not familar with travis too. :D But theoretically, it's just a `docker build -t xmysql .` build the image.

PR #16, opened 2017-12-12
Fixing pagination issue when using _p & _size

soaserele:
When using _p ang _size query params in URL, the first row is skipped.

Example:
_p = 1 & _size = 20
reqParams._index will be = (1 - 1) * 20 + 1 = 1; 
Generated SQL will be SELECT * FROM ? LIMIT 1,20. This SQL skips first element in the table
Correct SQL that should be generated is SELECT * FROM ? LIMIT 0,20

### DIFF
--- a/lib/xsql.js
+++ b/lib/xsql.js
@@ -197,7 +197,7 @@ class Xsql {
     }
 
     if ('_p' in reqParams && parseInt(reqParams._p) > 0) {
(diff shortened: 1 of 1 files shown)


o1lab:
@soaserele : thank you for finding this issue and submitting pr. I will take a look after xmas as Im currently away.

PR #31, opened 2018-04-29
add mysql port config

chocholand:
you can change mysql port ( 3306 by default )


o1lab:
Looks good for merge. Thank you @chocholand

PR #39, opened 2018-07-13
Bug fix for segfault

extremeshok:


o1lab:
@extremeshok : Thank you for pull request. I do not know much about these flags. Please leave a comment on what scenario was causing the issue (if any in particular). And how this helps.  Thank you.

PR #47, opened 2018-08-30
Call stored procedures

zbynekrybicka:
Hi. Here is solution of call procedures (no functions).
I will be pleasured for any feedback. Especially in safety.


o1lab:
@zbynekrybicka : thank you.

Could you please write some tests under [test](https://github.com/o1lab/xmysql/tree/master/tests) directory. 

If there are no stored procedures existing in sample.sql - create couple of them in sample.sql.

o1lab:
hey @zbynekrybicka - I see you have added tests. I think loading sql changes before test is making it fail. Do you want to have a relook into it ?

zbynekrybicka:
Sorry, I lost my source db and I can't connect to testing db.
You need only create procedure in SQL and call POST /_proc/*procname* and params are in body.

zbynekrybicka:
I apologize for delay. It's complete.

PR #48, opened 2018-08-31
Fixed version bug

pdme:
The current implementation is broken - the `_version` endpoint returns the version of the package `pkginfo` (0.4.1) instead of that of this package (0.4.9).

(Calling `const pkginfo = require('pkginfo')(module);` exposes the info on `module.exports`, not on the variable `pkginfo`.)


o1lab:
Thank you.

PR #55, opened 2018-09-15
Add docker compose

bvkatwijk:
Notes:
- I have altered the test script to pick up environment variables
- I have altered the entrypoint script to execute the incoming command
- The original `setup.sql` gave `Query limit exceeded` errors, so I've added `docker-setup.sql` which breaks the insert statements into smaller chunks.
- Added `dev.Dockerfile` which includes dev dependencies, I had to add python to the image
- I set port `3002` to prevent conflicts with current docker setup
- Initial runs set up a database, runs after it use the existing database. If the data should be cleared, `docker-compose down -v` can be used. This can be unclear for non-compose users.
- There are many more available options available (`watch` mode for dev/testing, ci setup, etc), I've kept it basic for now. I've also not altered the existing docker(file) setup. EDIT: I have altered entrypoint script. Please check whether this is undesirable.


o1lab:
Thank you @bvkatwijk for the detailed and well made notes with PR. I understand less about docker-compose and k8s - learning them now.

Can you please mention why python is included and how changing entrypoint changes  behaviour (if any).

bvkatwijk:
While installing dev dependencies of npm I got an error that python is required. Hence the python install. Same for `g++` and `make` packages.

I recommend the docker docs about entrypoint and command mechanics, but loosely in practice `entrypoint` is what should always happen (e.g. waiting for database), and `cmd` is a default action (starting the api generation, or running tests, or maybe other actions). Hence I've removed the api generation command from the entrypoint script, and it can be added to the `dockerfile` (same as in `dev.Dockerfile`).

PR #58, opened 2018-09-19
Add initial /swagger.json

bvkatwijk:
Done:
- Added [host:port]/swagger.json endpoint which can be targeted by a swagger-ui
- Added generated documentation per resource for GetAll, GetSingle, PostSingle and PutSingle
- Added entity definition for error class
- Added `db-api-dev` service (hot reload while coding) to the docker-compose setup
- Added `swagger-ui` service to docker-compose setup to validate the generated swagger docs

Not done:
- Entity definition (fields and types) per database resource. It was not clear to me where these were processed during initialization, or are they not resolved until the api call (`/describe`) is invoked?
- All other available endpoints

It's only a start and far from perfect, but it should be enough to be easily extendable for contributors following the pattern. Maybe the docs should include a note that they are far from complete currently?
Hope it helps!


wiwengweng:
Hi, is it only available on docker image? I cannot find this /swagger.json url on local deployment, which causes 404 not find when I try to import this url to my swagger-editor.

VanDerLars:
any progress here?

VanDerLars:
any progress here?

Issue #60, opened 2018-10-03
Is there a way to generate API only on specified tables ? We do have ignoreTables option.


efrenette11:
Please provide dev environment versions of your system

node -v
( xmysql requires node >= 7.6.0 )

npm -v
mysql --version
xmysql --version

- - -

Provide steps to reproduce this issue
or
Provide minimum code to reproduce this issue
or
HTTP request,response,error snapshot of this issue

- - -

Mark issue with suitable label if it is clear what this issue is.

o1lab:
185 seconds - no good! 

Could you please mention about the schema like how many tables in it and versions of mysql, os, node etc.

Are you using mysql from docker? in some cases docker mysql takes a little more time but 3+ minutes is new. 

Generating APIs only for tables required will be featured going ahead.

Bnaya:
I'm also interested in this feature. it be able to expose only small part of my tables.
Using blacklist just force me to create long list, that i need to keep upto date, or write code that do that

PR #61, opened 2018-10-25
 Configuration file - ignored tables

zbynekrybicka:
Hi. I resolved the problem with ignored tables. #60


o1lab:
@zbynekrybicka : Currently, tables can be ignored by providing a command line param (-i). Does the PR allow user to generate apis for interested tables only ? How does the command line usage look like ?

zbynekrybicka:
@o1lab xmysql -u root -h localhost -p password -d database
No more. This configuration file has available only one param "ignoredTables" but in the future will have more.

PR #75, opened 2019-03-28
eslint airbnb rules

v2io:
All JS files are are linted according to AirBnB Standards, This is start.
Renamed .eslintrc to ".eslintrc.js"


o1lab:
Thank you

PR #104, opened 2020-03-06
fix(xapi): remove green.bold and update mysql packages for deprecatio…

ans-4175:
- I removed green.bold on lib/xapi line 366 because it is undefined
- I also update mysql packages so that will remove deprecation warning


ans-4175:
fixing for issue https://github.com/o1lab/xmysql/issues/103

PR #130, opened 2021-03-19
Added --dynamic flag to allow dynamic queries

PeWu:


felipe2santoscom:
Thank you! Good solutions!

Thank you, good solution. But I want to leave my contribution and how I resolved it.

My problem:
I needed to publicly expose this on the internet with a security / authorization rule

Solution:
It didn't work on a docker at all, due to the release of port 3000 and a problem with the node's express service. I changed the strategy, uploaded a Node.js server (version 14.16.0 LTS) and installed xmysql and changed the xapi.js file according to print image

To expose the service on the internet, I initially used ngrok, but with the session expires in 2 hours, I started using the cloudflare argo tunnel https://developers.cloudflare.com/cloudflare-one/connections/connect-apps

Then I added a rule in the Cloudflare Firewall to block and release access (http.host eq "domain.com.br" and not any (http.request.headers ["myheader"] [*] == "hashcode"))

Grateful :)

PR #217, opened 2021-06-01
Switch from Calendly to Calendso

josephjacks:
Propose to switch from Calendly to Calendso.

https://calendso.com/NocoDB is still available. 👍

### DIFF
--- a/packages/nc-gui/layouts/default.vue
+++ b/packages/nc-gui/layouts/default.vue
@@ -47,7 +47,7 @@
               </x-icon>
             </a>
 
-            <a href="https://calendly.com/xgenecloud/demo" target="_blank">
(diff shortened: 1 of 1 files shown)


packages/nc-gui/layouts/default.vue
```diff
@@ -47,15 +47,15 @@
               </x-icon>
             </a>
 
-            <a href="https://calendly.com/xgenecloud/demo" target="_blank">
+            <a href="https://calendso.com/NocoDB" target="_blank">
```
gautamkrishnar, line 50: This url gives 404

gautamkrishnar (changes requested):

o1lab:
Will revisit this. Have shown issues to Peer.

PR #239, opened 2021-06-06
added scaleway plugin

bhanuc:
This adds scaleway object storage as an option.


o1lab:
Signing the git commit part is missing, please verify

bhanuc:
Opened other pr with correct sign

Issue #218, opened 2021-06-01
Bug/Feature Request


bhanuc:
Really great project, though while entering date currently no date selector is coming up. Is it a bug or something that hasn't been planned ?

o1lab:
Which database & version was this ?

bhanuc:
postgres and using the latest docker image- nocodb/nocodb:latest

pranavxc:
Please share the datatype of the field

bhanuc:
DataType is Date.

PR #240, opened 2021-06-06
added fix for date dialog for postgres

bhanuc:
This fixes issue -  https://github.com/nocodb/nocodb/issues/218


o1lab:
Signing the git commit part is missing, please verify

PR #247, opened 2021-06-07
feat: add dco check in Pull Request

wingkwong:
Add DCO check in PR - Inspired by https://github.com/nocodb/nocodb/pull/239#issuecomment-855371462


o1lab:
Thank you for thinking this through.

PR #353, opened 2021-07-16
NC_DATABASE_URL_FILE patch

janrito:
docker compose ECS integration only passes secrets via runtime available files. This allows for nocodb to be provided with a database configuration via a secret without too much hassle


pranavxc (changes requested):
Please sign the commits

For more information refer  : https://github.com/nocodb/nocodb/blob/master/contribute/HowToApplyLicense.md

pranavxc:
looks like the email provided is wrong

janrito:
> looks like the email provided is wrong

sorry, I've corrected it

PR #463, opened 2021-08-16
Expanded form fix for text overflow

AztrexDX:
There is text overflow issue that when text length is large then causes bug that button moves up and it doesn't looks good overall. So put a text limit using css.

Signed-off-by: Ayush Sahu <aztrexdx@gmail.com>
after changes
title header breaks when too much text


pranavxc:
@all-contributors please add @AztrexDX for code

PR #505, opened 2021-08-26
refactor: add numeric type to more pg column types

ejose19:


pranavxc:
@all-contributors please add @ejose19  for code

o1lab:
Was about to say ```numeric``` has to be handled in ModelXcMetaPg.ts. However, we already 've it there. All good :)

PR #561, opened 2021-09-23
Example docker-compose.yml for use with nginx-proxy-manager and 2 seperate networks

ArjenR:
Here my example for using nginx-proxy-manager for exposing nocodb to the outside. 
It creates 2 networks:
One for the nocodedb app container and mysql database. Thus mysql is not exposed beyond this network as it is with the traefik example. And one for nginx-proxy-manager and the nocodb app container.


o1lab:
Thank you for the quick PR

o1lab:
@all-contributors please add @ArjenR for code

PR #574, opened 2021-09-25
Add ignoreTLS to email plugin

eevleevs:
Happened to work with a server that needs `port:587, secure:false, ignoreTLS:true` to work. 
Just addded the new field, did not test. Please check.


eevleevs:
Will resend with signatures

PR #672, opened 2021-10-18
Update de.json

dahawk:
https://docs.google.com/spreadsheets/d/1CgkoYkpivme6JXYlmdB8wkyFJQ9NmtpjRRPfzO7cNjg/edit?usp=sharing

btw, great system for updating i18n files while keeping them nicely uniform.
There is no working translation for Star us on Github. Best bet would be to keep the en text.

BR Christof

### DIFF
--- a/packages/nc-gui/lang/de.json
+++ b/packages/nc-gui/lang/de.json
@@ -1,24 +1,24 @@
 {
   "projects": {
     "my_projects": "Meine Projekte",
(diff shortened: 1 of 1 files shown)


o1lab:
Fix is in [this](https://github.com/nocodb/nocodb/pull/672/commits/3aee2f15a8a362e6d6d92dec2d179663b9668c26)

Issue #140, opened 2021-05-11
Feature : Support Kanban view


o1lab:
- Meta DB Migration Script
  - two new columns in `nc_kanban_view_v2` 
    - `grp_column_id`: grouping field column id (used in stackedBy)
    - `meta`: meta data like stack order, colors, option titles etc.
- stackedBy in Toolbar
  - dropdown for selecting a Single Select column
  - rerendering kanban view based on the change
- Kanban View
  - Basic View Operations
    - Create View
    - Copy View
    - Rename View  
    - Delete View
  - Stack 
    - Create a new Stack 
    - Re-order Stack
    - Add records within a Stack
      - Preset grouping field value
    - Drag-and-drop record cards to other Stack
    - Scroll to fetch records
    - Dropdown
      - Add new record
      - Collapse Stack
      - Delete Stack
  - Record Card
    - Click to expand form to edit
    - Drag-and-drop records across stacks
  - Share View 
  - Misc
    - Sync grouping field options from other views (e.g. remove options in grid view)
  - Cypress

grepler:
When this is implemented, these are the features I'd love to see:

1) 'swimlanes' option, so that we can separate out columns by a multi-select / single-select field dimension.
   for reference, consider something similar to: https://ej2.syncfusion.com/vue/demos/#/bootstrap5/kanban/swimlane-template.html

2) show configurable field data on the cards
  for additional context, something similar to: https://ej2.syncfusion.com/vue/demos/#/bootstrap5/kanban/card-template.html


3) same card in multiple lanes, columns:
   control + click and drag to split a card and add to a multi-select field (cards with multiple values in a multi-select field should appear in multiple areas).

zacker330:
Hi guys, Is this feature implemented? I can't find out it in the releases list.

TechMasterElvy:
I am confused, I setup heroku and isntalled nocodb and studied the docs only to find no kanban, gantt or calendar is there thought it's shown in demos and on the website!

You could've at least added "Coming Soon"

nilsreichardt:
@o1lab I saw that https://github.com/nocodb/nocodb/pull/903 is merged 🎉  Is there a timeline when we can expect the kanban view feature?

khan95osu:
Agreed. The website shows a calendar and Kanban but it's not available for me

matthewblott:
Lads, this is your homepage!

[...]

PR #903, opened 2022-01-02
feat: kanban view

wingkwong:
Ref: #140 

## Change Summary

- Kanban view
  - Enable Kanban option on Navigation menu
  - Stack title color should match with that of the existing grouping field
  - Horizontal scrolling when there are numbers of stacks
  - Create
    - Render the kanban stack based on the grouping field
  - Edit
    - Edit the view data
    - Update the grouping field
  - Copy view
  - Delete view
  - Rename view
  - Share View Link Generation
  - Infinite Scrolling (10 more items per each reload)
  - Skeleton Loader Component
  
## Change type

- feat: (new feature for the user, not a new feature for build script

## Next Step

- More Cypress tests
- UI/UX tuning
- Share View


zacker330:
awsome feature 👍

patlecat:
When will this come, we really need it?

Issue #1034, opened 2022-02-03
Webhook to send email notification can't be cancelled


WeT-Klb:
**Please enter the following details**

Node: **v12.22.6**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **mysql2**
ProjectOnRootDB: **true**
RootDB: **mysql2**
PackageVersion: **0.84.8**

**Steps To Reproduce**
1. I installed SMTP-App, did the configuration and created a weebhook in table "Arztadressen"
 to send email notification when entries are changed
2. I deleted the webhook 
3. Even everything is empty now i still get notifications for changing entries
4. After uninstalling and reinstalling the SMTP-plugin the notifications are still sent!!!

**Expected behaviour**
After the weebhook is deleted there shouldn´t be any email notifications anymore

**Screenshots**
Creation of webhook:
![Webhook neu](https://user-images.githubusercontent.com/20600342/152325805-88164bca-0261-4f7a-9702-f1b4dc2a5829.jpg)

Canceling webhook:
![Webhook](https://user-images.githubusercontent.com/20600342/152323340-cff8d924-ba64-4002-8297-7c792f9503dd.jpg)

Join our discord : https://discord.gg/5RgZmkW for realtime help.

wingkwong:
Fixed in ``0.84.9``.

PR #1044, opened 2022-02-04
fix: update hook list after deleting webhook

wingkwong:
Ref: #1034 

## Change Summary

Currently the hooks can still be found in ``this.builder.hooks`` after deleting them. This PR is to add one more case ``tableXcHooksDelete`` in NcProjectBuilder listener to trigger ``onHooksUpdate()`` to load the latest hook from ``nc_hooks``.

## Change type

### DIFF
--- a/packages/nocodb/src/lib/noco/NcProjectBuilder.ts
+++ b/packages/nocodb/src/lib/noco/NcProjectBuilder.ts
@@ -304,10 +304,6 @@ export default class NcProjectBuilder {
         break;
 
       case 'tableXcHooksSet':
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/lib/noco/NcProjectBuilder.ts
```diff
@@ -308,6 +308,11 @@ export default class NcProjectBuilder {
         console.log(`Updated validations for table : ${data.req.args.tn}`);
         break;
 
+      case 'tableXcHooksDelete':
```
pranavxc, line 311: You can combine both 'tableXcHooksSet' and 'tableXcHooksDelete'

Issue #1047, opened 2022-02-04
Formulas Not Calculating after latest release


browngd:
I haven't used my NocoDb install for a few months.  After the latest update none of the formulas are performing calculations.  I have checked the formula and it is correct.  It is a simple subtraction formula (Size-Used) that I am using in the Remaining column to calculate the remaining amount.  It accepts the formula, but won't output anything to the cell with the formula.  Without formulas, I will have to move back to Airtable.  I really prefer to self-host to protect our data.  Thanks!

o1lab:
Could you please paste the project info of yours (right top menu)

I tried a simple example and it appears to work.

browngd:
Node: **v12.22.6**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **pg**
ProjectOnRootDB: **true**
RootDB: **pg**
PackageVersion: **0.84.8**

Thanks!

browngd:
If it helps any, I can't get the rollup function to work either.  I have also tried other formula functions such as CONCAT, +, -, * etc. and nothing works.  When I try the rollup function and I go to select the Child Table it says "No data available", but I have a linked record to that table and that works but it won't work for Rollup.  Also, if I sync the Metdata it removes all field options for all fields that are select or multi-select.  The values are still there, but they are shown like single line text and when I edit the field, the values are gone.  Also, I am using a Postgres database in another container for all my projects.  I am linked to my database container using NC_DB env variable.  And lastly, everything is running on a DS916+ Synology NAS device.

o1lab:
I tried the same things and are working.

> CONCAT, +, -, *

Please have a look at the [docs](https://docs.nocodb.com/setup-and-usages/formulas) on the usage. 

Arithmetic operators are on numeric operators only.

Please reach out to me on [discord](https://discord.gg/5RgZmkW)

And for sync: I've raised [this issue](https://github.com/nocodb/nocodb/issues/1049).

pranavxc:
Please check the browser network tab and server log and if anything unusual please share which will help us to resolve the issue.

browngd:
This was a new project on a new install.  Also, both of the fields I am trying to perform operations on are number fields.  That's what led me to try some other operations as well such as CONCAT since I use thqt regularly.  At this time, I haven't found any formula that works.  It's like it checks to see if it is a valid formula and then when I save it, it doesn't run.

browngd:
[...]

Issue #1046, opened 2022-02-04
URL field layout type is temporay lost after Metadata synchronization


scali:
NocoDB used as docker : true
NocoDB version : 0.84.8
Database used in NC_DB URL  : pg
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : pg
OS on which NocoDB is running : Ubuntu
Database version : 11.11

Steps to reproduce :
1. Go to 'Add new table' 
2. Keep to default new table template
3. Change the layout type to 'URL'
4. Add new row with a random URL
5. Go to 'Project Metadata' > 'Metadata' > click on 'Sync Now'
6. Back to the created table in step 1
7. See error : URL field is blank

**Expected behavior**
URL field should display the data

If we go to 'Edit' on the URL field, the layout is working as expected

o1lab:
Changing the column type from singleLineText to URL is leading to errors. Tried on PG.

@scali 
Any reason - step 5 was done ?

scali:
Our workflow is DB-first, so we are doing metadata sync regulary. I mention step 5 here just to reproduce the problem

o1lab:
Fixed in ```0.84.9```

PR #1050, opened 2022-02-05
fix: handle meta sync properly and keep proper column metadata

### LINKED ISSUE #1049: Sync shows schema changes for the changes done via GUI
Changes done via GUI are already synced and it should not be shown.

0.84.8

pranavxc:
closes #1049
re #1047
re #1046


o1lab:
- [ ] Disabling sync button when no changes identified is required.
- [ ] On sync, to see whether virtual column and UI data types are like before has to be confirmed.

PR #1161, opened 2022-02-11
fix postgres bigint mapping

vijayrathore8492:
## Change Summary

Bigint was incorrectly mapped to string for pgUI. This PR fixes it. 

## Change type

## Test/ Verification

Provide summary of changes.


wingkwong:
It is expected to use ``string`` since the numeric field couldn't display ``9223372036854775807`` which is the maximum value for ``bigint`` in Postgres . In fact, other client types would also follow the same approach in recent changes.

vijayrathore8492:
Okay, but then it's breaking the search on bigint column. It uses the `like` operation instead of `eq`, which generates an invalid SQL statement.

Issue #214, opened 2021-06-01
Ability to delete users


joelgombin:
I can add or remove users at a project level, but once a user is created I can't delete him at a global level.
Also, is there documentation as for how long invite links for new users are valid? It seems that after a short while (12 hours?) they don't work anymore.

dismantl:
I also noticed that a user that was removed from a project, and then re-added, will get the Editor role regardless of what role(s) the user had before being removed from the project.

Braincoke:
In the meantime I'm just deleting users directly from the database.
I tried using nocodb to edit nocodb's database without success (worth a try).

You just have to find the ID of your user in the table `xc_users`;

Then `DELETE FROM xc_users WHERE id=2;` and `DELETE FROM nc_projects_users WHERE user_id=2;`.

o1lab:
Super admin now have the rights to delete users.

sebkur:
Thanks for that update. Just wanted to mention because it was not obvious to me. There are multiple places within the app to manage users and not all of them have the ability to *really* delete users. You need to go back to the front page / dashboard as the super admin, then select the three dot menu in the top right corner, click "Account". The user management page that this brings you to has an actions column that contains the powerful management items such as "delete".

PR #1168, opened 2022-02-12
fix: add missing delete user logic

wingkwong:
Ref: #214 

## Change Summary

Currently you can only delete users in project level. If you accidentally invite someone with a wrong email address, there is no way to delete it in UI level. This PR is to cater this case. 

## Change type

## Test/ Verification

Create a new user

image

Remove this user from the current project

image

Remove this user from NocoDB

image


mertmit:
Failed to apply the patch. Make sure that nocohub develop and nocodb develop is in sync and then remove and add  tag.

PR #1415, opened 2022-03-07
feat: DATEADD formula for date arithmetics 

### LINKED ISSUE #1389: Formula should support date arithmetic
```IF(Category=="A lot",ADD(lastcontact,21))```

### PR DESCRIPTION
**Pull request by mertmit**
## Change Summary

Added a new formula called DATEADD which supports date arithmetics.
It takes 3 arguments DATEADD(date, number, 'interval')

- Date can be date or datetime, result changes according to given type
- Interval values depends on database but common ones work for each (day, month etc.)
- Number can be negative for subtraction.

Ref: closes #1389 

## Change type

## Test/ Verification

image
daytodate: DATEADD(date, 1, 'day')
monthtodatetime: DATEADD(datetime, 2, 'month')
isDateAfter: IF(NOW() < date, "true", "false")
isDatePlus10After: IF(NOW() < DATEADD(date,10,'day'), "true", "false")
(Date of execution: 2022-03-07)

DateTime Picker was giving an error on MySQL (timezone part) so I implemented a work around for it. Although I tried this with different versions, this error needs confirming it might be caused by my setup.

MSSQL date picker isn't working too(As much as I can inspect it tries to create a varchar without proper length as date column). I couldn't find a way to fix this but implemented the DATEADD function anyway.

### DIFF
--- a/packages/nc-gui/components/project/spreadsheet/components/editableCell/dateTimePickerCell.vue
+++ b/packages/nc-gui/components/project/spreadsheet/components/editableCell/dateTimePickerCell.vue
@@ -42,7 +42,11 @@ export default {
           .format('YYYY-MM-DD HH:mm')
       },
       set(val) {
-        this.$emit('input', val && dayjs(val).format('YYYY-MM-DD HH:mm:ssZ'))
+        if(this.$parent.sqlUi.name == 'MysqlUi') {
+          this.$emit('input', val && dayjs(val).format('YYYY-MM-DD HH:mm:ss'))
+        } else {
+          this.$emit('input', val && dayjs(val).format('YYYY-MM-DD HH:mm:ssZ'))
+        }
       }
     },
     parentListeners() {
--- a/packages/nc-gui/helpers/formulaList.js
+++ b/packages/nc-gui/helpers/formulaList.js
@@ -55,6 +55,7 @@ const validations = {
   IF: { validation: { args: { min: 2, max: 3 } } },
   SWITCH: { validation: { args: { min: 3 } } },
   URL: { validation: { args: { rqd: 1 } } },
+  DATEADD: { validation: { args: { rqd: 3 } } },
 };
 
 export default Object.keys(validations);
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: packages/nc-gui/components/project/spreadsheet/components/editableCell/dateTimePickerCell.vue [hunk 1]
```diff
@@ -42,7 +42,11 @@ export default {
           .format('YYYY-MM-DD HH:mm')
       },
       set(val) {
-        this.$emit('input', val && dayjs(val).format('YYYY-MM-DD HH:mm:ssZ'))
```
pranavxc, line 45: Can you share more information like in which version it's failed and datatype you had used?

mertmit: mysqlerror
Node: **v16.14.0**
Arch: **x64**
Platform: **win32**
Docker: **false**
Database: **mysql2**
ProjectOnRootDB: **false**
RootDB: **sqlite3**
PackageVersion: **0.84.14**

When I try to pick a datetime column on a MySQL table it gives this error. I didn't want to merge this into a feature but it showed up while I try to test the feature so I ended up with this solution.

Also I tried datetimes with timezones on different versions of MySQL. You can check this fiddle, which gives same error: https://www.db-fiddle.com/f/itVMsnmnr6LpB5mn4UimRp/0

mertmit:
I ran the failed test ('./scripts/cypress/integration/test/restMisc.js') on local with success. I think there was a problem on action (looks like its just a timeout issue on language change which is not even related with the changes). I am sure a rerun will be enough for it.

packages/nc-gui/components/project/spreadsheet/components/editableCell/dateTimePickerCell.vue
o1lab: Thank you for the effort @mertmit . Really appreciate it. Possible to get in touch via n _AT_ nocodb.com

PR #1557, opened 2022-03-30
fix: multiple after-update webhooks being triggered in text

### LINKED ISSUE #1453: multiple after-update webhooks being triggered when changing value in SingleLineText / LongText
When editing these cells, `onCellValueChange` will be triggered from time to time if the current value is different than the previous one to prevent data loss. However, if users configure after-update webhooks, they will be triggered along with `onCellValueChange`.

mertmit:
## Change Summary

Added cell saved detection for firing update hooks to prevent auto-save from firing webhooks.
Ref: fixes #1453 

## Change type


wingkwong (approved):
LGTM

o1lab:
hey @mertmit, thanks for sending this.

However this change leads to breaking changes in the desired behaviour of the APIs. 

If somebody is just using our APIs and not the GUI - they would have to then send an additional key `_cellSaved` in req.body within the API otherwise the webhook will never be triggered. 

So `_cellSaved` needs to go. Instead we can handle this in a reverse way - you can have a variable `_ignoreWebhook`. Only when `_ignoreWebhook` is present we can ignore it. And also the variable name makes sense during simple API invocation without GUI.

PR #1620, opened 2022-04-07
fix: update record counter with insertion/deletion

mertmit:
## Change Summary

Fixes record counter.

## Change type

## Test/ Verification

- [x] update when new row inserted
- [x] update when single row deleted
- [x] update when multiple row deleted


packages/nc-gui/components/project/spreadsheet/rowsXcDataTable.vue
```diff
@@ -1246,6 +1248,8 @@ export default {
           this.expandRow(data.length - 1, rowMeta)
         }
       }
+      await this.save()
```
wingkwong, line 1251: This will break the existing behaviour. 

Supposing users create some empty rows (with no input) and refresh the page, those empty rows will not be saved. However, if you add `await this.save()` here, it would save those empty rows as well which is not expected.

Therefore, you don't need this line. Please delete it.

packages/nc-gui/components/project/spreadsheet/rowsXcDataTable.vue
```diff
@@ -1246,6 +1248,8 @@ export default {
...
+      this.syncCount()
```
wingkwong, line 1252: Based on the previous comment, you don't need to call `syncCount()` in `insertNewRow()` because you should only call it upon a _successful_ insertion, i.e. somewhere in `save()`.

Please delete this line and add it in `save()`.

packages/nocodb/src/lib/noco/meta/api/dataApis/dataAliasApis.ts
```diff
@@ -130,6 +151,10 @@ router.get(
   '/data/:orgs/:projectName/:tableName/views/:viewName',
   ncMetaAclMw(dataList)
 );
+router.get(
+  '/data/:orgs/:projectName/:tableName/views/:viewName/count',
+  ncMetaAclMw(dataCount)
```
wingkwong, line 156: If you pull the latest code, you will see the change. We add the function name manually instead of reading `.name` property. This is because when we build the backend, the function name will be minified which will cause missing permission in ACL module.

Hence, you can simply change `ncMetaAclMw(dataCount)` to `ncMetaAclMw(dataCount, 'dataCount')`.

wingkwong (changes requested):

wingkwong:
The changes are okay. Please pull the latest code and do the following changes.

- We previously hid the row count for time-being. Please undo https://github.com/nocodb/nocodb/commit/97a0c09a1b20ecfa9634b93061d7b09328ad905d. 
- We've also changed the API paths. 
  - `/data/:orgs/:projectName/:tableName/views/:viewName/count` -> `/api/v1/db/data/:orgs/:projectName/:tableName/views/:viewName/count`
  - `/data/{orgs}/{projectName}/{tableAlias}/views/{viewName}/count` -> `/api/v1/db/data/{orgs}/{projectName}/{tableAlias}/views/{viewName}/count`

wingkwong (approved):

PR #1672, opened 2022-04-13
fix: toast error corrections & cache flush on delete

dstala:
Signed-off-by: Raju Udava <86527202+dstala@users.noreply.github.com>

## Change Summary
- ext db: project created with same name, toast persisted
- xcdb: project created with same name - error message corrected
- cache issue: insert record was failing for following scenario [was working with cache disabled]
  - create project, delete project, create another project with same name
  - insert record using bulk create api (failing while trying to retrieve project id using project title) 

## Change type

## Test/ Verification
Locally verified

### DIFF
--- a/packages/nc-gui/components/createOrEditProject.vue
+++ b/packages/nc-gui/components/createOrEditProject.vue
@@ -1394,7 +1394,7 @@ export default {
         this.projectCreated = true
       } catch (e) {
         this.$toast.error(await this._extractSdkResponseErrorMsg(e)).goAway(3000)
-        toast.goAway(100)
+        toast.goAway(0)
       }
 
       this.projectReloading = false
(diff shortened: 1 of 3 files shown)


packages/nc-gui/components/createOrEditProject.vue
```diff
@@ -1403,6 +1394,7 @@ export default {
         this.projectCreated = true
       } catch (e) {
         this.$toast.error(await this._extractSdkResponseErrorMsg(e)).goAway(3000)
+        toast.goAway(100)
```
wingkwong, line 1397: Suggest to make it `0` instead of `100` so that we don't see the blue toast just like in xcdb one

dstala: updated

wingkwong (approved):

Issue #1647, opened 2022-04-12
Currency field- validation fails if decimal fractions are used


dstala:
Node: **v14.18.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
Database: **sqlite3**
ProjectOnRootDB: **true**
RootDB: **sqlite3**
PackageVersion: **0.84.16**

- Create 'Currency' field
- Insert 3.3

```
Error: Validation failed : Invalid Currency Format
    at BaseModel.validate (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/src/lib/dataMapper/lib/sql/BaseModelSql.ts:124:17)
    at BaseModel.updateByPk (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/src/lib/dataMapper/lib/sql/BaseModelSql.ts:350:18)
    at RestCtrl.update (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/src/lib/noco/rest/RestCtrl.ts:71:34)
    at /Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/src/lib/noco/rest/RestBaseCtrl.ts:120:31
    at Layer.handle [as handle_request] (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/node_modules/express/lib/router/route.js:137:13)
    at RestCtrl.middleware (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/src/lib/noco/rest/RestCtrl.ts:341:16)
    at Layer.handle [as handle_request] (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/rajuudava/NocoDB/nocodb-v1/nocodb/packages/nocodb/node_modules/express/lib/router/route.js:112:3)
```

Issue #1352, opened 2022-02-24
[Feature] Currency Column Improvement


vijayrathore8492:
## Recommending Features

NocoDB thrives on community needs. Many of our existing features are derived from community requirements & we will continue to build that way. We invite ideas to make NocoDB better. Do note that, we being an open-source team are constrained by resources & will be able to pick only limited **high demand** ideas for release. So, if you wish to pick this feature request early - spread the word, discuss it in Discord/blog/Reddit, ask your peers & colleagues to vote up for this feature request!

**Need for this feature**
Current currency column is very basic and does not serve it's purpose fully. 

**Describe the solution you'd like**
Currency column creation window should offer a currency and a data type - INT, DECIMAL for the new column. 
The currency should be indicated in column header and every value in a column should be postfixed or prefixed with the currency.

**Describe alternatives you've considered**
I am using regular decimal column for now as currency column provides nothing extra.

wingkwong:
For the data type, you can click ``Show more options`` to choose it.

<img width="394" alt="image" src="https://user-images.githubusercontent.com/35857179/155513396-532180e1-b941-4890-93be-d11d008e27d1.png">

vijayrathore8492:
@wingkwong I am aware of that. But that brings no more advantage than a plain number column.

wingkwong:
@vijayrathore8492 Agree. Some also shared their thoughts about currency field (See [here](https://github.com/nocodb/nocodb/issues?q=is%3Aissue+is%3Aopen+currency)). We'll work on it accordingly.

mertmit:
@vijayrathore8492 I am happy to announce we have revised the currency column to support different locales and currency codes. I hope our new implementation in develop branch cover your needs. Feel free to update the issue if anything missing for you.
![image](https://user-images.githubusercontent.com/59797957/170151515-7d0ff620-d1d2-44d3-a7e8-62947d4fcc15.png)

Issue #1042, opened 2022-02-03
Currency validator crashing when not adding two decimal places


FerLSanchez:
NocoDB used as docker : true
NocoDB version : 0.848
Database used in NC_DB URL  : mysql
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : mysql
OS on which NocoDB is running : Ubuntu
Database version : 8.0.28-1debian10
--

**Steps To Reproduce**
1. Create a project
2. Create a table
3. Create a Currency column, leaving database part as default (decimal(10,2))
4. Add a new register with a decimal with a single decimal place, like 55.9
5. In previous noco versions there were a message regarding the validator, now it is just a 400

**Expected behavior**
It should work and save the value (it works for 55.91)

**Screenshots**
<img width="563" alt="Screenshot 2022-02-03 at 18 55 00" src="https://user-images.githubusercontent.com/2619609/152400756-a29d993b-9b22-457e-94f8-b3863c7097c9.png">

o1lab:
55,9 -> there is a comma instead of a decimal - Is that a typo ?

FerLSanchez:
Spanish localization for numbers... it works for 55,91... :/

FerLSanchez:
In fact, the only workaround I have right now is entering a two decimal number like 55,91 and correct it directly in database... :_(

wingkwong:
@FerLSanchez It fails because the currency format is not correct. Two decimal places are fixed. Correct formats like ``123``, or ``123.50``.  However, it should show the validation error in Form View instead of 400 for better UX. It should be the one in Grid View. Let me take this.

![image](https://user-images.githubusercontent.com/35857179/152524927-aa9d6adc-ae01-4254-bb18-fe4e4682c62b.png)

FerLSanchez:
123,50 (Spanish-like) doesn't work, the field deletes the 0, and go back to the error. If I try to use the point instead of the comma, the field goes to 0 automatically as fast as I click the key...
Let me know if I can help with anything. Thank you very much.

wingkwong:
I'm able to reproduce it. It seems like the validation starts on keydown. I'm able to input ``12.50`` if I type very fast (before the validation starts). Will take a look.

FerLSanchez:
Great! Thank you very much!

wingkwong:
Let's summary a bit first. The problems we are facing right now are 

[...]

Issue #572, opened 2021-09-25
Currency fields always displayed with $


bannmann:
I created a currency field with "Type in database" set to `money`. I use PostgreSQL.

In my table, I want to work with prices in Euro (€), but the values I enter are always displayed as US Dollar ($).

Could you please have NocoDB allow selection of the currency to be used? This should be selectable for each column (or at least on project level) as I imagine it's not useful to simply rely on global settings such as the server system locale or database locale.

nil0x42:
I need this too !

mertmit:
@bannmann @nil0x42 I am happy to announce we have revised the currency column to support different locales and currency codes. I hope our new implementation in develop branch cover your needs. Feel free to update the issue if anything missing for you.
![image](https://user-images.githubusercontent.com/59797957/170151515-7d0ff620-d1d2-44d3-a7e8-62947d4fcc15.png)

mertmit:
Confirmed this is a problem with 'money' type from postgreSQL I've created an issue you can follow updates from there https://github.com/nocodb/nocodb/issues/2255

PR #1687, opened 2022-04-16
feat/fix: currency column

mertmit:
## Change Summary

Currency locale and currency code can be customized.

Refs: #572 #1042 #1352 #1647

- [x] Currency Column
  - [x] Multiple Locale Support
Used Intl.NumberFormat to format and validate
  - [x] Multiple Currency Code Support (ISO 4217)
- [x] Validation
I propose changing validator to isFloat instead because with this implementation we handle currency part apart from the value
- [x] Migration
  - [x] Handle extra variables as colOptions instead of seperate meta columns
  - [x] Change validator for migrated columns
There is no need to migrate validators with latest implementation

## Change type

- [x] fix

## Test/ Verification

image
image

## Additional information

- Country code reference: https://en.wikipedia.org/wiki/ISO_4217


packages/nc-gui/components/project/spreadsheet/components/editColumn.vue
```diff
@@ -597,6 +620,16 @@ export default {
         this.newColumn.dtxp = this.column.dtxp
       }
 
+      if (this.newColumn.uidt === 'Currency' && this.column?.uidt !== 'Currency') {
```
pranavxc, line 623: Use `UiTypes` enum for referring uidt whenever possible. So in the above case, you can replace `Currency` with `UiTypes.Currency`, where you can import `UiTypes` from `nocodb-sdk` package.

pranavxc (changes requested):

wingkwong (commented):
Just a heads up. We are putting v2 migration stuff in different places. Here's something you should change soon.

1. Move the changes you made in `XcMigrationSource.ts` to `XcMigrationSourceV2.ts`
2. Move `nc_012_add_currency_column.ts` to `migrationsv2` instead of `migrations`. You may also need to change the file name as `nc_012` is being occupied. 

However, we haven't merged it to `develop` branch yet. Will ping you when it is ready.

mertmit:
I will prepare migrations according to your instructions and wait for your ping to push proper changes updated with latest develop branch.

Meanwhile I updated UI:
image

wingkwong:
@mertmit Now it is in develop branch. Please pull the latest code and merge it accordingly. Thanks.

mertmit:
**Caution:** The migration still misses changing validate column to 'isFloat' for existent currency columns.

wingkwong:
@mertmit Marking it as Draft first. Please request a review again when it is ready.

mertmit:
https://github.com/nocodb/nocodb/pull/1687/commits/202df2d731727e15e1a7e6c5cead7a233d37dd91
on this commit I've created a custom validators file as this might come in handy in future (if we want to bypass any function provided by validate or want to remap as in this case)

wingkwong:
@mertmit Please rebase and resolve the conflicts first.

o1lab:
Im wondering whether it would be good idea to not keep separate table for currency. That way we don't end up creating a lot of tables for each types. A simple json column or text column (stringify json) would be sufficient ? @wingkwong @pranavxc

packages/nocodb/src/lib/noco-models/Column.ts
```diff
@@ -495,6 +510,18 @@ export default class Column<T = any> implements ColumnType {
       }
     }
 
+    {
+      let currs = await NocoCache.getList(CacheScope.COL_CURRENCY, [id]);
+      if (!currs.length) {
+        currs = await ncMeta.metaList2(null, null, MetaTable.COL_ROLLUP, {
```
wingkwong, line 516: I think it should be `COL_CURRENCY` instead of `COL_ROLLUP`

wingkwong:
> Im wondering whether it would be good idea to not keep separate table for currency. That way we don't end up creating a lot of tables for each types. A simple json column or text column (stringify json) would be sufficient ? @wingkwong @pranavxc

@mertmit please correct me if I'm wrong.

I think the table is not for storing all the possible currency options but storing the meta currency (locale & code) that users choose. I think keeping it in a separate table would make it more clean.

mertmit:
@wingkwong yes it only keeps colOptions (column_id, currency_locale, currency_code for currency) in created meta table.
@o1lab actually this is how we handle colOptions for other columns with colOptions. (formula, lookup etc.)

wingkwong:
@mertmit Please rebase and request review again if it's ready. Thanks.

mertmit:
@wingkwong ready for a review. Thanks.

mertmit:
> https://github.com/nocodb/nocodb/pull/1687/commits/202df2d731727e15e1a7e6c5cead7a233d37dd91
> on this commit I've created a custom validators file as this might come in handy in future (if we want to bypass any function provided by validate or want to remap as in this case)

@wingkwong This note left from first state I reverted this change, now it is handled within custom validators which I stated in this comment. You can check the commit in this comment to review validation.

wingkwong (approved):

pranavxc (approved):

PR #1757, opened 2022-04-19
Fix - LTAR based filter

pranavxc:
### Resolved issues

- [x] Fixed belongs to column-based filter issue
- [x] Nested record mapping issue when referring to the same table as child/parent.


wingkwong (approved):
Verified.

Issue #1754, opened 2022-04-19
System fields not being appropriately marked for linked records on Postgres


tlonic:
**Please enter the following details**

please provide the following details :
NocoDB used as docker : true
NocoDB version : 0.90.3 
Database used in NC_DB URL  : pg
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : pg
OS on which NocoDB is running : linux
Node.js version if running as node : 

**Steps To Reproduce**
1. Create a new database.
2. Create a new table.
3. In that table, create a linked record.
4. Look at project metadata, notice that system columns associated with linked records are out of sync with schema, as described in https://github.com/nocodb/nocodb/issues/1753
5. Sync schema
6. Go back to table view, look at fields.
7. See that columns associated with linked records (i.e. table_id columns) are not being correctly marked as system columns.

**Expected behavior**
Columns created to do behind the scenes stuff for linked records should not be showing up as non-system columns.

**Screenshots**
![image](https://user-images.githubusercontent.com/2394967/163896292-7dd2f6c1-6b8d-4069-9baf-87e1c821fcb5.png)

wingkwong:
It exists when the meta is out of sync. I did it in this order 1 -> 5 -> 2 -> 3 -> 6 and it looks fine.

wingkwong:
Fixed in `0.90.7`.

Issue #1753, opened 2022-04-19
schema immediately out of sync on postgres when creating table, columns


tlonic:
please provide the following details :
NocoDB used as docker : true
NocoDB version : 0.90.3
Database used in NC_DB URL  : pg
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : pg

**Steps To Reproduce**
1. Create new table
2. Go to Metadata 
3. See that metadata associated with system fields is already out of sync: 
"Column type changed(id), Column type changed(created_at), Column type changed(updated_at)" 
4. Create new column
5. See that metadata associated with new column is out of sync,
"Column type changed(newcolumn)" 

**Expected behavior**
Column type should already be synced on creating a table and a column

**Screenshots**
![image](https://user-images.githubusercontent.com/2394967/163896074-470edda8-6b0f-4b27-ac99-2f2f7c27ce17.png)

wingkwong:
Able to reproduce. It seems this issue only exists in Postgres.

rajeshdhanda:
still facing the issue, updated data in postgres via psql command line, not able to sync in nocodb. also modifying table from UI doesn't update the swagger UI and CRUD endpoints. i.e. dropped a column and added a new column via psql did not reflect the changes in UI and creates Internal server error in CRUD endpoints.  
docker used : nocodb/nocodb:latest

PR #1759, opened 2022-04-19
fix: Postgres meta out of sync

pranavxc:
Extract column metadata using `sqlClient` instead of relying on the request payload to avoid column metadata mismatch.

re #1753 #1754

### DIFF
--- a/packages/nocodb/src/lib/noco/meta/api/columnApis.ts
+++ b/packages/nocodb/src/lib/noco/meta/api/columnApis.ts
@@ -30,6 +30,7 @@ import ncMetaAclMw from '../helpers/ncMetaAclMw';
 import { NcError } from '../helpers/catchError';
 import getColumnPropsFromUIDT from '../helpers/getColumnPropsFromUIDT';
 import mapDefaultPrimaryValue from '../helpers/mapDefaultPrimaryValue';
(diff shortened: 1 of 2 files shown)


wingkwong (approved):

o1lab:
Can we have more details on this issue ?
- Would this affect on how metadata is generated already ? If so should we've to version migration here ?

pranavxc:
> Can we have more details on this issue ?
> 
> * Would this affect on how metadata is stored already ?
> * If so should we've to write version migration ?

The metadata in the column meta table is valid and it won't break anything. The problem here is in metaDiff we are querying and getting column metadata and comparing with our metadata, in the case of pg they have an alias for certain datatype so it won't be the same while comparing but both are valid. In the PR we handling this case by querying column meta to make type consistent so it won't show that change in metadiff anymore. 

If any existing project show those changes either user can ignore them or they can click apply changes which won't break anything and it will update with the metadata from the query result

o1lab:
The side effect of this PR will be latency seen in column creation where
1. first we create columns
2. then make another meta query from postgres
3. and then save it to metadb

2nd step will get really costly for a sizeable db. We've to find a work around for upcoming releases.

Issue #1814, opened 2022-04-21
JSON type data is not getting saved from UI


o1lab:
Reported by @vijayrathore8492 via discord

mertmit:
I wasn't able to reproduce the issue.
@vijayrathore8492 Is it possible for you to provide an example JSON to reproduce the issue (also your project info would be awesome)

JSON used:
```
{
  "name": "Noco",
  "age": 2,
  "car": null
}
```

**Project Info:**
Node: **v14.18.0**
Arch: **x64**
Platform: **linux**
Docker: **false**
Database: **pg**
ProjectOnRootDB: **false**
RootDB: **sqlite3**
PackageVersion: **0.90.5**

wingkwong:
Able to reproduce. 

Steps to reproduce:

- create a JSON type column
- copy @mertmit 's JSON and paste to the cell
- Refresh the page
- The data is not saved

wingkwong:
@mertmit The root cause is 

`jsonEditableCell` emits `input` to `editableCell` which emits `save` to `XcGridView`. However, `save` in `xcGridView` only executes `editEnabled = {}` but not saving the actual value. This also applies to other files using `EditableCell`.

mertmit:
Then this should be also solved by https://github.com/nocodb/nocodb/pull/1817

wingkwong:
Okay. Haven't checked that yet. Will review soon. Thanks!

o1lab:
0.90.7

PR #1821, opened 2022-04-21
feat: improve json editor

mertmit:
## Change Summary

#1814 might be caused by invalid input (users should not be allowed to save invalid json)
- Disable saving for invalid input
- Show errors to user

## Change type

## Test/ Verification

image


wingkwong:
This is useful! I'd take that. However, this is not the root cause of the issue you mentioned. I left some comments in the issue. Please take a look.

PR #1854, opened 2022-04-25
feat: accept project id or title in data apis

pranavxc:


wingkwong (commented):
If we potentially set `id` in cache. Then we need to delete it like `title` when deleting the project.

```js
await NocoCache.del(`${CacheScope.PROJECT}:${project.id}`);
```

wingkwong:
Forget the comment. It's been covered in `deelDel`.

wingkwong (approved):

wingkwong:
Introducing `<scope>:ref:<titleOfID> = <project_id>` to avoid collision with `<scope>:<project_id>` for the case `titleOfID` = `id`

Issue #1835, opened 2022-04-22
List API failing if child table has a JSON column


vijayrathore8492:
**Please enter the following details**

please provide the following details :
NocoDB used as docker : true / false
NocoDB version : 0.90.3
Database used in NC_DB URL  : sqlite3
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : pg
OS on which NocoDB is running : MacOS
Node.js version if running as node : 12
Database version : 

**Steps To Reproduce**
1. Create 2 tables `table1` and `table2`
2. Create a json column in `table1`
3. Add some data in the tables
4. Create relationship between tables, `table2` has many `table1`
5. Add data in relationship column
6. Try to load `table2`, No data will load

**Expected behavior**
It should show data in `table2`

**Errors**
This is the error in API, 
`msg: "DataLoader must be constructed with a function which accepts Array<key> and returns Promise<Array<value>>, but the function did not return a Promise of an Array of the same length as the Array of keys.\n\nKeys:\n2,3,4,5,6,7,8,9,10,11,14,15\n\nValues:\n"`.
and in console I saw this error from postgres `could not identify an equality operator for type json'`

**Fix**
Workaround for now is to select text as type in database while creating JSON column.
Using `jsonb` instead of `json` as default database type may fix it.

Join our discord : https://discord.gg/5RgZmkW for realtime help.

Node version : **v12.22.1**
Arch type: **x64**
Platform: **linux**
Docker: **true**
Database: **sqlite3**
packageVersion: **0.9.27**

vijayrathore8492:
@o1lab Making `jsonb` as default type for json seems to fix it.

o1lab:
Thank you - will take a look

PR #1876, opened 2022-04-26
fix: use union all instead of union since rows will be unique

pranavxc:
re #1835


o1lab:
Needs to see if CI/CD failure is due to the fix

https://github.com/nocodb/nocodb/runs/6172300503?check_suite_focus=true#step:6:516

wingkwong (approved):

PR #1878, opened 2022-04-26
fix: belongs to header tooltip

mertmit:
## Change Summary

Show column name instead of 'undefined'

## Change type


wingkwong (approved):
LGTM.

image

image

Issue #1932, opened 2022-04-29
Dialog containing monaco editor closes on click


RawmanGrey:
JSON column fields have an option to be edited on wider modal, but clicking on opened dialog closes it. 

<details> 
  <summary>.gif</summary>
 
![Peek 2022-04-29 18-32](https://user-images.githubusercontent.com/61871501/165978191-e34259a9-4011-4718-b0cd-9d73f50143b9.gif)

</details>

Node: **v12.22.12**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **pg**
ProjectOnRootDB: **true**
RootDB: **pg**
PackageVersion: **0.90.8**

PR #1936, opened 2022-04-30
fix: monaco editor dialog closes on click

mertmit:
## Change Summary

Re #1932
Disables on outside click listener for monaco editor.
This way:
- the dialog doesn't close when you click inside
- keeps data when you minimize the dialog
- also makes it harder to accidentally close monaco editor while minimized (closed with proper buttons or selecting another cell)

## Change type


wingkwong (commented):
- Existing JSON Column: The issue is still reproducible. 
- Newly created JSON Column: The issue is not reproducible.

Issue #1964, opened 2022-05-04
Formula fields throw an error when referencing fields that start with a number


shtev21:
```
Node: **v16.14.0**
Arch: **x64**
Platform: **win32**
Docker: **false**
Database: **mysql2**
ProjectOnRootDB: **true**
RootDB: **mysql2**
PackageVersion: **0.84.16**
```

**Steps To Reproduce**

1. In any table, create a new column of any type and input a `Column name` that starts with a number (e.g. `1st`)
2. In the same table, create a new column of the `Formula` type and attempt to reference the column named `1st`
3. You should see an error: `Variable names cannot start with a number (foo) at character bar`
4. Cancel that menu and change the name of the column from `1st` => `First`
5. Edit the formula column and add a reference to the column `First` to see that it resolves successfully
6. Save the changes and close the edit menu
7. Rename the column back from `First` => `1st`
8. Edit the formula column to see that the reference has also changed to match
9. Note that this does not throw any errors unless you try to edit the reference at which point it will once again throw the variable name error.

**Expected Behavior**

Allow for numbers to be used as column names and to be referenced by formulae without issue
OR
Have consistent behaviour by enforcing stricter column naming rules

**Screenshots**

![image](https://user-images.githubusercontent.com/2514805/166605334-ec4e39b9-4b30-4618-a92c-fa357ced80d9.png)
![image](https://user-images.githubusercontent.com/2514805/166605354-1f4ec736-afd1-4e28-ad32-c4efffe3b27a.png)

wingkwong:
preliminary findings:

- `formula` takes the correct column id
- `formula_raw` still takes the old value

PR #1979, opened 2022-05-05
fix: formula

wingkwong:
Ref: #1964 (partially)

## Change Summary

- Introduce a custom curly hook for column_name, e.g. {column_name}
  - Why? If we have a column `123`, we can't distinguish whether `123` in `ADD(123,123)` is numeric or column.  We can solve it using `ADD({123}, 123)`.
- Update `formula_raw` when the reference column name is changed.
  - Just keep it consistent with `formula`. 
- Update Formula Documentation for the new changes as well as the outdated content
- Code Refactoring
  - centralise formula helpers in sdk 
- UI Refactoring
  - Previously the menu list blocks the save / cancel button 
  image

## Change type

## Test/ Verification

- Try reference the column name with `{}`. E.g. `CONCAT({Title}, {Title2})`.
- Try reference the column name without `{}`.
- Try numeric & string combination
- Try those formulas in Documentation

## Additional information

- Version Migration
  - There is no need for version migration as we are using `formula` which refers the columns using column id. We render substitute the id with alias and render `{}` for columns if applicable.


o1lab:
The hyper link in the modal `formulas` is not redirecting or opening.

pranavxc (approved):

Issue #1955, opened 2022-05-03
Single select issues (reported in discord)


o1lab:
Hi, I'm trying out NocoDB and I'm loving it! I have two questions, though: 
- is it possible to reorder records, like drag&drop or some other way? 
- I'm using docker latest image and when adding a new SingleSelect field, the options provided are not saved. But I can edit the field and add them after. Did anyone come across this problem? It seems it didn't happen in a version I tried prior to 0.90

mertmit:
The issue seems reproducible on PostgreSQL and SQLite.
Also there seems to be another SingleSelect related issue on MSSQL.

PR #1986, opened 2022-05-06
fix: single/multi select options at creation

mertmit:
## Change Summary

Re #1955 
As options are handled from meta for databases beside MySQL the options should be appended to created column meta.

## Change type


o1lab:
LGTM.

PR #2026, opened 2022-05-12
Feat : UIDT impmporovements

pranavxc:
## Change Summary

UI datatype improvements and customization.

## Change type

- [x] Rating type customisation and improvements
- [x] Checkbox type customisation
- [x] Email/URL - optional validation and rendering as link

### Rating

Screenshot 2022-05-12 at 2 07 59 PM
Screenshot 2022-05-12 at 2 08 13 PM

### Checkbox

Screenshot 2022-05-12 at 2 09 34 PM
Screenshot 2022-05-12 at 2 10 32 PM


wingkwong (commented):
LGTM. Just one thought. 

If I set max rating to 8, select 8, then change the max rating to 4. What is the value supposed to be? The current behaviour is taking the previous value which is 8 but you can only select 4 for others. May need to discuss whether we should take the upper bound or just use the existing value.

dstala:
1. Gallery view, we do not show 'rating' & show numbers instead
Screenshot 2022-05-13 at 3 55 00 PM
2. v1, if numbers were high (rating did not have any restriction in v1) - it gets mapped to default 5* in v2

PR #2069, opened 2022-05-18
Feat - UUID auto generate and bug fixes

pranavxc:
## Change Summary

- [x] [Enhancement]Default table name population in table create dialog
- [x] [Feat]Allow autogenerated id option in table create API
- [x] [Bug]Hide system fields from expanded nested row dialog
- [x] [Bug]Flickering effect in the Dashboard
- [x] Sort based on created_at if pk is not auto-increment


dstala:
1. Show system fields, don't enable migration-specific fields. by design?
Screenshot 2022-05-18 at 1 04 34 PM

PR #2120, opened 2022-05-23
Feat - Next release improvements and bug fixes

pranavxc:
## Change Summary

- [x] [Enhancement] Webhook autofill with {{json data}}
- [x] [Bug] Nested lookup missing when LTAR is hidden
- [x] [Bug] Handling special chars in column/table name
- [x] [Bug] Multi/Single select rendering and inserting corrections
- [x] [Fix] Flickering in the Dashboard
- [x] [Enhancement] Hide system fields from expanded nested row dialog
- [x] [Enhancement] Remove virtual columns from expanded form while adding row


wingkwong:
image

Is this supposed not to show for editors & viewers? The difference is In current develop branch we are showing it for them also. That explains REST Base VIEW share failures in cypress tests.

https://github.com/nocodb/nocodb/blob/8c2a16c6783ccfd232f68614cba87275e67c7f91/scripts/cypress/integration/spec/roleValidation.spec.js#L217

If so, then we need to add an else-if-block for viewers to set `navDrawListCnt` to `1` (view list) rather than `2` (view list + the list from the above screenshot).

wingkwong (approved):

Issue #2127, opened 2022-05-23
[Bug] Airtable import not working in Docker if mapped to a different port


pranavxc:
### Summary

Airtable import is not working in Docker if mapped to a different port other than `8080`.

### How to reproduce

```
docker run -p 9000:8080 nocodb/nocodb:0.91.0
```

![Screenshot 2022-05-23 at 6 13 12 PM](https://user-images.githubusercontent.com/61551451/169821996-50a1e73a-64c3-4804-9aa0-ffe80d574455.png)

PR #2134, opened 2022-05-24
fix: Support Airtable import even mapped to a different port

pranavxc:
## Change Summary

In the existing release it Airtable import will fail in docker if mapped to a port other than `8080`.

## Change type

## Test/ Verification

Screenshot 2022-05-24 at 11 11 12 AM

re #2127


pranavxc:
> @pranavxc Since PR Release failed, I triggered nightly build on [f99ad57](https://github.com/nocodb/nocodb/commit/f99ad570bee60d13ca5f964cca811f39128cdd3b).
> 
> I got `Error: connect ECONNREFUSED 127.0.0.1:80`.
> 
> image
> 
> Here's the command I used. Not sure if I missed some env var or not.
> 
> ```
> docker run -d --name nocodb-pr2134 \
> -p 7777:8080 \
> -e NC_BASEURL_INTERNAL=7777 \
> nocodb/nocodb-timely:0.91.0-20220524-1028
> ```

In the case of docker, you can ignore the `NC_BASEURL_INTERNAL` since it will populate based on `PORT`. If you are providing it need to be a full URL like `http://localhost:8080`. 


Within docker port `7777` is not available so the value should be `http://localhost:8080` unless you are changing the `PORT` env value.

wingkwong (approved):
Oh yes. I misread it. Verified using your command.

PR #2205, opened 2022-05-30
feat: hook active switch

willnewii:
## Change Summary

add hook active switch

image
image

## Change type

if database use sqlite, hooks records active field default value is null. will affect old data.
How should this issue be handled？


bcakmakoglu:
@willnewii 
This PR is sadly made for the old GUI, which is now gone.

If possible I would kindly ask you to update to the latest GUI and open a new PR with this feature :)

Thank you for your time and contribution so far though =)

PR #2259, opened 2022-06-04
feat: duration

wingkwong:
### Support formats

image

h:mm
image

h:mm:ss
image

h:mm:ss.s
image

h:mm:ss.ss
image

h:mm:ss.sss
image

## Change type


packages/nc-gui/package.json
```diff
@@ -28,6 +28,7 @@
     "jsep": "^1.3.6",
     "locale-codes": "^1.3.1",
     "material-design-icons-iconfont": "^5.0.1",
+    "moment": "^2.29.3",
```
o1lab, line 31: Screenshot 2022-06-05 at 09 36 17

Moment is quite heavy. I guess we'd used an alternative. Please check.

karishmashuklaa: @wingkwong We already have <a href="https://github.com/nocodb/nocodb/blob/develop/packages/nc-gui/package.json">dayjs</a> as a dependency

wingkwong: @karishmashuklaa You're right. Will switch to dayjs then.

wingkwong: @karishmashuklaa dayjs may not a way to go as it returns `NaN` if we input `minutes: 65`, while momentjs would return `hours: 1` and  `minutes: 5` which is expected. Will think about it.

wingkwong: Let's ditch both

packages/nc-gui/helpers/durationHelper.js
```diff
@@ -0,0 +1,201 @@
...
+
+// pad zero
+// mm && ss
+// e.g.  3 -> 03
+// e.g. 12 -> 12
+// sss
+// e.g.  1 -> 001
+// e.g. 10 -> 010
+const padZero = (val, isSSS = false) => {
+  if (isSSS) {
```
pranavxc, line 38: you can simplify it by using 

```js
return `00${val}`.slice(isSSS ? -3 : -2)
```

or by using `padStart`

```js
return (val + '' ).padStart(isSSS ? 3 : 2, '0')
```

wingkwong:
Closing as it includes lots of unnecessary commits. Will create a new one.

PR #2290, opened 2022-06-08
modification of the traduction

QuentinDstl:
I aslo add some stuff according to the README (it was a different display so i didn't add a lot)


o1lab:
Thank you. 

@all-contributors please add @QuentinDstl for code

Issue #2301, opened 2022-06-09
Project Invite: Lost admin access to my project


ed-parsadanyan:
```
Node: **v12.22.12**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **pg**
ProjectOnRootDB: **true**
RootDB: **pg**
PackageVersion: **0.91.7**
```

1. Go to Team&Auth -> User management -> New User
2. Invite Team: by accident entered the same email as the admin account, but picked the role "viewer"
4. Clicked 'Invite' and now due do access restrictions don't see anything. It seems that I downgraded myself.

**Expected Behavior**

An error that the user already exists

o1lab:
If user exists, the invite leads to update of the user role. We should send back an error that `user@gmail.com exists with these roles editor.. etc`

ed-parsadanyan:
@o1lab Thanks! is there a way how to get access back? Do I need to change something in postgres DB directly?

wingkwong:
@ed-parsadanyan Seems behind the scene it didn't change the roles in DB but in cache. In this case, you may just restart the application so that you'd get the correct cache.

ed-parsadanyan:
@wingkwong Thanks! I can confirm that
`sudo docker-compose stop
sudo docker-compose up -d
`

resolved the access issues

PR #2303, opened 2022-06-09
fix: missing logic in userInvite

wingkwong:
## Change Summary

Invite and Edit user under User Management would trigger the same function `saveUser`. For the latter one, it would trigger `projectUserUpdate` while the other one would triggers `projectUserAdd`, which later calls `userInvite`. However, Add User To Project would also trigger the same as Invite. In current `userInvite` function, if the user exists, it would be updated and the cache will be changed also. And this logic should be only for the existing users who are not in the project.

Check if the target user is in the current project or not. If so, throw an error.

ref: #2301 

## Change type

## Test/ Verification

User Exists + Already in the project

image

User Exists + Not in the project

image

User Not Exists

image


o1lab:
Tests are through. LGTM

Issue #2306, opened 2022-06-09
REST API JSON output limited to 1000


fishnux:
```
NocoDB used as docker : true 
NocoDB version : 
Database used in NC_DB URL  : mysql 
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : (unsure)
OS on which NocoDB is running : Linux
Node.js version if running as node : As Docker
Database version : Unsure
```

**Steps To Reproduce**

Do a Get All REST API with limit 10000 offset 0 on table that has more than 1000 rows

**Expected Behavior**

Returns more than 1000 rows (assuming table has about 4000 rows)

Defined ENV variables:
```
DB_QUERY_LIMIT_MAX = 10000
DB_QUERY_LIMIT_DEFAULT = 100
```

wingkwong:
@fishnux Given the ENV variables you've set, how many records you've got? 100?

fishnux:
> @fishnux Given the ENV variables you've set, how many records you've got? 100?

I get exactly 1000 elements in the JSON, in the database I have about 4000 rows. The "PageInfo" also seemed to indicate all 4000 rows were returned (despite only having 1000 elements), though I don't remember for sure

wingkwong:
Can you show us `pageInfo` in your returned json?

fishnux:
> Can you show us `pageInfo` in your returned json?

```
{
  "totalRows": 4050,
  "page": 1,
  "pageSize": "10000",
  "isFirstPage": true,
  "isLastPage": true
}
```

PR #2241, opened 2022-06-02
docs: add information about maximum value for `limit` in REST API

nilsreichardt:
## Change Summary

image

When I used the REST API I wanted to know how big the `limit` parameter can be. In my case, it was 1,000. Even with a value of `100.000` I got 1,000 results back (the table contains 50,000 email addresses).

Therefore, I added this information to the documentation. 

## Change type

- [ ] feat: (new feature for the user, not a new feature for build script)
- [ ] fix: (bug fix for the user, not a fix to a build script)
- [x] docs: (changes to the documentation)
- [ ] style: (formatting, missing semi colons, etc; no production code change)
- [ ] refactor: (refactoring production code, eg. renaming a variable)
- [ ] test: (adding missing tests, refactoring tests; no production code change)
- [ ] chore: (updating grunt tasks etc; no production code change)

## Test/ Verification

Provide summary of changes.

## Additional information / screenshots (optional)

Anything for maintainers to be made aware of

PR #2318, opened 2022-06-10
enhancement: make query limit configurable

wingkwong:
## Change Summary

ref: #2306, #2241

In v2,  the values for `limitDefault`, `limitMin`, and `limitMax` are hardcoded. This PR is to make them configurable.

## Change type

## Test/ Verification

Using Sakila - Inventory with 4581 records in total.

## Test 1: Incoming limit  > `DB_QUERY_LIMIT_MAX = 1234`

```js
var axios = require("axios").default;

var options = {
  method: 'GET',
  url: 'http://localhost:8080/api/v1/db/data/noco/p_lf52np1h1deon6/Inventory/views/Inventory',
  params: {limit: '2000', offset: '0', where: ''},
  headers: {
    'xc-auth': '<REDACTED>'
  }
};
axios.request(options).then(function (response) {
  console.log(response.data.list.length);
}).catch(function (error) {
  console.error(error);
});
```

```
1234
````

## Test 2: Incoming limit  <= `DB_QUERY_LIMIT_MAX = 1234`

```js
var axios = require("axios").default;

var options = {
  method: 'GET',
  url: 'http://localhost:8080/api/v1/db/data/noco/p_lf52np1h1deon6/Inventory/views/Inventory',
  params: {limit: '1200', offset: '0', where: ''},
  headers: {
    'xc-auth': '<REDACTED>'
  }
};
// same as above
```

```
1200
````

## Test 3: Without setting `DB_QUERY_LIMIT_MAX`

```js
var axios = require("axios").default;

var options = {
  method: 'GET',
  url: 'http://localhost:8080/api/v1/db/data/noco/p_lf52np1h1deon6/Inventory/views/Inventory',
  params: {limit: '1200', offset: '0', where: ''},
  headers: {
    'xc-auth': '<REDACTED>'
  }
};
// same as above
```

By default, the value is `1000`. This case is same as `Incoming limit  > 1000 (max)`.

```
1000
````

## Test 4: Without setting `DB_QUERY_LIMIT_MAX`

```js
var axios = require("axios").default;

var options = {
  method: 'GET',
  url: 'http://localhost:8080/api/v1/db/data/noco/p_lf52np1h1deon6/Inventory/views/Inventory',
  params: {limit: '800', offset: '0', where: ''},
  headers: {
    'xc-auth': '<REDACTED>'
  }
};
// same as above
```

By default, the value is `1000`. This case is same as `Incoming limit  <= 1000 (max)`.

```
800
````

### CODE REVIEW

File: packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts [hunk 1]
```diff
@@ -70,9 +70,9 @@ class BaseModelSqlv2 {
   private _columns = {};
 
   private config: any = {
-    limitDefault: 25,
-    limitMin: 1,
-    limitMax: 1000
+    limitDefault: process.env.DB_QUERY_LIMIT_DEFAULT || 25,
```
pranavxc, line 73: We need to handle negative and invalid numbers here, something like the following we can do.

```js
limitDefault: Math.max(+process.env.DB_QUERY_LIMIT_DEFAULT || 25,1)
```

pranavxc (changes requested):

pranavxc (approved):

o1lab:
We need docs update for this ?

pranavxc:
> We need docs update for this ?

Screenshot 2022-06-10 at 3 43 22 PM


Need update over here(100 =>1000), I will update that part.

PR #2338, opened 2022-06-13
fix: insufficient session expiration

wingkwong:
## Change Summary

Added a new column `token_version` in User. Compare with the jwt token_version.

ref: #2262 

## Change type

## Test

- Log into same account using two different browsers
- Change the password in browser A -> force logout
- Reload the page in browser B -> force logout
- Change the password in browser A -> force logout -> login again
- Reload the page in browser B -> force logout

### DIFF
--- a/packages/nocodb/src/lib/meta/api/userApi/userApis.ts
+++ b/packages/nocodb/src/lib/meta/api/userApi/userApis.ts
@@ -71,7 +71,8 @@ export async function signup(req: Request, res: Response<TableType>) {
         password,
         email_verification_token,
         invite_token: null,
-        invite_token_expires: null
+        invite_token_expires: null,
+        email: user.email
       });
     } else {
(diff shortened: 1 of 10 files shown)


packages/nocodb/src/lib/meta/api/userApi/userApis.ts
```diff
@@ -285,7 +291,8 @@ async function passwordChange(req: Request<any, any>, res): Promise<any> {
   await User.update(user.id, {
     salt,
     password,
-    email: user.email
+    email: user.email,
+    token_expired: true
```
pranavxc, line 295: What will happen if there are multiple browser sessions (multiple jwt tokens) and one of the users sign-in using new credentials? other tokens will be still valid, right since `token_expired` is set to `true`?

pranavxc (approved):

PR #2399, opened 2022-06-16
Feat: JSON import

pranavxc:
## Change Summary

Import data and table from a JSON array or object.

## Change type

https://user-images.githubusercontent.com/61551451/174236263-d5f47682-6666-465a-b463-548f3d3856f8.mov

### DIFF
--- a/packages/nc-gui/components/import/JSONImport.vue
+++ b/packages/nc-gui/components/import/JSONImport.vue
@@ -23,7 +23,7 @@
           </v-tab>
 
           <v-tab-item>
-            <div class="nc-excel-import-tab-item ">
+            <div class="nc-json-import-tab-item ">
               <div
                 class="nc-droppable d-flex align-center justify-center flex-column"
                 :style="{
@@ -57,7 +57,7 @@
             </div>
           </v-tab-item>
           <v-tab-item>
-            <div class="nc-excel-import-tab-item align-center">
+            <div class="nc-json-import-tab-item align-center">
               <div class="pa-4 d-100 h-100">
                 <v-form ref="form" v-model="valid">
                   <div class="d-flex">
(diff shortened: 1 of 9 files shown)


o1lab:
We can call - `Normalise nested` to `Flatten nested keys`

pranavxc:
> 1. Can you attach the testing json in description? Any specific format required?
> 2. Mind to explain when to tick normalise nested?
>    image
> 3. When I click "Select file to Upload", it shows
>    image

1. JSON can be an object or an array of objects with at least one element.
2. I will add a tooltip for the normalize option. In short it will flatten the object if there is any nested object. For example if normalise enabled following json `{ "a" : { "b" : 1 } }` will treat as `{ "a_b" : 1 }`.
3. File-based import I will add the correction.

wingkwong (commented):
4. null handling 
image
5. better add some samples for users to follow (e.g. just header / multiple records etc)

wingkwong:
Sorry. Deleted the remote branch & closed the PR by mistake. Reopening it again.

Besides, importing below json fails. Please take a look.

```json
 {
     "country" : "Country 1",
     "lang" : "Language 1"
 }
```

packages/nc-gui/components/import/JSONImport.vue
```diff
@@ -0,0 +1,479 @@
...
+      let file
+      if (ev.dataTransfer.items) {
+        // Use DataTransferItemList interface to access the file(s)
+        if (ev.dataTransfer.items.length && ev.dataTransfer.items[0].kind === 'file') {
+          file = ev.dataTransfer.items[0].getAsFile()
+        }
+      } else if (ev.dataTransfer.files.length) {
+        file = ev.dataTransfer.files[0]
+      }
+
```
wingkwong, line 404: Suggested to check if the inputting file is a json here. If not, return error message using `this.$toast.error(...)`. Just like other import types.

packages/nc-gui/components/import/JSONImport.vue
```diff
@@ -0,0 +1,479 @@
...
+              </template>
+            </v-checkbox>
+          </div>
+        </div>
+      </v-card>
+    </v-dialog>
+
+    <v-tooltip bottom>
+      <template #activator="{on}">
+        <input
```
wingkwong, line 191: As it's only for JSON, we should only allow json type. Also by doing so, other types will be dimmed so that users can easily select the target file.

```
accept=".json"
```

wingkwong (commented):
Better revise `JSON: JSON`. No idea what it means.

image

pranavxc:
> Better revise `JSON: JSON`. No idea what it means.
> 
> image

Replaced with `JSON Import`

wingkwong (approved):

Issue #1469, opened 2022-03-22
Failed to Update Row - 400


HelloImCamden:
**Steps To Reproduce**
1. Add Row
2. Edit any information in row
3. Save Row
Error 400

**Expected behavior**
Row would be saved. 

**Screenshots**
https://cdn.discordapp.com/attachments/728443930621902869/955903010503028766/unknown.png

Docker logs: https://www.toptal.com/developers/hastebin/conevihimi.sql

Node: v12.22.10
Arch: x64
Platform: linux
Docker: true
Database: mysql2 ( 10.3.31-MariaDB-0+deb10u1 )
ProjectOnRootDB: false
RootDB: sqlite3
PackageVersion: 0.84.15

wingkwong:
Finding :

- The column name includes a character `?` which causes this error.

Cirsten:
I have same problem, when changing column type

Node: **v12.22.12**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **pg**
ProjectOnRootDB: **false**
RootDB: **pg**
PackageVersion: **0.91.7**

PR #2424, opened 2022-06-18
fix: handle ? in column name when inserting & updating

wingkwong:
## Change Summary

If there is `?` in column name, knex will see it as binding, even escape it with `\\?`. (ref: #1469)

## Change type

image


packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -313,8 +314,10 @@ class BaseModelSqlv2 {
     qb.count(this.model.primaryKey?.column_name || '*', {
       as: 'count'
     }).first();
-
-    return ((await qb) as any).count;
+    const res = (await this.dbDriver.raw(
+      qb.toQuery().replaceAll('\\?', '?')
```
pranavxc, line 318: Is `.replaceAll('\\?', '?')` required here ?  This can leads to problem if table name / alias contain `?` in pg database.

packages/nocodb/src/lib/models/Model.ts
```diff
@@ -399,13 +399,18 @@ export default class Model implements TableType {
...
     const insertObj = {};
     for (const col of await this.getColumns()) {
       if (isVirtualCol(col)) continue;
-      const val =
-        data?.[sanitize(col.column_name)] ?? data?.[sanitize(col.title)];
-      if (val !== undefined) insertObj[sanitize(col.column_name)] = val;
+      const val = data?.[col.column_name] ?? data?.[col.title];
+      if (val !== undefined) {
+        insertObj[
+          col.column_name.replace(/([^\\]|^)(\?+)/g, (_, m1, m2) => {
```
pranavxc, line 409: You can use the `sanitize` method here instead of repeating

pranavxc (changes requested):

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
wingkwong: Missed the sanitise part for column name. Also use `unsanitise` on the final query. Tested in all dbs.

packages/nocodb/src/lib/models/Model.ts
wingkwong: Good spot. Forget to update this one. Just updated.

pranavxc (approved):

PR #2438, opened 2022-06-20
fix: webhook triggers

mertmit:
## Change Summary

Re #2443
- Fixed webhook trigger
- mixins
  - isAutoSaved: for columns that auto-saved while editing without triggering webhook
  - isManualSaved: for columns that are not compatible with saving while editing (post formatting of cell ex. currency)

## Change type

## Tests

**Webhooks fired from every UIType successfully: **
image


wingkwong (commented):
Please include the original issue statement for the record.

wingkwong (approved):

pranavxc (approved):

PR #2454, opened 2022-06-22
fix nested field projection

vijayrathore8492:
Signed-off-by: Vijay Kumar Rathore <professional.vijay8492@gmail.com>

## Change Summary

In the data list API Field projection is not working for nested columns if the 'fields' query parameter set

## Change type

## Test/ Verification

In the list API, if you provide `fields` and `nested[relatedTable][fields]` both, nested projection does not get applied.

### DIFF
--- a/packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/helpers/getAst.ts
+++ b/packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/helpers/getAst.ts
@@ -86,7 +86,7 @@ const getAst = async ({
             (!fields?.length || fields.includes(col.title)) &&
             value
           : fields?.length
-          ? fields.includes(col.title)
+          ? fields.includes(col.title) && value
           : value
     };
   }, Promise.resolve({}));


packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/helpers/getAst.ts
```diff
@@ -86,7 +86,8 @@ const getAst = async ({
             (!fields?.length || fields.includes(col.title)) &&
             value
           : fields?.length
-          ? fields.includes(col.title)
+          ? fields.includes(col.title) &&
+            (col.uidt !== UITypes.LinkToAnotherRecord || value)
```
pranavxc, line 90: Can you briefly explain why this change is required?

pranavxc: Got it, better to change as follows `fields.includes(col.title) && value`

```js
    return {
      ...(await obj),
      [col.title]:
        allowedCols && (!includePkByDefault || !col.pk)
          ? allowedCols[col.id] &&
            (!isSystemColumn(col) || view.show_system_fields) &&
            (!fields?.length || fields.includes(col.title)) &&
            value
          : fields?.length
          ? fields.includes(col.title) && value
          : value
    };
```

pranavxc (changes requested):

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/helpers/getAst.ts
vijayrathore8492: seems to be working, I will make the changes.

PR #2471, opened 2022-06-23
Feat/executable in GitHub release

pranavxc:
## Change Summary

Build executables and attach them to the GitHub release.

## Change type

## Test/ Verification

Tested and verified in the forked repo: https://github.com/pranavxc/nocodb/runs/7024678227?check_suite_focus=true

### Steps followed for testing 

- Created a draft release with the tag `0.91.7` 
- Triggered action (`Release : Executables`) with input  `0.91.7`

#### Expected output

- Attachment should be present in the draft release - https://github.com/pranavxc/nocodb/releases/tag/0.91.7
- The homebrew repo should get updated - https://github.com/nocodb/homebrew-nocodb/commit/b9aef5dd101ac79da122d8d3e07f2ca001116a25


wingkwong (commented):
Can you explain what are those *.node for ?

image

wingkwong (commented):
What is the reason to have a separate workflow file i.e. release-timely-executables.yml`? It seems most of the logic is same. Wondering if it'd be better to just keep in one file and use conditions to control the timely logic based on a specific flag. What do you think?

pranavxc:
> What is the reason to have a separate workflow file i.e. release-timely-executables.yml`? It seems most of the logic is same. Wondering if it'd be better to just keep in one file and use conditions to control the timely logic based on a specific flag. What do you think?

Even I thought about it since you are doing the same with other releases. But for time being I kept it independent and simple, once everything starts working we can combine them.

wingkwong (approved):
Go for it.

wingkwong:
@tomorrowcheng Mind to ping me in Discord (ID: 3278) for the discussion?

tomorrowcheng:
> @tomorrowcheng Mind to ping me in Discord (ID: 3278) for the discussion?

thanks for your reply，I have found the cause of the problem，no need for discussion

Issue #2523, opened 2022-07-01
When I try to choose a value in a many-to-many relationship NocoDB's interface doesn't show me options.


lcfd:
```
Node: **v12.22.12**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **pg**
ProjectOnRootDB: **true**
RootDB: **pg**
PackageVersion: **0.92.0**
```

**Steps To Reproduce**

1. Go to a table with a linked field to another table with a many-to-many relationship
2. Add a new row
3. Try to pick a value in the many-to-many column
4. No options are available

```
msg: "... where ((((\"created_at\" ilike '%%%%')))) order by \"id\" asc limit 10 - operator does not exist: timestamp without time zone ~~* unknown"
```

**Expected Behavior**

The other table is full of records so I expect to see values to choose from below the `filter query` input box.

**Screenshots**

![CleanShot 2022-07-01 at 11 04 20](https://user-images.githubusercontent.com/9001053/176863415-4c1d288d-a142-4200-9057-38e44ff28791.png)

![CleanShot 2022-07-01 at 10 59 26](https://user-images.githubusercontent.com/9001053/176862237-65973607-90b1-4df2-8347-a776312efffc.png)

Thank you!

mertmit:
This is about having non-text field in like condition (CreatedAt) related with #2517 fix should be applicable for both

justinmayer:
I have also run into this issue, which OP has documented very well. (Bravo, OP!)

Could a maintainer add this issue to the [next milestone](https://github.com/nocodb/nocodb/milestone/8), with the intention of ensuring this problem is addressed in the next bug-fix update?

mertmit:
@lcfd Is it possible for you to confirm if this resolved on develop branch (or if you prefer docker/executables you can use the ones provided #2531) as I couldn't reproduce the exact case

Issue #2517, opened 2022-06-30
Filtering related records is not possible when system column is used as primary value


pavvel11:
```
Node: **v14.18.1**
Arch: **x64**
Platform: **freebsd**
Docker: **false**
Database: **pg**
ProjectOnRootDB: **false**
RootDB: **pg**
PackageVersion: **0.92.0**
```

**Steps To Reproduce**

1. Create two tables 
2. Add relation (Link To Another Record column)
3. Add at least one record in both tables
4. In the first table set one of the system columns (id, CreatedAt, UpdatedAt) as primary value (_Set as Primary Value_ option on the column)
5. In the second table click plus button to add relation and use Filter query.
Filtering is not working - HTTP 400 is returned from the server. 
Example errors:
`{"msg":"select \"test4\".\"id\" as \"Id\", \"test4\".\"created_at\" as \"CreatedAt\", \"test4\".\"updated_at\" as \"UpdatedAt\", \"test4\".\"test3_id\" as \"test3_id\", \"test4\".\"title6\" as \"title6\" from \"test4\" where (\"test3_id\" not in (select \"id\" from \"test3\" where \"id\" = '1') or \"test3_id\" is null) and ((\"id\" ilike '%1%')) limit 10 - operator does not exist: integer ~~* unknown"}`
`{"msg":"select \"test3\".\"id\" as \"Id\", \"test3\".\"created_at\" as \"CreatedAt\", \"test3\".\"updated_at\" as \"UpdatedAt\" from \"test3\" where (\"id\" not in (select \"test3_id\" from \"test4\" where \"id\" = '2' and \"test3_id\" is not null) or \"id\" is null) and ((\"created_at\" ilike '%11%')) limit 10 - operator does not exist: timestamp without time zone ~~* unknown"}`

**Expected Behavior**
Filtering should work no matter which column is used as a primary value.

dstala:
```
POST /api/v1/db/meta/views/vw_94yfmx247j3yfm/filters 200 351 - 8.601 ms
[...]

PR #2531, opened 2022-07-01
fix: pg cast type for like condition

mertmit:
## Change Summary

Re #2517 #2523
Cast text type for like condition on postgres (similar to old customKnex)

## Change type

image
image


pranavxc (approved):

RawmanGrey:
@mertmit, great thanks for bringing back dates filtering! 

Just tried on docker 'nocodb/nocodb-timely:0.92.0-pr-2531-20220701-1255' to test and found out it working nicely : )


May you please, if feature can be addressed with worthy effort, add an ability to filter (with gt, ge, lt, le, btw, nbtw operators) date-time fields by UTC timestamp, too? This can open doors to complex filtering, e.g. by edited last week, by past 24 hours, or by any date intervals... Delicious : )

Issue #807, opened 2021-11-24
Option to Prevent Download


doodlemania2:
## Recommending Features

Would be wonderful if a public shared view could be configured to not permit CSV export.

**Need for this feature**
I value the data I'm sharing and am happy to share it visibly. But would prefer to not have the data wander off.

**Is your feature request related to a problem? Please describe.**
Nope

**Describe the solution you'd like**
When publicly sharing (non-authenticated view), have the ability to turn off the export to CSV feature.

**Describe alternatives you've considered**
Different tool :)

**Additional context**
n/a

o1lab:
Sure, we will consider this :)

gemurdock:
Wouldn't it still be accessible via API? Not exactly sure how NocoDB works here. Have yet to figure out if GUI fetches data via API as well, but I think this could be an additional issue.

doodlemania2:
> Wouldn't it still be accessible via API? Not exactly sure how NocoDB works here. Have yet to figure out if GUI fetches data via API as well, but I think this could be an additional issue.

It's a reasonable question - in this instance, I don't need it to be impossible to copy down (one could also just grab screenshots or do table selects), but a CSV export on the shared page makes it a bit TOO easy :)

PR #2552, opened 2022-07-04
enhancement: allow disabling download csv on a shared view

wingkwong:
## Change Summary

- Add a boolean column `meta` in `nc_views_v2`. By default, `meta. allowCSVDownload` is `true` in order to cover the existing cases.
- Add a new option for users to choose if there is a download csv option in shared view (ref: #807)
- Render MoreActions based on the flag - Currently the only option in MoreActions is download csv
- Include Download Allowed column in Shared View List. Show '✔️ or ❌ for grid view only. N/A otherwise.
- Update related cypress tests

## Change type

## Test/ Verification

Enable Download (Default)

image

image

Disable Download

image

image

Shared View List

image


packages/nocodb/src/lib/migrations/v2/nc_018_add_download_in_view.ts
```diff
@@ -0,0 +1,38 @@
+import Knex from 'knex';
+import { MetaTable } from '../../utils/globals';
+
+const up = async (knex: Knex) => {
+  await knex.schema.alterTable(MetaTable.VIEWS, (table) => {
+    table.boolean('download').defaultTo(true);
```
pranavxc, line 6: What about keeping a meta(`text` type) column that holds a JSON string?  Extending view-specific metadata becomes more manageable that way we don't have to write more migrations.

pranavxc (approved):

dstala:
@wingkwong 
Locally also I could see behavior where on first click `share view`, modal did not open. I am assuming it to be with browser cache or some state maintained in the browser (just a guess).  

I have captured responses for the first click & subsequent click here
In response, meta was set to null in first & was set to `meta: "{\"allowCSVDownload\":true}"` in next

```
{"id":"vw_pgm7npenhzh557","base_id":"ds_n0tt53v2vtm3s3","project_id":"p_qyrmkywgv9qo0t","fk_model_id":"md_iclrlpb3hb406d","title":"Payment","type":3,"is_default":1,"show_system_fields":null,"lock_type":"collaborative","uuid":"6f2b9206-2858-460b-a611-4032b68fb399","password":null,"show":1,"order":1,"created_at":"2022-07-09 02:54:30","updated_at":"2022-07-09 02:54:30","meta":null}
```

```
{"id":"vw_pgm7npenhzh557","base_id":"ds_n0tt53v2vtm3s3","project_id":"p_qyrmkywgv9qo0t","fk_model_id":"md_iclrlpb3hb406d","title":"Payment","type":3,"is_default":1,"show_system_fields":null,"lock_type":"collaborative","uuid":"6f2b9206-2858-460b-a611-4032b68fb399","password":null,"show":1,"order":1,"created_at":"2022-07-09 02:54:30","updated_at":"2022-07-09 02:54:30","meta":"{\"allowCSVDownload\":true}"}
```

wingkwong:
@dstala Yes. Currently the behaviour is a bit changed. The options will be stored in `meta` and will be persisted, i.e. when the switcher is off, next time when you open the share view dialog, it's still off as it's reading the value from `meta`.

mertmit (commented):
In my opinion handling view meta's parse and stringify operations within view model is more appropriate(as in column meta). This way we can assume meta is an object (no need to parse or stringify every time)
Waiting your opinion on this @wingkwong

mertmit (approved):

PR #2613, opened 2022-07-08
Nuxt3 - Sidebar and view based data loading

pranavxc:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.


packages/nc-gui-v2/components/cell/Rating.vue
```diff
@@ -35,12 +35,13 @@ const localState = computed({
...
       <template #item="{ isFilled, click }">
-        <v-icon v-if="isFilled" :size="15" :color="ratingMeta.color" @click="click">
-          <MdiStarIcon />
-        </v-icon>
-        <v-icon v-else :color="ratingMeta.color" :size="15" class="nc-cell-hover-show" @click="click">
-          <MdiStarOutlineIcon />
-        </v-icon>
+        <!--        todo : custom color and icon -->
+        <!--        <v-icon v-if="isFilled"- :size="15" :color="ratingMeta.color" @click="click"> -->
+        <MdiStarIcon v-if="isFilled" @click="click" />
```
bcakmakoglu, line 40: To apply color to an icon you can use windicss -> `<MdiStarIcon class="text-primary" />`

Or in this case probably something like `<MdiStarIcon :class="`text-[${ratingMeta.color}]` />

packages/nc-gui-v2/components/virtual-cell/BelongsTo.vue
```diff
@@ -315,7 +315,7 @@ export default {
     <!--    <template v-if="!isForm"> -->
     <div class="chips d-flex align-center img-container flex-grow-1 hm-items">
       <template v-if="value || localState">
-        <ItemChip :active="active" :item="value" :value="value[primaryValueProp]" />
+        <ItemChip :active="active" :item="value" :model-value="value[primaryValueProp]" />
```
bcakmakoglu, line 318: should be v-model, shouldn't it? Or do you not want the value updated?

packages/nc-gui-v2/composables/useViewData.ts
```diff
@@ -72,7 +76,13 @@ export default (meta: Ref<TableType> | ComputedRef<TableType> | undefined) => {
       return o
     }, {})
 
-    const insertedData = await $api.dbTableRow.create('noco', project?.value.id as string, meta?.value.id as string, insertObj)
+    const insertedData = await $api.dbViewRow.create(
+      'noco',
```
bcakmakoglu, line 80: can we export the name as a const somewhere?

maybe in lib/constants or something, so we don't have to use the 'noco' string repeadetly.
Maybe something like
```ts
export const NocoName = 'noco' as const 
```

(as const forces the string to be 'noco' not 'string', only for typescript though)

packages/nc-gui-v2/composables/useViews.ts
```diff
@@ -0,0 +1,15 @@
+import type { FormType, GalleryType, GridType, KanbanType, TableType } from 'nocodb-sdk'
+import type { Ref } from 'vue'
+import { useNuxtApp } from '#app'
+
+export default function (meta: Ref<TableType>) {
+  const views = ref<Array<GridType | FormType | KanbanType | GalleryType>>()
```
bcakmakoglu, line 6: You would do me a big favor by using the `[]` notation instead of `Array<>` 😄 

in TS we usually use `string[]` and not `Array<string>` even though both are the exact same.
In this case it would be
```ts
const views = ref<(GridType | FormType | KanbanType | GalleryType)[]>()
```
meaning each element of the array can be a GridType, FormType etc.

or 
```ts
const views = ref<GridType[] | FormType[] ...>()
```

which would mean it's an array of __only__ GridType elements etc.

packages/nc-gui-v2/package.json
```diff
@@ -14,6 +14,7 @@
     "jwt-decode": "^3.1.2",
     "nocodb-sdk": "file:../nocodb-sdk",
     "socket.io-client": "^4.5.1",
+    "vite-plugin-vuetify": "^1.0.0-alpha.12",
```
bcakmakoglu, line 17: What's this for? :)

packages/nc-gui-v2/components/virtual-cell/BelongsTo.vue
pranavxc: just want to pass the primary value to display... edit will open an expanded form so may be I will switch it back to value, 2 way binding is not necessary here

packages/nc-gui-v2/composables/useViewData.ts
pranavxc: make sense I will keep a file

packages/nc-gui-v2/composables/useViews.ts
pranavxc: ya sure I will keep it in that format from now onwards.... any specific reason to follow that pattern?

packages/nc-gui-v2/package.json
pranavxc: Sorry, I forgot to remove it... will remove it

packages/nc-gui-v2/composables/useViews.ts
bcakmakoglu: No particular reason 😅 its just the usual convention

bcakmakoglu (approved):

Issue #2565, opened 2022-07-05
[Feature] Export XLSX file


LouisDelbosc:
**Need for this feature**
You can do CSV export but some of my users are not very familiar with CSV. I'd like to enable XLSX export file.

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. 

**Describe the solution you'd like**
I'd like an action bellow or above CSV export to have XLSX export.

PR #2658, opened 2022-07-12
Feat/export xlsx

LouisDelbosc:
## Change Summary

Add export xlsx button and on nc-gui.
Here the [issue#2565](https://github.com/nocodb/nocodb/issues/2565)

## Change type

## Test/ Verification

- Add new routes to receives csv raw data (instead of files)
- Add option on `More` button to have an xlsx export
- Modify sdk for the client to handle the new export type

Screenshot 2022-07-12 at 17 23 53
Screenshot 2022-07-12 at 17 24 09

I decide to have the file construction done by frontend because it was simpler for the backend logic (not duplicate or refactor convoluted code which can be fragile).
If you would like another way of doing things, let me know and I'd gladly implement them.

### DIFF
--- a/packages/nc-gui/components/project/spreadsheet/components/MoreActions.vue
+++ b/packages/nc-gui/components/project/spreadsheet/components/MoreActions.vue
@@ -28,6 +28,23 @@
             </span>
           </v-list-item-title>
         </v-list-item>
+
+        <v-list-item
+          v-t="['a:actions:download-xlsx']"
+          dense
+          @click="exportXlsx"
+        >
+          <v-list-item-title>
+            <v-icon small class="mr-1">
+              mdi-download-outline
+            </v-icon>
+            <span class="caption">
+              <!-- Download as XLSX -->
+              {{ $t('activity.downloadToto') }} XSLX
+            </span>
+          </v-list-item-title>
+        </v-list-item>
+
         <v-list-item
           v-if="_isUIAllowed('csvImport') && !isView"
           v-t="['a:actions:upload-csv']"
@@ -136,7 +153,6 @@ export default {
 
       reader.readAsText(file);
     },
-
     async extractCsvData() {
       return Promise.all(
         this.data.map(async r => {
@@ -220,6 +236,13 @@ export default {
         })
       );
     },
+    async exportXlsx() {
+      console.log('export')
+      console.log({title: this.meta.title, otherTitle: this.selectedView.title})
+      console.log({meta: this.meta.title, selectedview: this.selectedView})
+      const res = await this.$api.dbViewRow.export('noco', this.projectName, this.meta.title, this.selectedView.title, 'xlsx')
+      return Promise.resolve('Success').then(() => console.log('res', res))
+    },
     async exportCsv() {
       let offset = 0;
       let c = 1;
@@ -228,6 +251,8 @@ export default {
         while (!isNaN(offset) && offset > -1) {
           let res;
           if (this.publicViewId) {
+            console.log('publicViewId')
+            console.log(publicViewId)
             res = await this.$api.public.csvExport(this.publicViewId, ExportTypes.CSV, {
               responseType: 'blob',
               query: {
@@ -251,6 +276,7 @@ export default {
               },
             });
           } else {
+            console.log('else')
             res = await this.$api.dbViewRow.export(
               'noco',
               this.projectName,
--- a/packages/nc-gui/components/project/spreadsheet/components/MoreActions.vue
+++ b/packages/nc-gui/components/project/spreadsheet/components/MoreActions.vue
@@ -40,7 +40,7 @@
             </v-icon>
             <span class="caption">
               <!-- Download as XLSX -->
-              {{ $t('activity.downloadToto') }} XSLX
+              {{ $t('activity.downloadXLSX') }}
             </span>
           </v-list-item-title>
         </v-list-item>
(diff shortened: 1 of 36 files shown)


o1lab:
Thanks for the PR - appreciate it.

However, expectation of export to excel here can mean export the entire database. 
Since we are already supporting csv export - importing that csv to excel is a straightforward process.
Please get in touch with us before taking a feature implementation in discord.

LouisDelbosc:
> However, expectation of export to excel here can mean export the entire database.
Since we are already supporting csv export - importing that csv to excel is a straightforward process.
Please get in touch with us before taking a feature implementation in discord.

Thanks for the answer. I have numerous users (and potential users) on my nocodb instance who don't know how to import a CSV on excel.
For you and me, it is a straightforward process, but since nocodb want to be a *no code* solution and since I have this issue with my users and the many more who will follow, I thought it would be a good idea to have it here.

I'm open for any critics and way to solves this problem. If you know a way to do it without having to contribute, I'd be very pleased.

packages/nocodb/src/lib/utils/projectAcl.ts
```diff
@@ -26,6 +26,7 @@ export default {
     dataGroupBy: true,
     commentsCount: true,
     exportCsv: true,
+    exportCsvData: true,
```
wingkwong, line 29: What's difference with between `exportCsv` and `exportCsvData`? Why not using `exportExcel`?

packages/nocodb-sdk/src/lib/globals.ts
```diff
@@ -13,6 +13,7 @@ export enum RelationTypes {
 
 export enum ExportTypes {
   CSV = 'csv',
+  CSV_DATA = 'csvData',
```
wingkwong, line 16: Should it be `EXCEL`?

packages/nocodb-sdk/src/lib/Api.ts
```diff
@@ -2554,7 +2554,7 @@ export class Api<
       orgs: string,
       projectName: string,
       tableName: string,
-      type: 'csv' | 'excel',
+      type: 'csv' | 'excel' | 'csvData',
```
wingkwong, line 2557: Why adding a new one instead of using existing `excel`?

packages/nocodb-sdk/src/lib/Api.ts
```diff
@@ -2917,7 +2917,7 @@ export class Api<
       projectName: string,
       tableName: string,
       viewName: string,
-      type: 'csv' | 'excel',
+      type: 'csv' | 'excel' | 'xlsx',
```
wingkwong, line 2920: When to use `xlsx` and when to use `csvData`?

packages/nocodb/src/lib/meta/api/exportApis.ts
```diff
@@ -18,6 +18,19 @@ async function exportCsv(req: Request, res: Response) {
   res.send(data);
 }
 
+async function exportCsvData(req: Request, res: Response) {
```
wingkwong, line 21: probably make a generic function, i.e. combining into `exportCsv`

packages/nocodb/src/lib/meta/api/dataApis/dataAliasExportApis.ts
```diff
@@ -39,4 +56,10 @@ router.get(
   ncMetaAclMw(csvDataExport, 'exportCsv')
 );
 
+router.get(
+  '/api/v1/db/data/:orgs/:projectName/:tableName/views/:viewName/export/csvData',
```
wingkwong, line 60: naming - similar to other comments.

wingkwong (commented):
Other comments.

1. Extra empty cells
image

2. Discrepancy between exported data (See LastUpdated)
image

packages/nocodb/src/lib/utils/projectAcl.ts
LouisDelbosc: The xlsx file is created on the frontend. The backend send data as CSV (just data, not file) and the frontend use this data to build a XLSX file.
That's why the endpoint I created is not `excel` or `xlsx` because the backend doesn't know about all theses things, it only sends CSV file or CSV data.

packages/nocodb-sdk/src/lib/globals.ts
LouisDelbosc: As I said, the backend doesn't build the excel file. But maybe it'd be better to do so ?

packages/nocodb-sdk/src/lib/Api.ts
LouisDelbosc: Because I made the backend send data instead of file.

yohanboniface:
> I did not see that beforehand thank you. After looking for information, it seems excel store datetime using numbers, I check on numbers or google spreadsheet and it doesn't seems to works as wanted. I need a real excel to check if we have a datetime reading this information. Otherwise, i'm quite disappointed :(

Are you setting the right cell type ?

packages/nc-gui/components/project/spreadsheet/components/MoreActions.vue
```diff
@@ -220,6 +237,58 @@ export default {
...
+                      direction,
+                      fk_column_id,
+                    }))
+                ),
+                filterArrJson: JSON.stringify(this.reqPayload && this.reqPayload.filters),
+              },
+              headers: {
+                'xc-password': this.reqPayload && this.reqPayload.password,
+              },
+            });
```
mertmit, line 267: You should handle shared view case as well. Nothing exported + no corresponding endpoint on backend.

packages/nc-gui/components/project/spreadsheet/components/MoreActions.vue
```diff
@@ -220,6 +237,58 @@ export default {
...
+          } else {
+            res = await this.$api.dbViewRow.export(
+              'noco',
+              this.projectName,
+              this.meta.title,
+              this.selectedView.title,
+              ExportTypes.CSV_DATA
+            )
+            const { data } = res
+            const xlsx = XLSX.read(data.csvData, { type: 'string' })
```
mertmit, line 277: Seems like you can use `raw: true` to avoid unwanted processing of dates, also looks like it fixes extra empty cells.

PR #2670, opened 2022-07-13
feat: enable grouped filter and some css corrections

pranavxc:
## Change Summary

- Enabled support for grouped filter
- CSS corrections 

## Change type

## Test/ Verification

Equivalent to => `where FirstName like '%race%' OR (FirstName like '%vi%' and like '%k%')`

Screenshot 2022-07-13 at 6 06 40 PM

### DIFF
--- a/packages/nc-gui/components/project/spreadsheet/components/ColumnFilter.vue
+++ b/packages/nc-gui/components/project/spreadsheet/components/ColumnFilter.vue
@@ -7,14 +7,12 @@
             <div class="d-flex" style="gap: 6px; padding: 0 6px">
               <v-icon
                 v-if="!filter.readOnly"
-                :key="i + '_3'"
                 small
                 class="nc-filter-item-remove-btn"
                 @click.stop="deleteFilter(filter, i)"
               >
                 mdi-close-box
               </v-icon>
-              <span v-else :key="i + '_1'" />
               <v-select
                 v-model="filter.logical_op"
                 class="flex-shrink-1 flex-grow-0 elevation-0 caption"
@@ -50,7 +48,7 @@
           <template v-else>
             <v-icon
               v-if="!filter.readOnly"
-              :key="i + '_3'"
+              :key="i + '_1'"
               small
               class="nc-filter-item-remove-btn"
               @click.stop="deleteFilter(filter, i)"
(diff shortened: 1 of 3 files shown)


wingkwong (commented):
1. Just a reminder. There is `npm run fix:prettier` in `nc-gui` now. I helped you run it already.

2. Is Filter Group only under a specific Filter? If so, would it be better to show + Add Filter Group when there is at least one Filter added? Not sure if the below case is expected or not.

  - Create several Filter Groups
  
    image
  
  - Add a Filter within the first Filter Group
  
    image
  
  - Apply Changes -> A new Filter is created outside the Filter Group
  
    image

3. Duplicate keys detected just by creating a Filter
image

4. Not sure how to reproduce but when I tried adding multiple Filter and Filter Groups, at some point, it hung up. 

5. Just an idea. Should we take the set of the Filters? As `Title = A` happens 3 times and `Title2 = B` happens 2 times, the result is same as `Title = A` (1 time) && `Title2 = B` (1 time). 

image

pranavxc:
> 1. Just a reminder. There is `npm run fix:prettier` in `nc-gui` now. I helped you run it already.
> 2. Is Filter Group only under a specific Filter? If so, would it be better to show + Add Filter Group when there is at least one Filter added? Not sure if the below case is expected or not.
> 
> * Create several Filter Groups
>   image
> * Add a Filter within the first Filter Group
>   image
> * Apply Changes -> A new Filter is created outside the Filter Group
>   image
> 
> 3. Duplicate keys detected just by creating a Filter
>    image
> 4. Not sure how to reproduce but when I tried adding multiple Filter and Filter Groups, at some point, it hung up.
> 5. Just an idea. Should we take the set of the Filters? As `Title = A` happens 3 times and `Title2 = B` happens 2 times, the result is same as `Title = A` (1 time) && `Title2 = B` (1 time).
> 
> image

1. Thanks for pointing out :+1:

2. The above case is expected, a group can be at the beginning as well, but the `and / or` dropdown is not required since it doesn't make sense will update that part.
3. I will handle the duplicate key issue.
4. It Could be related to a duplicate key that I will try to reproduce

wingkwong (approved):

PR #2837, opened 2022-07-26
feat(gui-v2): add views sidebar

bcakmakoglu:
## Change Summary

* Implement views sidebar in project view

## Change type

## Test/ Verification

Provide summary of changes.


dstala:
- [ ] scroll needs fix. bottom menu options should persist. create view options should move above bottom menu options on scroll
https://loom.com/share/9311a383d56645e58ac0a674d3965fdc
- [ ] primary view delete allowed; should not be allowed (as in v1)
- [ ] delete view confirmation message required? if we intend to retain
  - [ ] bring pop-up to the center of page
  - [ ] i18n update 
- [ ] `API snipped`, `Hiring`, `Webhooks` buttons should be frozen at the bottom
  
Screenshot 2022-07-27 at 11 12 20 PM

- [ ] click on secondary view- two tabs open & view menu disappears; ignore if its WIP item
  
Screenshot 2022-07-27 at 11 46 09 PM

- [ ] migration pending
  - [ ] edit view
  - [ ] drag re-order
  - [ ] tool tip
  
Screenshot 2022-07-27 at 11 22 58 PM

bcakmakoglu:
Without sample data my view looks much different 😆 
I really need to connect a sample db to have the same experience.

bcakmakoglu:
The dblclick name edit was a suggestion by Naveen. 
A validation is missing though, which I will add now.

bcakmakoglu:
I can't reproduce the renaming issue, can you check if the issue persists or explain what exactly you clicked/pressed, I did what i saw in the video and didn't get any error.

dstala:
[Observations post rebase to https://github.com/nocodb/nocodb/pull/2837/commits/4410408b6cbc91bd9554057d7c1b07049127bd69]

- [x] Liked the way you have made `view create` menu fixed at the bottom
- [ ] Copy view option is also omitted for primary view, along with delete view. We should persist with copy view option
- [ ] Side bar height should be fixed to window size. Currently we have a scroll even when no views are created. might be because of spreadsheet height, not sure.
- [ ] Default view, on click- opens another tab. 
- [ ] Side bar toggle isn't working
- [x] Add little pad space after side bar toggle button 
- [ ] Rename issue persists still.
  - Double click on view name
  - Modify name
  - Press `Enter`
- [ ] Delete view pop-up, accept `Enter` key as `Yes`.

bcakmakoglu:
Where's is the "Sidebar" toggle supposed to be?
I don't see any in the old design either.

The popconfirm can't be brought into the middle. Either we switch to a full on modal (with overlay darkening the backdrop) or the popconfirm stays at the element that triggers it.

Still can't reproduce the renaming problem - can someone else confirm that the issue exists with them as well? What browser were you using? Any other steps?

Issues regarding `Tabs` are out of my hand 😄 
They're being developed in another PR if I'm not mistaken

bcakmakoglu:
Ah I just realized that the Sidebar toggle is that small door icon. Sort of confusing because that much more indicates a _sign out_ action than a sidebar toggle 😅

o1lab:
@pranavxc : Have edited Raju's comment and made a mention of you. See if you could discuss with @bcakmakoglu.

bcakmakoglu:
Regarding drag sort: Pranav is implementing the drag-sort feature in the left-side treeview.
I would just wait for him to finish it and re-use the implementation for this sidebar, instead of doing the same work twice now and possibly ending up with differing solutions.

dstala:
> Still can't reproduce the renaming problem - can someone else confirm that the issue exists with them as well? What browser were you using? Any other steps?
  
SQLite root DB, XCDB project. Chrome. Will re-install node_modules and try once

bcakmakoglu:
@dstala It was apparently a browser-related issue where across browsers the blur event was triggered differently.
Should work correctly now.

dstala:
> @dstala It was apparently a browser-related issue where across browsers the blur event was triggered differently. Should work correctly now.

Verified

dstala:
> The popconfirm can't be brought into the middle. Either we switch to a full on modal (with overlay darkening the backdrop) or the popconfirm stays at the element that triggers it.
  
Would be good IMO if we keep it uniform for table delete/view delete. I will introduce delete view pop up message in i18n

bcakmakoglu:
@dstala Should I use a modal for delete verificiation now?

o1lab:
Let's keep the modal for delete - same everywhere. Like in middle of screen (a bit boring). Later we can optimise - like in current implementation where it gets shown near where the action happens.

bcakmakoglu:
@dstala Drag sorting and delete modal are now implemented (as well as some general refactoring in the code to simplify and split up the code, so please make sure everything works as expected 😅)

dstala:
@bcakmakoglu 

- [ ] Liked the way you have made the view icon & drag icon mutually exclusive. Neat. Can we bring in a similar concept for table/SqlView drag && fields drag? 
- [ ] View re-order isn't retained across page reload. Check [here](https://loom.com/share/eca8b64cc52b4817a429244d37d9324e)
- [ ] Delete modal is good! I tried to add `keydown.enter="onDelete"`; did not help. Can you help fix this? I prefer all modals to consume the `escape` key as cancel & `enter` key as accept/ok. 
- [ ] Default view names for grids are now `Grid1`, `Grid2` and so on. This deviates from our earlier names derived out of table names. Is this by design?
- [ ] Toast design is not uniform. Intended that way for different categories? I prefer the one we have in this PR if we plan to keep one. 

Screenshot 2022-07-30 at 12 13 23 AM
Screenshot 2022-07-30 at 12 13 46 AM

bcakmakoglu:
@dstala
1. Yeah sure, we can do that - but I'd put that into a new issue / PR as it's not in the scope of the view sidebar
2. The issue should now be fixed (hopefully 😅) but I'd like to propose that we change the way sorting works in the future
	* Currently I have to pass an `order` value for a single item as an argument to the view update
	* This requires that we do some hacky calculations to make sure the ordering works without having to update all other rows sort order value, which makes this not only tedious in the frontend but imo very error prone because we can't really know the correct sort order by looking at it's value. They're not 1,2,3,4,5 .... n but 1,2, 3, 3.5, 4, 4.5 ... n 
	* Instead I would like to send the api an array of sorted id's that represent the current sort order of views, e.g. `api.dbView.updateOrder(['id1', 'id2' ... 'idn'])`
3. Done, keypress is now enabled on creating a view and deleting a view modals. They're also slightly more centered on the screen (though this is a deviation from ant designs default)
4. No particular reason. I can change it, though I feel that using `Grid-1` etc. as a name suggestion for a `Grid` type view works well - it indicates what sort of view is being used 😄 
5. Tab navigation is not yet fully functional as far as I know - so that will of course be fixed but it's not part of the views sidebar; It should be handled in a separate issue + PR 
6. I used the Ant design notifications, whereas the others have used `vue-toastification` which was the previous toast provider (in GUI v1). In my opinion we should stick with the ant design one as it already fits the existing design (+ one less dependency) - it would be good to open another issue for this; I can go through the code and just exchange vue-toastification with an ant-design notification.
When and where toasts are supposed to be sent, we can of course discuss :) It's not a big task to add/remove them.

dstala:
- [ ] reorder issue still persists. order is not retained if we close and re-open the tab (or close and re-open the project). 
- [x] delete view modal & `keydown.enter` worked fine.

bcakmakoglu:
@dstala Reordering should work now

PR #2899, opened 2022-07-30
refactor(gui-v2): Added DateTime Picker

mustafapc19:
Ref: #2891 

UI:
image
image


bcakmakoglu (approved):

dstala:
0730 1900
@mustafapc19 
- [ ] Remove option `x` and calendar icon are mutually exclusive. I ended up accidentally clearing off contents while trying to open date-time modal by clicking on the calendar icon. Is this default behavior of ant view or can be overridden?
- [ ] possible to add a header on top of HH & MM columns - would add more clarity. if its just a standard thing, we can stick to it though. just my opinion
- [ ] press space bar after clicking the date-time cell. it scrolls down spreadsheet page. can we ignore any key stroke after clicking this cell?

Issue #2927, opened 2022-08-01
[Data types] duration, currency, URL, formula, email


dstala:
- common
  - [ ] font size are not uniform. ignore if this will be taken up during spread sheet work;
  
![Screenshot 2022-08-01 at 3 49 31 PM](https://user-images.githubusercontent.com/86527202/182127763-9feb0271-27eb-4bb4-b186-4941b936e71f.png)

  - [ ] remove blue border for edit
  
![Screenshot 2022-08-01 at 3 50 55 PM](https://user-images.githubusercontent.com/86527202/182128047-9bb8b3b1-4c96-45ec-b3d3-a9500616c2f6.png)

- duration
- currency
- URL
  - [ ] display URL in blue link as in v1? v1 image appended here
  
![Screenshot 2022-08-01 at 3 55 27 PM](https://user-images.githubusercontent.com/86527202/182128767-f1a8e7cc-d4ab-4fa1-abac-141acba5a0b7.png)
 
- formula
- email
  - [ ] font size should remain size when we change to edit mode. currently, it shrinks.

dstala:
> on editing a cell part of the formula, the cell value in the formula column didn't update. ignore if this will be picked up during spreadsheet activity

this will be picked up as a new issue

PR #2930, opened 2022-08-01
fix(gui-v2): cell follow ups

wingkwong:
## Change Summary

ref: #2927

common - font size are not uniform

- Will revise in spreadsheet PR

display text was highlighted in blue in v1 for currency cell. is it by design to mark everything uniform in v2?

- Checked with Mert a month ago. It's not expected to show in Blue. In v2, they will display normally.

URL  - validation failure case, no pop-up error displayed

- in v1, i don't see a pop-up error

Email - validation failure case, no pop-up error displayed

- in v1, I don't see pop-up error. Current logic s that if it s not a valid email, it renders as a text instead of link with mailto

On editing a cell part of the formula, the cell value in the formula column didn't update. ignore if this will be picked up during spreadsheet activity

- I don't see there is an option to change the formula now. Will come back when it's ready.

## Change type

### DIFF
--- a/packages/nc-gui-v2/components/cell/Email.vue
+++ b/packages/nc-gui-v2/components/cell/Email.vue
@@ -2,40 +2,36 @@
 import { computed } from '#imports'
 import { isEmail } from '~/utils'
 
-const { modelValue: value } = defineProps<Props>()
-
-const emit = defineEmits(['update:modelValue'])
-
-const editEnabled = inject<boolean>('editEnabled')
-
 interface Props {
   modelValue: string
 }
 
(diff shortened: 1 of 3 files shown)


dstala:
@wingkwong 

- for the validation scenario pop-up, my bad - hadn't specified that I had enabled validation explicitly. below is the pop-up (toast) message on v1. As Pranav mentioned, API errors will be consumed as part of the spreadsheet. so we can defer.
  
Screenshot 2022-08-01 at 8 49 11 PM (2)

dstala (approved):
verified changes locally

Issue #2921, opened 2022-08-01
vue3: TimePicker Cell


mustafapc19:

dstala:
- Invalid time displayed for a valid time configured in v1
![Screenshot 2022-08-01 at 7 12 11 PM](https://user-images.githubusercontent.com/86527202/182161334-de8374ac-cd02-4c77-83eb-5e448897216f.png)

bcakmakoglu:
Haven't checked but if the code is ok maybe the plugins aren't installed because people import dayjs from 'dayjs' and not from utils 😅 
Relevant PR: #2933

mustafapc19:
@dstala @bcakmakoglu Seems to be the issue with dayjs

mustafapc19:
@dstala We can use 24hours format as well. 12hr looked better so I chose that. I am fine with both ways. I personally find 24 hours harder to calculate 😅

dstala:
@mustafapc19 

pull latest changes - could still see `invalid time`
- could see entered values being stored in DB (MySQL)
- spreadsheet cell, it's still showing `invalid time`

PR #2939, opened 2022-08-01
fix/gui-v2-fixed-issue-with-time-picker

mustafapc19:
Ref: [Comment](https://github.com/nocodb/nocodb/issues/2921#issuecomment-1201272932)
Fixes issue where `TimePicker` bugs out with mysql

### DIFF
--- a/packages/nc-gui-v2/components/cell/TimePicker.vue
+++ b/packages/nc-gui-v2/components/cell/TimePicker.vue
@@ -22,13 +22,20 @@ const localState = $computed({
     if (!modelValue) {
       return undefined
     }
+    let dateTime = dayjs(modelValue)
 
-    if (!dayjs(modelValue).isValid()) {
+    if (!dateTime.isValid()) {
+      dateTime = dayjs(modelValue, 'HH:mm:ss')
+    }
+    if (!dateTime.isValid()) {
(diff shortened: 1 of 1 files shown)


packages/nc-gui-v2/components/cell/TimePicker.vue
```diff
@@ -22,13 +22,20 @@ const localState = $computed({
...
+    if (!dateTime.isValid()) {
+      dateTime = dayjs(`1999-01-01 ${modelValue}`)
+    }
+    if (!dateTime.isValid()) {
       isTimeInvalid = true
       return undefined
     }
 
-    return dayjs(modelValue)
+    return dateTime
```
wingkwong, line 38: What about `dateTime.format('HH:mm:ss')` ?

wingkwong:
Please rebase & run lint again

packages/nc-gui-v2/components/cell/TimePicker.vue
mustafapc19: @wingkwong Ant date picker expects a dayjs object and takes a date format too. So this is taken care of.

wingkwong: Oh ok then.

wingkwong (approved):

Issue #2923, opened 2022-08-01
vue3: JSON cell


mustafapc19:

dstala:
@mustafapc19 

- [ ] Monaco editor dropped?
- [ ] Error message reporting frequency is quite high - currently is on every keystroke. Let us have it on `enter` key or `change of focus` (moving away from cell)
  
![Screenshot 2022-08-08 at 4 54 29 PM](https://user-images.githubusercontent.com/86527202/183407450-1850d430-d03d-4bfe-b31d-873299bc8396.png)
  
- [ ] Edit cell needs fix
  
![Screenshot 2022-08-08 at 4 56 35 PM](https://user-images.githubusercontent.com/86527202/183407846-10315cf5-c96c-42e0-a6a9-1e01812927d9.png)

Verified
- [x] Insert & Clear cell
- [x] Sort & Filter

mustafapc19:
@dstala I don't see issue from my side. We are still using Monaco editor

PR #2946, opened 2022-08-02
vue3: Added `Json` cell

mustafapc19:
Ref: #2923 

@wingkwong Have refactored `Editor` component. Have tested it with `Json` cell and `JSON File` of quick import and seems to work fine. Can you verify if the changes are fine since there is a chance that I might miss on some edge case.

Issue not solved:
- Clicking on cancel doesn't toggle the edit mode off as that is out of scope of this PR

UI:
image
image
image
image
image


wingkwong (changes requested):
Create a valid JSON record in V1. Open in V2. See the error.

image

mustafapc19:
@wingkwong I can't seem to reproduce it 
image

I created a json in v1 and seems work fine in v2

mustafapc19:
@wingkwong The issue must be solved now

wingkwong (changes requested):
The previous issue has been resolved. Here's the new one.

In V1, create a new row as below. Open in V2.

image

Click `Save` - you can keep clicking to see below behaviour. 

image

Expected behaviours:

- once save is clicked, it should save and exit edit mode
- the content shouldn't be stringified.

mustafapc19:
@wingkwong The root issue for this series problems that we ran into here, is that the `modelValue` first given to this component is a `json` object but when we have to update the `modelValue`, we have to send the string representation of the json object other wise backend will error out since backend is only handling the case where the given data is of the string form. 

And switching the edit mode is not in the scope of this PR. We are changing the value of the `editEnabled` but it is not being updated. Pranav said to ignore it for now.

mustafapc19:
@wingkwong Have solved the issue and have tested it as best I could. Please test it again. There is a good chance there might be more issues

One issue I found was regarding a empty column, where when we open an empty column we get 
image
 which is saveable as (null is a valid json). But when we save it saves the cell with "null" and cell shows 
image

This behaviour is also in v1.

wingkwong:
Depending on https://github.com/nocodb/nocodb/pull/2992. Need rebase after that PR is merged.

wingkwong:
Now I can't even see the value orz

v1:
image


v2:
image

mustafapc19:
@wingkwong It seems to be working alright for me.

wingkwong:
ok now. seems cached

wingkwong (approved):

wingkwong:
In our editor, there is a function called `format`. I think it'd be nice to have a button to format the json data.

mustafapc19:
@dstala True. I think this was not an issue before. I guess recently pressing enter to save the cell has been merged. Lets see if we can override that behaviour.

mustafapc19:
@dstala I have disabled `enter` to save for json cells and now json cell will format on save as well.

wingkwong (approved):

pranavxc (approved):

packages/nc-gui-v2/components/cell/Json.vue
```diff
@@ -0,0 +1,132 @@
...
+      JSON.parse(val)
+      error = undefined
+    } catch (e: any) {
+      error = e
+    }
+  },
+)
+
+watch(
+  () => editEnabled,
```
bcakmakoglu, line 85: instead of `() => editEnabled` you can also just pass `editEnabled` to the watcher.
Watchers accept refs/computeds as a watch source, doesn't have to be a function getter.

packages/nc-gui-v2/components/cell/Json.vue
```diff
@@ -0,0 +1,132 @@
...
+      <div class="flex flex-row justify-between pt-1 pb-2">
+        <a-button type="text" size="small" @click="isExpanded = !isExpanded">
+          <FullScreenExitIcon v-if="isExpanded" class="h-2.5" />
+          <FullScreenIcon v-else class="h-2.5" />
+        </a-button>
+        <div class="flex flex-row">
+          <a-button type="text" size="small" :onclick="clear"><div class="text-xs">Cancel</div></a-button>
+          <a-button type="primary" size="small" :disabled="!!error || localValue === vModel"
+            ><div class="text-xs" :onclick="onSave">Save</div></a-button
+          >
```
bcakmakoglu, line 105: broken formatting

bcakmakoglu (changes requested):

packages/nc-gui-v2/components/cell/Json.vue
wingkwong: This one is caused by lint. Will move to Misc issue at this moment.

bcakmakoglu (approved):

PR #2963, opened 2022-08-02
feat(gui-v2): currency column

wingkwong:
## Change Summary

ref: #2951 

- add `CurrencyOptions.vue` with validatiors
- add `CurrencyUtils.ts`
- include `CurrencyOptions.vue` in `EditOrAdd.vue`
- handle Pg with `money` case

## Change type

## Test/ Verification

image |image
|---|---|

image|image
|---|---|


packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
```diff
@@ -0,0 +1,102 @@
+<script setup lang="ts">
+import { useColumnCreateStoreOrThrow, useProject } from '#imports'
+import { currencyCodes, currencyLocales, validateCurrencyCode, validateCurrencyLocale } from '@/utils/currencyUtils'
```
pranavxc, line 3: export `currencyUtils` from `utils/index.ts` and replace the above line with 

```
import { currencyCodes, currencyLocales, validateCurrencyCode, validateCurrencyLocale } from '@/utils'
```

pranavxc (approved):

packages/nc-gui-v2/components/dlg/AirtableImport.vue
```diff
@@ -336,7 +336,8 @@ onBeforeUnmount(() => {
           </div>
         </a-card>
         <div class="flex justify-center items-center">
-          <a-button v-if="showGoToDashboardButton" class="mt-4" size="large" @click="dialogShow = false">Go to Dashboard</a-button
+          <a-button v-if="showGoToDashboardButton" class="mt-4" size="large" @click="dialogShow = false"
+            >Go to Dashboard</a-button
           >
```
bcakmakoglu, line 341: Broken formatting
`>` shouldn't be by themselves (even though eslint accepts this, it looks bad)

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
```diff
@@ -0,0 +1,102 @@
...
+import { currencyCodes, currencyLocales, validateCurrencyCode, validateCurrencyLocale } from '@/utils'
+
+interface Option {
+  label: string
+  value: string
+}
+
+const { formState, validateInfos, setAdditionalValidations, sqlUi, onDataTypeChange, onAlter } = useColumnCreateStoreOrThrow()
+
+const validators = computed(() => {
```
bcakmakoglu, line 12: Unnecessary computed I assume
There seems to be no case, where the validators have to be re-computed, so a regular `const` should do the trick.

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
```diff
@@ -0,0 +1,102 @@
...
+})
+
+// set additional validations
+setAdditionalValidations({
+  ...validators.value,
+})
+
+const { isPg } = useProject()
+
+const currencyList = ref(currencyCodes)
```
bcakmakoglu, line 48: unnecessary ref - `currencyList` is never updated and thus will never cause a re-render of view, so a `ref` is not necessary.
As a rule: `ref` `reactive` `computed` are only useful if they're meant to reactively trigger a view (re-)render.
If they're going to be static, you can save some memory by not using a reactive value.

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
```diff
@@ -0,0 +1,102 @@
...
+
+const currencyLocaleList = ref(currencyLocales())
```
bcakmakoglu, line 50: Another unnecessary ref

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
```diff
@@ -0,0 +1,102 @@
...
+      <a-form-item v-bind="validateInfos['meta.currency_locale']" label="Currency Locale">
+        <a-select
+          v-model:value="formState.meta.currency_locale"
+          size="small"
+          class="w-52"
+          show-search
+          :filter-option="filterOption"
+          :disabled="isMoney && isPg"
+        >
+          <a-select-option v-for="(currencyLocale, i) in currencyLocaleList ?? []" :key="i" :value="currencyLocale.value">
```
bcakmakoglu, line 76: `for ... of ...` instead of `in`

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
bcakmakoglu: Also add `|| []` so it can be avoided in the `for ... of ...` loop in v-for down below.

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
bcakmakoglu: Also add || [] so it can be avoided in the for ... of ... loop in v-for down below.

packages/nc-gui-v2/components/smartsheet-column/CurrencyOptions.vue
```diff
@@ -0,0 +1,102 @@
...
+      <a-form-item v-bind="validateInfos['meta.currency_code']" label="Currency Code">
+        <a-select
+          v-model:value="formState.meta.currency_code"
+          class="w-52"
+          show-search
+          :filter-option="filterOption"
+          size="small"
+          :disabled="isMoney && isPg"
+        >
+          <a-select-option v-for="(currencyCode, i) in currencyList ?? []" :key="i" :value="currencyCode">
```
bcakmakoglu, line 92: see above (for ... of ...)

bcakmakoglu (changes requested):

packages/nc-gui-v2/components/dlg/AirtableImport.vue
wingkwong: Caused by `npm run lint`. May fix it later.

bcakmakoglu (approved):

Issue #2981, opened 2022-08-03
vue3: Rollup column


mustafapc19:

dstala:
@mustafapc19 

- [ ] Edit column for rollup wasn't supported in v1 (we were just allowing column name change in v1, but not rollup configuration). In v2, we have provisioned configuration change as well - it's good, but it doesn't work currently. Either we can revert it to be like v1 or fix it.
  
![Screenshot 2022-08-08 at 5 09 01 PM](https://user-images.githubusercontent.com/86527202/183410111-6fac65c4-4e68-48b3-8d0c-e02e281ec925.png)

- [ ] Not possible to edit only column name, without feeding existing configuration again
  
![Screenshot 2022-08-08 at 5 11 50 PM](https://user-images.githubusercontent.com/86527202/183410343-dc0d13da-a3b6-418c-9959-12153a4cbdb7.png)


Verified
- [x] Add rollup column
- [x] Rollup cell is not editable
- [x] Rollup cell clear not allowed

PR #2988, opened 2022-08-03
vue3: Added rollup column

mustafapc19:
Ref: #2981 

UI:
image
image
image
image
image
image

### DIFF
--- a/packages/nc-gui-v2/components/smartsheet-column/EditOrAdd.vue
+++ b/packages/nc-gui-v2/components/smartsheet-column/EditOrAdd.vue
@@ -1,5 +1,6 @@
 <script lang="ts" setup>
-import { UITypes } from 'nocodb-sdk'
+import { UITypes, isVirtualCol } from 'nocodb-sdk'
+import RollupOptions from './RollupOptions.vue'
 import { computed, inject, useColumnCreateStoreOrThrow, useMetas, watchEffect } from '#imports'
 import { MetaInj } from '~/context'
 import { uiTypes } from '~/utils/columnUtils'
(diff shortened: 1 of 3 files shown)


packages/nc-gui-v2/composables/useColumnCreateStore.ts
```diff
@@ -198,7 +198,8 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
       }
       onSuccess()
     } catch (e: any) {
-      toast.error(await extractSdkResponseErrorMsg(e))
+      const error = await extractSdkResponseErrorMsg(e)
+      if (error) toast.error(await extractSdkResponseErrorMsg(e))
```
wingkwong, line 202: `await extractSdkResponseErrorMsg(e)` must return something. I think L201 is not necessary and you can just use `toast.error(await extractSdkResponseErrorMsg(e))`

packages/nc-gui-v2/components/smartsheet-column/EditOrAdd.vue
```diff
@@ -90,16 +91,17 @@ watchEffect(() => {
...
-
-      <div>
-        <div
-          class="text-xs cursor-pointer text-grey nc-more-options my-2 flex align-center gap-1 justify-end"
-          @click="advancedOptions = !advancedOptions"
-        >
-          {{ advancedOptions ? $t('general.hideAll') : $t('general.showMore') }}
-          <component :is="advancedOptions ? MdiMinusIcon : MdiPlusIcon" />
-        </div>
+      <RollupOptions v-if="formState.uidt === UITypes.Rollup" />
```
wingkwong, line 94: why not just `SmartsheetColumnRollupOptions ` ?

packages/nc-gui-v2/components/smartsheet-column/EditOrAdd.vue
```diff
@@ -1,5 +1,6 @@
 <script lang="ts" setup>
-import { UITypes } from 'nocodb-sdk'
+import { UITypes, isVirtualCol } from 'nocodb-sdk'
+import RollupOptions from './RollupOptions.vue'
```
wingkwong, line 3: I think you don't need to import it. You can use `SmartsheetColumnRollupOptions` directly as it is imported in `components.d.ts`

packages/nc-gui-v2/composables/useColumnCreateStore.ts
mustafapc19: @wingkwong It is. In form validation error it raises an validation error which causes empty toastr to come up. Doing this will make only error handed by Noco sdk to come up.

wingkwong: Maybe better handle it in `extractSdkResponseErrorMsg` (Line 2)

wingkwong (approved):

pranavxc (approved):

packages/nc-gui-v2/composables/useColumnCreateStore.ts
mustafapc19: @wingkwong It will add extra complications to a function which is used in a lot of places. I think handling this way is better?

wingkwong:
@mustafapc19 please rebase & resolve conflict

PR #2992, opened 2022-08-03
fix: grid view corrections

pranavxc:
## Change Summary

- Cell active outline correction
- Enter key-based navigation
- Option to toggle edit enabled

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui-v2/components/smartsheet/Cell.vue
+++ b/packages/nc-gui-v2/components/smartsheet/Cell.vue
@@ -18,11 +18,21 @@ interface Emits {
 
 const { column, ...props } = defineProps<Props>()
 
-const emit = defineEmits(['update:modelValue', 'save', 'navigate'])
+const emit = defineEmits(['update:modelValue', 'save', 'navigate', 'cancel'])
 
 provide(ColumnInj, column)
 
-provide(EditModeInj, toRef(props, 'editEnabled'))
+provide(
+  EditModeInj,
+  computed({
+    get() {
+      return props?.editEnabled
+    },
+    set() {
+      return emit('cancel')
+    },
+  }),
+)
 
 let changed = $ref(false)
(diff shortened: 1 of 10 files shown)


packages/nc-gui-v2/components/smartsheet/Cell.vue
```diff
@@ -17,14 +18,25 @@ interface Emits {
...
+  EditModeInj,
+  computed({
+    get() {
+      return props?.editEnabled
+    },
+    set() {
+      return emit('cancel')
+    },
+  }),
+)
```
bcakmakoglu, line 35: Why not use another `vModel` instead of a regular computed prop? You could then just bind it as `<Cell v-model:edit-enabled="prop" />`

bcakmakoglu (changes requested):

packages/nc-gui-v2/components/smartsheet/Cell.vue
pranavxc: Ya should be fine, In parent I can do something like

```
js:edit-enabled="editEnabled && selected.col === colIndex && selected.row === rowIndex"
@update:edit-enabled="editEnabled  = false"
```

bcakmakoglu: If you use `v-model:edit-enabled` the `@update:edit-enabled` binding should be unnecessary
`v-model` automatically binds the `update` event

bcakmakoglu (approved):

packages/nc-gui-v2/components/smartsheet/Cell.vue
pranavxc: But there is an extra `&& selected.col === colIndex && selected.row === rowIndex` part

PR #3011, opened 2022-08-04
vue3: Added Date column

mustafapc19:
Ref: #3010 

UI:
image
image


packages/nc-gui-v2/components/smartsheet-column/DateOptions.vue
```diff
@@ -0,0 +1,25 @@
+<script setup lang="ts">
+import { useColumnCreateStoreOrThrow } from '#imports'
+import { dateFormats } from '~/utils/dateTimeUtils'
```
wingkwong, line 3: As we've `export * from './dateTimeUtils'` in `~/utils/index.ts`, we don't need to specify which utils file. Hence, just do it as below 

```js
import { dateFormats } from '~/utils'
```

wingkwong (approved):

Issue #2631, opened 2022-07-11
vue3 : Gallery


o1lab:

dstala:
@pranavxc @mustafapc19 

Observations; we can route appropriately if some of the items listed here were not under development scope of this PR

- [ ] Field `active` badge appears on creating Gallery view
  
![Screenshot 2022-08-09 at 9 26 58 AM](https://user-images.githubusercontent.com/86527202/183561123-91621012-cb87-45c7-ac90-7f1f2a5f56f0.png)

- [ ] `Expanded form` should pop on clicking card (Pranav?)
  
![Screenshot 2022-08-09 at 9 30 32 AM](https://user-images.githubusercontent.com/86527202/183561879-f15406d9-3f03-4a84-b32c-ec9bc38126c1.png)

- [ ] Mark checkbox cell in gallery view as read-only. Click on it should not toggle its value
  
![Screenshot 2022-08-09 at 9 38 45 AM](https://user-images.githubusercontent.com/86527202/183562237-47df2e2f-a621-4153-8bfd-9b0e0aecc2b1.png)
   
![Screenshot 2022-08-09 at 9 38 39 AM](https://user-images.githubusercontent.com/86527202/183562231-b5673f98-9c38-4292-9846-effe68a02115.png)

- [ ] Filter & sort need hard refresh / reload (Pranav?)
- [ ] Fields enable & disable 'menu' is not reactive. changes are reflected only after toggling menu (close and open)

PR #3021, opened 2022-08-05
vue3: Added gallery view

mustafapc19:
Ref: #2631 

Issues:
Known Scrolling has an offset 
image

UI:
image

### DIFF
--- a/packages/nc-gui-v2/components/smartsheet/Gallery.vue
+++ b/packages/nc-gui-v2/components/smartsheet/Gallery.vue
@@ -53,42 +53,45 @@ const isRowEmpty = (record: any, col: any) => {
 </script>
 
 <template>
-  <div class="grid grid-cols-4 gap-x-4 gap-y-4 pt-4 overflow-y-scroll h-full px-3">
-    <div v-for="(record, recordIndex) in data" :key="recordIndex" class="flex flex-col">
-      <a-card hoverable class="!rounded-lg h-full">
-        <template #cover>
-          <img v-if="record.row[coverImageColumn?.title]" class="h-52 rounded-t-lg" :src="coverUrl(record)" />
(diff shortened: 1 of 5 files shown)


o1lab:
Confirming if multiple images within a card mean - we can go through images at the card level?

packages/nc-gui-v2/components/smartsheet/Gallery.vue
```diff
@@ -12,235 +13,86 @@ const isPublicView = false
...
-                        :sql-ui="sqlUi"
-                        :is-locked="isLocked"
-                        class="xc-input body-2"
-                        :meta="meta"
-                      /> -->
-                <!--                </v-col> -->
-                <!--              </v-row> -->
+<!-- TODO: Fix scrolling -->
+<template>
+  <div class="flex flex-col h-100 min-h-0 w-100">
```
bcakmakoglu, line 57: don't use `h-100` or `w-100` - those are vuetify styles.
Use `h-full` and `w-full` - TailwindCSS classes.

packages/nc-gui-v2/components/smartsheet/Gallery.vue
```diff
@@ -12,235 +13,86 @@ const isPublicView = false
...
+    <div class="nc-gallery-container min-h-0 flex-1 grid grid-cols-4 gap-x-4 gap-y-4 my-4 px-3">
```
bcakmakoglu, line 58: `gap-x-4` and gap-y-4` are sort of redundant. Just use `gap-4`

bcakmakoglu (changes requested):

mustafapc19:
@o1lab No right now only the first image is shown as it is in v1. Should we add a carousal like flow to show multiple images?

mustafapc19:
@o1lab I missed the image carousal in v1. Will add it here as well 👍

PR #3022, opened 2022-08-05
feat(gui-v2): Single/multi select

mertmit:
## Change Summary

Re #2969 #2968
- Select options
- Single select cell
- Multi select cell

## Change type

## Test/ Verification
single1
image
image

### DIFF
--- a/packages/nc-gui-v2/components/cell/MultiSelect.vue
+++ b/packages/nc-gui-v2/components/cell/MultiSelect.vue
@@ -1,76 +1,97 @@
 <script lang="ts" setup>
+import type { SelectOptionType } from '~~/../nocodb-sdk/build/main/index.js'
 import { computed, inject } from '#imports'
 import { ColumnInj } from '~/context'
 
 interface Props {
-  modelValue: string | null
+  modelValue: string | string[] | undefined
 }
 
 const { modelValue } = defineProps<Props>()
 
 const emit = defineEmits(['update:modelValue'])
 
+const { isMysql } = useProject()
+
 const column = inject(ColumnInj)
 
-const options = computed(() => column?.value?.dtxp?.split(',').map((v) => v.replace(/\\'/g, "'").replace(/^'|'$/g, '')) || [])
+const options = computed(() => {
+  if (column?.colOptions) {
+    const opts = column.colOptions ? column.colOptions.options.filter((el: SelectOptionType) => el.title !== '') || [] : []
+    for (const op of opts.filter((el: SelectOptionType) => el.order === null)) {
+      op.title = op.title.replace(/^'/, '').replace(/'$/, '')
+    }
+    return opts
+  }
+  return []
+})
 
-const localState = computed({
-  get() {
-    return modelValue?.match(/(?:[^',]|\\')+(?='?(?:,|$))/g)?.map((v: string) => v.replace(/\\'/g, "'"))
-  },
-  set(val?: string[]) {
-    emit('update:modelValue', val?.filter((v) => options.value.includes(v)).join(','))
-  },
+const vModel = computed({
+  get: () =>
+    modelValue
+      ? typeof modelValue === 'string'
+        ? isMysql
+          ? modelValue.split(',').sort((a, b) => {
+              const opa = options.value.find((el: SelectOptionType) => el.title === a)
+              const opb = options.value.find((el: SelectOptionType) => el.title === b)
+              return opa.order - opb.order
+            })
+          : modelValue.split(',')
+        : modelValue
+      : [],
+  set: (val) => emit('update:modelValue', val.join(',')),
 })
 </script>
 
 <template>
-  <!--
-  <v-select
-    v-model="localState"
-    :items="options"
-    hide-details
-    :clearable="!column.rqd"
-    variation="outlined"
-    multiple
-  />
--->
-
-  <v-combobox
-    v-model="localState"
-    :items="options"
-    multiple
-    chips
-    flat
-    dense
-    solo
-    hide-details
-    deletable-chips
-    class="text-center mt-0"
+  <a-select
+    v-model:value="vModel"
+    mode="multiple"
+    class="w-full"
+    placeholder="Select an option"
+    :bordered="false"
+    show-arrow
+    :show-search="false"
   >
-    <!--    <template #selection="data"> -->
-    <!--      <v-chip -->
-    <!--        :key="data.item" -->
-    <!--        small -->
-    <!--        class="ma-1 " -->
-    <!--        :color="colors[setValues.indexOf(data.item) % colors.length]" -->
-    <!--        @click:close="data.parent.selectItem(data.item)" -->
-    <!--      > -->
-    <!--        {{ data.item }} -->
-    <!--      </v-chip> -->
-    <!--    </template> -->
-
-    <!--    <template #item="{item}"> -->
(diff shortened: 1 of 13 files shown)


o1lab:
Having a bit of curves on multiselect labels would make it look better.

wingkwong (commented):
Could you also replace the color picker in `CheckboxOptions` & `RatingOptions` with `GeneralColorPicker`?

mertmit:
Rebased & implemented the color picker, there is an ongoing issue with updating cell meta when options updated, I will work on that

dstala:
Observations

- [ ] Rename option; entry gets removed from table
- [ ] Close button overlap with `down arrow` is easy to miss
- [ ] Menu was not updated on rename operation until hard refresh
- [ ] No error displayed on trying to add empty option type (v1 allowed, discarded empty option)
- [ ] keyboard arrow key movement after selection of option cell not working
- [ ] allow key-press to be used to narrow down options

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -2,6 +2,7 @@
 import type { SelectOptionType } from '~~/../nocodb-sdk/build/main/index.js'
```
bcakmakoglu, line 2: Change import to
```ts
import type { SelectOptionType } from 'nobodb-sdk'
```

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -19,6 +20,8 @@ const editEnabled = inject(EditModeInj, ref(false))
 const active = inject(ActiveCellInj, ref(false))
 
 const selectedIds = ref<string[]>([])
+const aselect = ref<any>(null)
```
bcakmakoglu, line 23: Instead of using `any` type you could just use the `Select` component as a type.
```ts
import type { Select as AntSelect } from 'ant-design-vue'

const aSelect = ref<typeof AntSelect>()
```

Also the `null` initializer is unnecessary.

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -53,10 +56,36 @@ const selectedTitles = computed(() =>
     : [],
 )
 
+const handleKeys = (e: any) => {
```
bcakmakoglu, line 59: Should be `KeyboardEvent` instead of `any`

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -53,10 +56,36 @@ const selectedTitles = computed(() =>
...
+      break
+  }
+}
+
+const handleClose = (e: any) => {
+  if (aselect.value) {
+    const selectClick = aselect.value.$el.contains(e.target)
+    if (!selectClick) {
+      isOpen.value = false
+    }
```
bcakmakoglu, line 76: Can be shortened to 
```ts
if (aselect.value && !aselect.value.$el.contains(e.target)) {
	isOpen.value = false
}
```

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -53,10 +56,36 @@ const selectedTitles = computed(() =>
...
 onMounted(() => {
   selectedIds.value = selectedTitles.value.map((el) => {
     return options.value.find((op: SelectOptionType) => op.title === el).id
   })
+  document.addEventListener('click', handleClose)
+})
+
+onUnmounted(() => {
+  document.removeEventListener('click', handleClose)
 })
```
bcakmakoglu, line 89: Replace this with [useEventListener](https://vueuse.org/core/useeventlistener/)

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -67,19 +96,32 @@ watch(
...
 )
+
+watch(
+  () => isOpen.value,
+  (n, _o) => {
+    if (n === false) {
+      aselect.value.blur()
+    }
+  },
+)
```
bcakmakoglu, line 107: Watch doesn't need a getter function if you're going to watch a `ref` or `computed` value.
```ts
watch(isOpen, () => { ... })
```

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -53,10 +56,36 @@ const selectedTitles = computed(() =>
...
+      e.preventDefault()
+      isOpen.value = false
+      break
+    case 'Enter':
+      e.stopPropagation()
+      break
+  }
+}
+
+const handleClose = (e: any) => {
```
bcakmakoglu, line 71: use `MouseEvent` instead of `any`

packages/nc-gui-v2/components/cell/MultiSelect.vue
```diff
@@ -99,6 +143,28 @@ watch(
...
+  font-style: normal;
+  height: 12px;
+  line-height: 1;
+  text-align: center;
+  text-transform: none;
+  transition: color 0.3s ease, opacity 0.15s ease;
+  width: 12px;
+  z-index: 1;
+  margin-right: -6px;
+  margin-left: 3px;
```
bcakmakoglu, line 160: I'm okay with regular CSS but I'd rather see tailwind used instead, to keep it more consistent.

As an example:
```css
.ms-close-icon {
	@apply cursor-pointer text-sm leading-4 text-center;
	@apply flex w-[12px] h-[12px] z-1 ml-[3px] mr-[-6px];
	@apply before:block hover:text-white/45;
}
```

some stuff might not be possible of course, you can fall back to regular css there but I won't force you to change it 😄

packages/nc-gui-v2/components/cell/SingleSelect.vue
```diff
@@ -16,6 +16,9 @@ const isForm = inject<boolean>('isForm', false)
 const editEnabled = inject(EditModeInj, ref(false))
 const active = inject(ActiveCellInj, ref(false))
 
+const aselect = ref<any>(null)
```
bcakmakoglu, line 19: Same comments as in`MultiSelect.vue` basically - replace `any` types etc.

bcakmakoglu (changes requested):

dstala:
If items below are not under column scope, keep it open. I will re-test after integrating with develop

Open items
- [ ] Close button overlap with `down arrow` is easy to miss [still persists]
- [ ] Column rename: column data disappears till a hard refresh / reload
- [ ] Edit select data; skip saving (cancel). Changes persist till hard refresh / reload
- [ ] Remove option; Changes are not reflected till hard refresh / reload 
- [ ] Add a single select column, add options & few records; delete column; add single select column with same name - older record data persists. [column was never removed?]
- [ ] Switch from single select to multi select. Existing records remain in single-select mode until hard refresh / reload
- [ ] After clicking `add option`, move cursor inside text-box added automatically; currently, we need to click again in text box before starting to type select option data.
- [ ] Default option not working
- [ ] Text overflow's object if the option text is lengthier
- [ ] Warning not displayed for migration from multi-select to single-select
- [ ] Converting multi-select to single-select needs hard-refresh / reload

Verified
- [ ] Options order
- [ ] Options order after re-order
- [ ] Options colour edit- save
- [ ] Options text edit- save
- [ ] Use of options in filter
- [ ] Use of options in sort 
- [ ] Use of options in sort even after re-ordering select options
- [ ] Right click options

Issue #3031, opened 2022-08-06
vue3: API Snippet


o1lab:

dstala:
@mustafapc19 

- [ ] Right view pane IMO isn't required, as the snippet is small
- [ ] Tab headers, we can mark all caps as in v1 (OR) change following as `JavaScript`, `PHP`, `NocoDB-SDK`
  
  
![Screenshot 2022-08-11 at 12 05 41 PM](https://user-images.githubusercontent.com/86527202/184077194-4a308931-eeeb-494c-8169-ffc6d6621287.png)
  
- [ ] Sub-option-menu (next to clipboard copy) always show's AXIOS, even if AXIOS is not part of drop-down
- [ ] Sub-option-menu should not be displayed for `C`, `Java`, `Ruby`, `PHP`

PR #3035, opened 2022-08-08
vue3: API Snippet

mustafapc19:
Ref: #3031 

Know issues:
- Typing in Monaco editor will render this tooltip (in v1 the workaround is to have language config in Monaco editor config and not enable readonly mode of Monaco editor, but override editor value change callback and ignore editor value changes, but it causes console errors)
image
- Has minimap(this [PR](https://github.com/nocodb/nocodb/pull/2946) adds support for optional minimap support)
- `queryParams` will be properly typed by @pranavxc 

UI:
image

### DIFF
--- a/packages/nc-gui-v2/components/smartsheet/sidebar/MenuBottom.vue
+++ b/packages/nc-gui-v2/components/smartsheet/sidebar/MenuBottom.vue
@@ -91,7 +91,7 @@ function onOpenModal(type: ViewTypes, title = '') {
       </a-tooltip>
     </a-menu-item>
 
-    <SmartsheetSidebarMenuApiSnippet :model-value="showApiSnippet" @update:model-value="(val) => (showApiSnippet = val)" />
+    <SmartsheetSidebarMenuApiSnippet v-model="showApiSnippet" />
     <div class="flex-auto justify-end flex flex-col gap-4 mt-4">
       <button
         class="flex items-center gap-2 w-full mx-3 px-4 py-3 rounded !bg-primary text-white transform translate-x-4 hover:(translate-x-0 shadow-lg) transition duration-150 ease"
--- a/packages/nc-gui-v2/components/smartsheet/sidebar/MenuBottom.vue
+++ b/packages/nc-gui-v2/components/smartsheet/sidebar/MenuBottom.vue
@@ -12,9 +12,11 @@ const emits = defineEmits<Emits>()
 const { $e } = useNuxtApp()
 
 const isView = ref(false)
+let showApiSnippet = $ref(false)
 
 function onApiSnippet() {
   // get API snippet
+  showApiSnippet = true
   $e('a:view:api-snippet')
 }
 
@@ -89,6 +91,7 @@ function onOpenModal(type: ViewTypes, title = '') {
       </a-tooltip>
     </a-menu-item>
 
+    <SmartsheetSidebarMenuApiSnippet :model-value="showApiSnippet" @update:model-value="(val) => (showApiSnippet = val)" />
     <div class="flex-auto justify-end flex flex-col gap-4 mt-4">
       <button
         class="flex items-center gap-2 w-full mx-3 px-4 py-3 rounded !bg-primary text-white transform translate-x-4 hover:(translate-x-0 shadow-lg) transition duration-150 ease"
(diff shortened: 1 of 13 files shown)


packages/nc-gui-v2/components/smartsheet/sidebar/MenuBottom.vue
```diff
@@ -94,6 +96,7 @@ function onOpenModal(type: ViewTypes, title = '') {
       </a-tooltip>
     </a-menu-item>
 
+    <SmartsheetSidebarMenuApiSnippet :model-value="showApiSnippet" @update:model-value="(val) => (showApiSnippet = val)" />
```
bcakmakoglu, line 99: why not just use `v-model="showApiSnippet"`?
It does the exact same as `:model-value="value" @update:model-value="value = $event"`

packages/nc-gui-v2/components/smartsheet/sidebar/menu/ApiSnippet.vue
```diff
@@ -0,0 +1,183 @@
...
+import HTTPSnippet from 'httpsnippet'
+import { useClipboard } from '@vueuse/core'
+import { useToast } from 'vue-toastification'
+import { ActiveViewInj, MetaInj } from '~/context'
+
+const props = defineProps<Props>()
+
+const emits = defineEmits(['update:modelValue'])
+
+const toast = useToast()
```
bcakmakoglu, line 11: can we use ant-design-vue `notification` instead of `toast`?
We were considering moving to the notifications anyway 😄

packages/nc-gui-v2/components/smartsheet/sidebar/menu/ApiSnippet.vue
```diff
@@ -0,0 +1,183 @@
...
+}
+
+const { project } = $(useProject())
+const { appInfo } = $(useGlobal())
+const meta = $(inject(MetaInj))
+const view = $(inject(ActiveViewInj))
+const { xWhere } = useSmartsheetStoreOrThrow()
+const { queryParams } = $(useViewData(meta, view as any, xWhere))
+const { copy } = useClipboard()
+const { $state } = useNuxtApp()
```
bcakmakoglu, line 24: This is a double call technically.
`useGlobal` will return `$state` if it's already available, so you can just get `token` etc. from `useGlobal`

```ts
const { token, appInfo } = $(useGlobal())
```

packages/nc-gui-v2/components/smartsheet/sidebar/menu/ApiSnippet.vue
```diff
@@ -0,0 +1,183 @@
...
+  },
+  {
+    name: 'java',
+  },
+  {
+    name: 'c',
+  },
+]
+
+const selectedClient = $ref<string | undefined>(langs[0].clients && langs[0].clients[0])
```
bcakmakoglu, line 63: Seems a bit confusing, why confirm the existence of `langs[0].clients` if it is definetly defined anyway?
TypeScript should know that it has to be defined

packages/nc-gui-v2/composables/useProject.ts
```diff
@@ -11,6 +11,7 @@ export function useProject(projectId?: MaybeRef<string>) {
   const _projectId = $computed(() => unref(projectId))
 
   const project = useState<ProjectType>('project')
+  const appInfo = useState<any>('appInfo')
```
bcakmakoglu, line 14: Is this still necessary or can this be removed?
`appInfo` seems to be stored in the global state now so this seems like two states.

packages/nc-gui-v2/plugins/state.ts
```diff
@@ -25,6 +26,12 @@ export default defineNuxtPlugin((nuxtApp) => {
   /** set i18n locale to stored language */
   nuxtApp.vueApp.i18n.locale.value = state.lang.value
 
+  try {
+    state.appInfo = await $api.utils.appInfo()
```
bcakmakoglu, line 30: should probably be `state.appInfo.value` 
Yo could also just destructure `state`

```ts
const { appInfo } = useGlobal()

...

appInfo.value = ...
```

bcakmakoglu (changes requested):

packages/nc-gui-v2/components/smartsheet/sidebar/menu/ApiSnippet.vue
mustafapc19: @bcakmakoglu Because clients can be undefined as well. It is optional. We hide the dropdown for client in such situations.

bcakmakoglu (approved):

PR #3057, opened 2022-08-09
refactor(gui-v2): ui updates

pranavxc:
## Change Summary

Changes suggested by @o1lab 

## Change type

## Test/ Verification

Provide summary of changes.


bcakmakoglu:
@o1lab @dstala @pranavxc 

##

image

1. Duplicated Icons (sidebar collapsed + topbar on the grid has the same icons)
2. Spacing is not adequate, why was the spacing reduced in the first place? We have way too much space that's unfilled anyway, we need to reduce the whitespace not increase it.
3. The icons on the topbar make no sense to me. They're not fitting the the topbar - as I mentioned once before, everything else in that bar is made with an icon and text, suddenly there's some icons and they're spaced with dots. Why? That is horrible design.
4. Tooltip alignments are broken in topbar icons because of the faulty spacing.

##

image

1. Why was the input made to be "invisible"? Looks like it's disabled to me. Can we not keep the regular input style or at least an input style similar to the signin form?  or is there any particular benefit we hope to have by making our inputs as hard to discern as possible?

##

image

1. Sidebar left active element has another color than the right sidebar active element (icon on the left is blue, icon on the right is pink) - should it not be uniform? It's a minor thing and doesn't matter too much but blue on blue seems off.

##

image

1. Why the decoloring? Makes the elements a lot less prominent (is that the goal?) and robs the sidebar of some visual effect imo. Maybe instead of white on white we can at least give it a slight gray shade or a very light blue.

## 

Honestly, it's your choice but I would revert almost all the changes here except for the active tab color.

bcakmakoglu:
image

1. Broken menu - hovering on the menu will trigger all elements to apply hover styles
2. Menu trigger is hover instead of click, which seems sort of confusing. I'd rather use a click on the dots instead of hover.

bcakmakoglu:
image

1. Maybe we can make the `X` more prominent on a selected tab as well? It looks a bit off like that.

o1lab:
Thanks for the patience and review :D 

As discussed, in attempt to swiftly improve things related to layout - we did this.

> 1. Duplicated Icons (sidebar collapsed + topbar on the grid has the same icons)

We can remove the icons from right nav drawer

> 2. Faulty Spacing in the topbar and sidebar - I think the spacing from before was good (the right sidebar).

We can remove the icons from right nav drawer

> 3. The icons on the topbar make no sense to me. They're not fitting the the topbar - as I mentioned once before, everything else in that bar is made with an icon and text, suddenly there's some icons and they're spaced with dots.

@bcakmakoglu : see how best to go about this. I need them to be part of topbar. I guess reload icon is a bit universal and it has a tooltip. For the other two - see how best we can do this. On smaller screens if we can make the additional options go inside three dots - it will be the right behaviour.

> Menu trigger is hover instead of click, which seems sort of confusing. I'd rather use a click on the dots instead of hover.
Agreed


> Why the decoloring? Makes the elements a lot less prominent (is that the goal?) and robs the sidebar of some visual effect imo. Maybe instead of white on white we can at least give it a slight gray shade or a very light blue.

'We're hiring' was outlined in v1. And I think API snippet was too much in the eye as a CTA. Hence took out the colors.


> Why was the input made to be "invisible"? Looks like it's disabled to me. Can we not keep the regular input style or at least an input style similar to the signin form? The inputs are, in my opinion, really hard to discern this way. Especially if the screen is a decent size.

Yes, agreed - something like signin will help.

> Sidebar left active element has another color than the right sidebar active element (icon on the left is blue, icon on the right is pink) - should it not be uniform? It's a minor thing and doesn't matter too much but blue on blue seems off.

Let the active icon be blue.

dstala:
- [ ] `Search text box` & attached `search icon box` above Project-tree border is slightly mis-aligned
- [ ] Close all tabs; size of `Add / Import` shrinks. 

Screenshot 2022-08-11 at 2 52 11 PM

PR #3082, opened 2022-08-11
fix(gui-v2): select improvements

mertmit:
## Change Summary

- Fix empty option handling for multiselect
- Use column_name instead of title for queries
- Hide arrow from single select if a value already selected
- Multiselect cell overflow

## Change type


dstala (approved):
Consider following in the next PR
- Error for empty select option: Currently we display as `Form validation failed`. We can be more specific and convey `Empty option not allowed`
- Checkmark added to mulit-select selections is good. We can add it for single select as well - to bring in UI uniformity
- Convert from multi-select to single-select, warning is pending; and we need hard refresh for icon change to appear.

PR #3134, opened 2022-08-13
fix(gui-v2): column edit/add handling

mertmit:
## Change Summary

- Use local state for column options
- editOrAddProvider -> this way no unnecessary injections
- useColumn -> use computedRefs instead to make changes on column live

## Change type


dstala (approved):
Verified

o1lab:
To avoid conflicts, will merge it. Any review comments from @bcakmakoglu can be seen from Mert. 😄

Issue #3154, opened 2022-08-15
vue3 : Keep the right nav drawer closed by default


o1lab:
- [x] Let's keep the right nav drawer closed by default - so there is more space for spreadsheet
- [x] And once toggled, the toggled state has to be saved in browser. Currently, if you open a new table - the toggled state goes back to default.

bcakmakoglu:
Why would the toggled have to be saved?
Never seen any other page save my sidebar of collapsed/open, which imo is strange UX.

PR #3167, opened 2022-08-16
fix(gui-v2): right sidebar state

wingkwong:
## Change Summary

- remove RightSidebarInj and use state value instead. ref: #3154 

## Change type


bcakmakoglu:
Left sidebar state isn't saved, why should the right one be saved in storage now?

This PR is overkill for just having the sidebar collapsed by default. 

```ts
provide(RightSidebarInj, ref(false))
```
Is enough to keep it collapsed on page-load.

Also there is discussion to move the right sidebar completely to the left, so this would all be removed anyway. 

I would decline this PR.

bcakmakoglu:
If this is suppposed to go through and the sidebar state is supposed to be saved, please use the composable for Sidebars (`useSidebar`) and extend on it to make it peristable. Keeps our composables separated by their logical concern instead of just dumping everything into global state 😄 

You can just add a key/id to the props
```ts
interface UseSidebarProps {
  hasSidebar?: boolean
  isOpen?: boolean
  storageKey?: string // if a storageKey is passed, use that key for localStorage
}
```

And update the composable to this:
```ts
const [setup, use] = useInjectionState((props: UseSidebarProps = {}) => {
  let isOpen = ref(props.isOpen ?? false)
  let hasSidebar = ref(props.hasSidebar ?? true)

  function toggle(state?: boolean) {
    isOpen.value = state ?? !isOpen.value
  }

  function toggleHasSidebar(state?: boolean) {
    hasSidebar.value = state ?? !hasSidebar.value
  }

  if (props.storageKey) {
    const storage = toRefs(useStorage(props.storageKey, { isOpen, hasSidebar }, localStorage).value)
    isOpen = storage.isOpen
    hasSidebar = storage.hasSidebar
  }

  watch(
    hasSidebar,
    (nextHasSidebar) => {
      if (!nextHasSidebar) toggle(false)
    },
    { immediate: true },
  )

  watch(
    isOpen,
    (nextIsOpen) => {
      if (nextIsOpen && !hasSidebar.value) toggleHasSidebar(true)
    },
    { immediate: true },
  )

  return {
    isOpen,
    toggle,
    hasSidebar,
    toggleHasSidebar,
  }
}, 'useSidebar')
```

bcakmakoglu:
Another super minor note:
This shouldn't be a `fix` as it's not actually fixing any bug.
It should be, in my opinion, a `feat` as it's introducing a new feature here.

wingkwong:
Closing. Will be handled in a new PR.

Issue #3156, opened 2022-08-15
vue3 : Colors for primary-value, has-many, belongs-to, many-to-many


o1lab:
- [x] Text
  - [x] primary-value can have purple
  - [x] others : can have bluish
- [x] Icons in column header need color like we'd earlier.
- [x] Fixed column height
- [ ] Adjust column height
- [x] Column header - make it as pointer

PR #3172, opened 2022-08-16
fix(gui-v2): minor improvements

mertmit:
## Change Summary

Re #3156
- add active border to treeview
- merged Team & Settings into single menu item
- column edit menu fixed placement, trigger changed to click instead of hover
- colored primary value
- relational type colored icons
- project dropdown level with header
- relational column item chip color revised

## Change type


wingkwong (commented):
Cell heights vary based on the number of records.

Table without records

image

Table with 1 record

image

Table with lots of records 

image

pranavxc (approved):

wingkwong (approved):

Issue #2952, opened 2022-08-02
vue3: Percent Column


wingkwong:
Percent doesn't exist in v1 but in this [PR](https://github.com/nocodb/nocodb/pull/2581)

dstala:
- percent column- modal is different from other as we bring in default input outside (I kind of like it though)
- the configurations on the cell didn't work (precision, default). that will be under pranav scope? 
- default provisioning currently repeats as default number in main menu and default value in expanded menu

mustafapc19:
@dstala There are some complications integrating this [PR](https://github.com/nocodb/nocodb/pull/2581) since backend part(`conditionV2`) has changed a bit. On discussion with @pranavxc have decided to skip the options for now till gui-v2 is completed. Should we hide new options for percent for now or keep things as it is?

dstala:
@mustafapc19 
let us make all features equivalent to v1 (current master, as is). we can consider rest as enhancements once we have vue2===vue3

mustafapc19:
@dstala In vue2(v1) percentage column has no options and cell is just a `FloatCell`. Taking that path for now 👍 

cc: @wingkwong

PR #3180, opened 2022-08-16
Reverted percent to how percent is handled in v1

mustafapc19:
Ref: [Comment](https://github.com/nocodb/nocodb/issues/2952#issuecomment-1209012588)

UI:
image
image


wingkwong (commented):
I'd suggest comment out those code with some comments instead of deleting so that it's more easy to enable it back later on? What do you think?

mustafapc19:
@wingkwong Makes sense

mustafapc19:
@wingkwong Have included the old code 👍

wingkwong (approved):

Issue #2980, opened 2022-08-03
vue3: Email Json, Phonenumber, url, time column


mustafapc19:

dstala:
@mustafapc19 

Combined observations 

Email / Phone number
- [ ] Too many toast alerts as we type email address. Trigger validation check only after `enter` key is pressed or after `change in cell focus`

URL
- [ ] v1 did allow storing of `www.google.com`, `google.com` with validation enabled. v2 mandates it to be of format `https://google.com`
- [ ] no error message displayed on validation failure. Just record isn't stored

Time
- [ ] Records for this cell are not persistent; disappear on reload
- [ ] Since we are displaying time in 24H format within cell, would be good if we remove AM/PM and just display 0~23H instead.
  
![Screenshot 2022-08-08 at 10 00 22 PM](https://user-images.githubusercontent.com/86527202/183467457-1d5ce265-8a23-4691-945e-0e53349b0dbc.png)

PR #3185, opened 2022-08-16
Added showing error message on invalid url entry

mustafapc19:
Ref: https://github.com/nocodb/nocodb/issues/2980#issuecomment-1208348855

Screenshot:
image
image


wingkwong (changes requested):
1. Validation should be only performed when users have selected `Accept only valid URL`. Now validation is triggered anyway.

image
 
2. The value shouldn't be saved if the URL is invalid. Now error message is shown but the value is stored. (you can refresh / reload to see it)

mustafapc19:
@wingkwong Regarding 2nd point can you tell me the steps to reproduce? In my case when I am reloading/refreshing its reverting to the last saved/validated value.

wingkwong:
Now it seems fine for the second point. However, with `Accept only valid URL` enabled, I couldn't set it empty.

wingkwong:
Another thought 

- Create a URL column with `Accept only valid URL` disabled
- Input data with some invalid URL
- Edit the column. Enabled `Accept only valid URL`

In this case, I think it'd be great if those cells have some indicators to show that they are invalid so that users can easily spot them to change the values.

mustafapc19:
@wingkwong Have added ui indication for URL being invalid. Attached the screenshot in the PR description.

packages/nc-gui-v2/components/cell/Url.vue
```diff
@@ -35,14 +41,37 @@ const url = computed(() => {
...
 
-  <nuxt-link v-else-if="isValid" class="text-sm underline hover:opacity-75" :to="url" target="_blank">{{ value }} </nuxt-link>
+    <span v-else class="w-9/10 overflow-ellipsis overflow-hidden">{{ value }}</span>
 
-  <span v-else>{{ value }}</span>
+    <div v-if="!isValid && value?.length && !editEnabled" class="mr-1 w-1/10">
+      <a-tooltip placement="top">
+        <template #title> Invalid URL </template>
+        <div class="flex flex-row items-center">
+          <MaterialSymbolsWarning />
```
wingkwong, line 70: Suggest to color it.

image

wingkwong (commented):
`qweq` shouldn't be a hyperlink. (Accept only valid URL: false)

v2:

image

v1: 
image

wingkwong (approved):

Issue #3200, opened 2022-08-17
[vue3] Lookup shows empty cell


dstala:
<img width="977" alt="Screenshot 2022-08-17 at 12 52 33 PM" src="https://user-images.githubusercontent.com/86527202/185059082-2ff0d74e-757e-4628-a67c-7b6fd0697e02.png">

dstala:
@mertmit observation is same for rollup as well
  
![Screenshot 2022-08-17 at 3 54 00 PM](https://user-images.githubusercontent.com/86527202/185096346-45a77a44-4005-4c70-93c7-a438fc25ad1d.png)

mertmit:
imports are mismatching on VirtualCell.vue but reverting them also cause an issue I will be working on that issue (circular reference)

PR #3204, opened 2022-08-17
fix: Lookup, Rollup, Count

mertmit:
## Change Summary

Re #3200

## Change type


packages/nc-gui-v2/components/smartsheet/VirtualCell.vue
```diff
@@ -14,13 +14,13 @@ const ManyToMany = defineAsyncComponent(() => import('../virtual-cell/ManyToMany
 
 const BelongsTo = defineAsyncComponent(() => import('../virtual-cell/BelongsTo.vue'))
 
-const Rollup = defineAsyncComponent(() => import('../virtual-cell/HasMany.vue'))
+const Rollup = defineAsyncComponent(() => import('../virtual-cell/Rollup.vue') as any)
```
bcakmakoglu, line 17: We really gotta find a way to fix these circular component dependencies 😢

bcakmakoglu (approved):

Issue #3220, opened 2022-08-18
vue3: password hints were more reactive in v1


dstala:
v2 indicates about only 8 character limit, doesn't convey about number, a capital & special character requirement
  

![Screenshot 2022-08-18 at 10 19 48 AM](https://user-images.githubusercontent.com/86527202/185296432-c651e434-1829-4ae4-9bda-205e831b4e2d.png)
![Screenshot 2022-08-18 at 10 20 01 AM](https://user-images.githubusercontent.com/86527202/185296438-551663fc-5f37-49fc-a45b-5e2aae9d99e4.png)
![Screenshot 2022-08-18 at 10 20 09 AM](https://user-images.githubusercontent.com/86527202/185296443-c09f4cb4-1543-4413-9059-a8584a433190.png)
![Screenshot 2022-08-18 at 10 20 17 AM](https://user-images.githubusercontent.com/86527202/185296446-9f344c59-f0d3-47e2-979c-75524c7dae80.png)

PR #3223, opened 2022-08-18
Improved email and password validation of signup page

mustafapc19:
Ref: #3220

- Added more reactive password hint.
- Email validation doesn't show email is required and invalid in the case of empty email field.

UI:
image
image

### DIFF
--- a/packages/nc-gui-v2/pages/signup/[[token]].vue
+++ b/packages/nc-gui-v2/pages/signup/[[token]].vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { validatePassword } from 'nocodb-sdk'
 import {
   definePageMeta,
   extractSdkResponseErrorMsg,
@@ -43,7 +44,7 @@ const formRules = {
     {
       validator: (_: unknown, v: string) => {
         return new Promise((resolve, reject) => {
-          if (isEmail(v)) return resolve(true)
+          if (!v?.length || isEmail(v)) return resolve(true)
           reject(new Error(t('msg.error.signUpRules.emailInvalid')))
         })
       },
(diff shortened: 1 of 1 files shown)


packages/nc-gui-v2/pages/signup/[[token]].vue
```diff
@@ -43,17 +44,23 @@ const formRules = {
     {
       validator: (_: unknown, v: string) => {
         return new Promise((resolve, reject) => {
-          if (isEmail(v)) return resolve(true)
+          if (!v?.length || isEmail(v)) return resolve(true)
```
Comment (bcakmakoglu) [inline packages/nc-gui-v2/pages/signup/[[token]].vue:47]: Wouldn't this accept *any* type of string now, regardless of being a valid email or not?
The conidition is now `if (string has length OR string is email-like) then { ... }`

packages/nc-gui-v2/pages/signup/[[token]].vue
```diff
@@ -43,17 +44,23 @@ const formRules = {
...
           reject(new Error(t('msg.error.signUpRules.emailInvalid')))
         })
       },
       message: t('msg.error.signUpRules.emailInvalid'),
     },
   ],
   password: [
-    // Password is required
-    { required: true, message: t('msg.error.signUpRules.passwdRequired') },
```
Comment (bcakmakoglu) [inline packages/nc-gui-v2/pages/signup/[[token]].vue:55]: Why was the `required` rule removed?

bcakmakoglu (changes requested):

packages/nc-gui-v2/pages/signup/[[token]].vue
Comment (mustafapc19) [reply packages/nc-gui-v2/pages/signup/[[token]].vue:55]: I have added a custom validator which checks for this as well.

Comment (mustafapc19) [reply packages/nc-gui-v2/pages/signup/[[token]].vue:55]: @bcakmakoglu

packages/nc-gui-v2/pages/signup/[[token]].vue
Comment (mustafapc19) [reply packages/nc-gui-v2/pages/signup/[[token]].vue:47]: @bcakmakoglu when string has no length this validator will be disabled effectively which is required since there is a validator above this which checks if a string is empty. This is a fix for the case of empty string where both of the validation error message comes up.
image

packages/nc-gui-v2/pages/signup/[[token]].vue
Comment (bcakmakoglu) [reply packages/nc-gui-v2/pages/signup/[[token]].vue:55]: Ok 😄

dstala (approved):

Issue #3119, opened 2022-08-12
vue3 : Grid - add new row is not opening the form


o1lab:
- Click add new row -> new row form is not opening.

PR #3225, opened 2022-08-18
Made `Add new row` button functional

mustafapc19:
ref: #3119

- Clicking on `Add new row` button creates an empty row and opens the modal.
- If `cancel` button is clicked in `expanded form` then the created row will be deleted (this is not in v1).


bcakmakoglu (changes requested):

wingkwong (commented):
Grid.vue

`readonly` -> `readOnly`

wingkwong (approved):

bcakmakoglu (approved):

Issue #3210, opened 2022-08-17
Handle additional params in database connection


o1lab:
This is related to the issue we discussed it yesterday.

- When `options=project=project.name` & `ssl=true` is present we need to handle it in
  - [x] NC_DB : turning these additional params as part of database config
  - [x] DATABASE_URL : turning these additional params into NC_DB params
  - [x] Project create (GUI)

We should also try to make this as a generic solution for other params

PR #3306, opened 2022-08-23
feat: additional parameters for external database connections

mertmit:
## Change Summary

Re #3210
- Allow any customized parameter for external db configurations
- Added new SSL mode: Allowed
- Fixed edit json modal
  - Use ant-design modal
  - Only save if user accept the changes

## Change type

## Test/ Verification

image
image

### DIFF
--- a/packages/nc-gui-v2/pages/index/index/create-external.vue
+++ b/packages/nc-gui-v2/pages/index/index/create-external.vue
@@ -1,5 +1,6 @@
 <script lang="ts" setup>
 import { Form, Modal, message } from 'ant-design-vue'
+import type { RawValueType } from 'ant-design-vue'
 import {
   clientTypes,
   computed,
@@ -96,6 +97,24 @@ const onClientChange = () => {
   populateName(formState.title)
 }
 
+const onSSLModeChange = (v: RawValueType) => {
+  switch (v) {
+    case 'No':
+      delete formState.dataSource.connection.ssl
+      break
+    case 'Allowed':
+      formState.dataSource.connection.ssl = true
+      break
+    default:
+      formState.dataSource.connection.ssl = {
+        ca: '',
+        cert: '',
+        key: '',
+      }
+      break
+  }
+}
+
 const inflectionTypes = ['camelize', 'none']
 const configEditDlg = ref(false)
 
@@ -120,7 +139,7 @@ const onFileSelect = (key: 'ca' | 'cert' | 'key', el: HTMLInputElement) => {
 }
 
 const sslFilesRequired = computed<boolean>(() => {
-  return formState?.sslUse && formState.sslUse !== 'No'
+  return formState?.sslUse && formState.sslUse !== 'No' && formState.sslUse !== 'Allowed'
 })
 
 function getConnectionConfig() {
@@ -335,7 +354,7 @@ onMounted(() => {
           <a-collapse-panel key="1" :header="$t('title.advancedParameters')">
             <!--            todo:  add in i18n -->
             <a-form-item label="SSL mode">
-              <a-select v-model:value="formState.sslUse" @change="onClientChange">
+              <a-select v-model:value="formState.sslUse" @select="onSSLModeChange">
                 <a-select-option v-for="opt in sslUsage" :key="opt" :value="opt">{{ opt }}</a-select-option>
               </a-select>
             </a-form-item>
@@ -385,13 +404,13 @@ onMounted(() => {
             <a-divider />
 
             <a-form-item :label="$t('labels.inflection.tableName')">
-              <a-select v-model:value="formState.inflection.inflectionTable" @change="onClientChange">
+              <a-select v-model:value="formState.inflection.inflectionTable">
                 <a-select-option v-for="type in inflectionTypes" :key="type" :value="type">{{ type }}</a-select-option>
               </a-select>
             </a-form-item>
 
             <a-form-item :label="$t('labels.inflection.columnName')">
-              <a-select v-model:value="formState.inflection.inflectionColumn" @change="onClientChange">
+              <a-select v-model:value="formState.inflection.inflectionColumn">
                 <a-select-option v-for="type in inflectionTypes" :key="type" :value="type">{{ type }}</a-select-option>
               </a-select>
             </a-form-item>
--- a/packages/nc-gui-v2/pages/index/index/create-external.vue
+++ b/packages/nc-gui-v2/pages/index/index/create-external.vue
@@ -2,7 +2,8 @@
 import { Form, Modal, message } from 'ant-design-vue'
 import type { SelectHandler } from 'ant-design-vue/es/vc-select/Select'
 import {
-  certTypes,
+  CertTypes,
+  SSLUsage,
   clientTypes,
   computed,
   extractSdkResponseErrorMsg,
@@ -16,7 +17,6 @@ import {
   projectTitleValidator,
   readFile,
   ref,
-  sslUsage,
   useApi,
   useI18n,
   useNuxtApp,
@@ -24,6 +24,7 @@ import {
   watch,
 } from '#imports'
 import { ClientType } from '~/lib'
+import { DefaultConnection, SQLiteConnection } from '~/utils'
 import type { ProjectCreateForm } from '~/utils'
 
 const useForm = Form.useForm
@@ -47,7 +48,7 @@ let formState = $ref<ProjectCreateForm>({
     inflectionColumn: 'camelize',
     inflectionTable: 'camelize',
   },
-  sslUse: sslUsage.No,
+  sslUse: SSLUsage.No,
   extraParameters: [],
 })
 
@@ -58,7 +59,7 @@ const customFormState = ref<ProjectCreateForm>({
     inflectionColumn: 'camelize',
     inflectionTable: 'camelize',
   },
-  sslUse: sslUsage.No,
+  sslUse: SSLUsage.No,
   extraParameters: [],
 })
 
@@ -103,33 +104,39 @@ const onClientChange = () => {
   populateName(formState.title)
 }
 
-const onSSLModeChange = ((mode: sslUsage) => {
-  switch (mode) {
-    case sslUsage.No:
-      delete formState.dataSource.connection.ssl
-      break
-    case sslUsage.Allowed:
-      formState.dataSource.connection.ssl = 'true'
-      break
-    default:
-      formState.dataSource.connection.ssl = {
-        ca: '',
-        cert: '',
-        key: '',
-      }
-      break
+const onSSLModeChange = ((mode: SSLUsage) => {
+  if (formState.dataSource.client !== ClientType.SQLITE) {
+    const connection = formState.dataSource.connection as DefaultConnection
+    switch (mode) {
+      case SSLUsage.No:
+        delete connection.ssl
+        break
+      case SSLUsage.Allowed:
+        connection.ssl = 'true'
+        break
+      default:
+        connection.ssl = {
+          ca: '',
+          cert: '',
+          key: '',
+        }
+        break
+    }
   }
 }) as SelectHandler
 
 const updateSSLUse = () => {
-  if (formState.dataSource.connection.ssl) {
-    if (typeof formState.dataSource.connection.ssl === 'string') {
-      formState.sslUse = sslUsage.Allowed
+  if (formState.dataSource.client !== ClientType.SQLITE) {
+    const connection = formState.dataSource.connection as DefaultConnection
+    if (connection.ssl) {
+      if (typeof connection.ssl === 'string') {
+        formState.sslUse = SSLUsage.Allowed
+      } else {
+        formState.sslUse = SSLUsage.Preferred
+      }
     } else {
-      formState.sslUse = sslUsage.Preferred
+      formState.sslUse = SSLUsage.No
     }
-  } else {
-    formState.sslUse = sslUsage.No
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -235,14 +298,7 @@ onMounted(() => {
...
     <general-noco-icon class="color-transition hover:(ring ring-accent)" :class="[isLoading ? 'animated-bg-gradient' : '']" />
 
-    <div
-      class="color-transition transform group absolute top-5 left-5 text-4xl rounded-full bg-white cursor-pointer"
-      @click="navigateTo('/')"
-    >
-      <MdiChevronLeft class="text-black group-hover:(text-accent scale-110)" />
-    </div>
-
-    <h1 class="prose-2xl font-bold self-center my-4">{{ $t('activity.createProject') }}</h1>
```
bcakmakoglu, line 245: Why was this removed and replaced by a `h3`?
Now there's no navigation option in the form. 😢

packages/nc-gui-v2/utils/validation.ts
```diff
@@ -127,3 +127,17 @@ export const importExcelUrlValidator = {
     })
   },
 }
+
+export const extraParameterValidator = {
+  validator: (rule: any, value: any, callback: (errMsg?: string) => void) => {
+    for (const param of value) {
+      if (param.key === '') {
+        callback('Parameter key cannot be empty')
```
bcakmakoglu, line 135: Use of callback is deprecated, please use Promise.reject instead (otherwise it throws warnings in the console)

packages/nocodb/src/lib/utils/NcConfigFactory.ts
```diff
@@ -155,17 +195,21 @@ export default class NcConfigFactory implements NcConfig {
         },
       } as any;
     } else {
+      const parsedQuery = {}
+      for (const [key, value] of url.searchParams.entries()) {
+        const fnd = knownQueryParams.find((param) => param.parameter === key || param.aliases.includes(key))
```
bcakmakoglu, line 200: sort of obscure variable name 😅

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -77,21 +90,43 @@ const validators = computed(() => {
...
+}
+
 const onClientChange = () => {
   formState.dataSource = { ...getDefaultConnectionConfig(formState.dataSource.client) }
+  populateName(formState.title)
 }
 
-const inflectionTypes = ['camelize', 'none']
-const configEditDlg = ref(false)
+const onSSLModeChange = (v: any) => {
```
bcakmakoglu, line 102: Should not be `any`. 
Either use an enum or just use `v: 'No' | 'Allowed' | string`
If possible don't use capitalized values either and change it to `no` and `allowed` or (just use a boolean value instead?)

bcakmakoglu (changes requested):

packages/nc-gui-v2/pages/index/index/create-external.vue
mertmit: oh this should be by mistake, probably while rebasing. thank you for pointing out!

packages/nc-gui-v2/pages/index/index/create-external.vue
mertmit: This causes @select event to fire a type error, what should I use to avoid this
image

bcakmakoglu: I opened a separate PR with proposed type fixes #3307 
If it's okay with you, you can merge that into this PR 😄

o1lab:
Just remembered - in create project gui, we can have import from URL? You could use same function in backend by putting it in SDK.

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -452,6 +452,7 @@ onMounted(() => {
 
             <a-divider />
 
+
```
bcakmakoglu: Unnecessary whitespace 😄

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -77,51 +93,91 @@ const validators = computed(() => {
...
+  }
+}) as unknown as SelectHandler
 
-// populate database name based on title
-watch(
-  () => formState.title,
-  (v) => (formState.dataSource.connection.database = `${v?.trim()}_noco`),
-)
+const updateSSLUse = () => {
+  if (formState?.dataSource?.connection?.ssl) {
```
bcakmakoglu, line 124: Unnecessary `?`
`formState` should not be undefined
`dataSource` can be undefined? Should probably not be undefined either so the check might be unnecessary as well.
`connection` should probably be defined as well but that one I didn't check.

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -77,51 +93,91 @@ const validators = computed(() => {
...
+    if (typeof formState.dataSource.connection.ssl === 'string') {
+      formState.sslUse = 'Allowed'
```
bcakmakoglu, line 126: Are these required to be Capitalized words? If not, I would be glad if we change it to lowercase `no` `allowed` etc. because it's easier to avoid mistakes when everything is lowercase.

Even better yet, if we can instead use an Enum for this that'd make it optimal I'd say (then you can keep the Capitalized words, it wouldn't matter)

```ts
enum SSLUse {
	No,
	Allowed,
	Preferred,
}
```

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -77,51 +93,91 @@ const validators = computed(() => {
...
+const importURL = ref('')
+const configEditDlg = ref(false)
+const importURLDlg = ref(false)
 
 const caFileInput = ref<HTMLInputElement>()
 const keyFileInput = ref<HTMLInputElement>()
 const certFileInput = ref<HTMLInputElement>()
 
-const onFileSelect = (key: 'ca' | 'cert' | 'key', el: HTMLInputElement) => {
+const onFileSelect = (key: 'ca' | 'cert' | 'key', el?: HTMLInputElement) => {
```
bcakmakoglu, line 152: Maybe we can use an `enum` for the ssl cert types as well? 😄

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -209,15 +265,48 @@ const testConnection = async () => {
...
+const handleOk = () => {
+  formState = { ...customFormState.value }
+  configEditDlg.value = false
+  updateSSLUse()
+}
+
 // reset test status on config change
 watch(
-  () => formState.dataSource,
+  () => formState?.dataSource,
```
bcakmakoglu, line 296: Unnecessary `?`

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -209,15 +265,48 @@ const testConnection = async () => {
...
   () => (testSuccess.value = false),
   { deep: true },
 )
 
+// populate database name based on title
+watch(
+  () => formState?.title,
```
bcakmakoglu, line 303: unnecessary `?`

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -77,51 +93,91 @@ const validators = computed(() => {
...
+
 const onClientChange = () => {
   formState.dataSource = { ...getDefaultConnectionConfig(formState.dataSource.client) }
+  populateName(formState.title)
 }
 
-const inflectionTypes = ['camelize', 'none']
-const configEditDlg = ref(false)
+const onSSLModeChange = ((mode: 'No' | 'Allow' | string) => {
+  switch (mode) {
```
bcakmakoglu, line 106: `if ('ssl' in formState.dataSource.connection)`
otherwise TS complains that `.ssl` does not exist on `formState.dataSource.connection`

packages/nc-gui-v2/pages/index/index/create-external.vue
```diff
@@ -209,15 +265,48 @@ const testConnection = async () => {
   }
 }
 
+const handleImportURL = async () => {
+  if (!importURL.value || importURL.value === '') return
+
+  const connectionConfig = await api.utils.urlToConfig({ url: importURL.value })
```
bcakmakoglu, line 271: Could you add a `todo` that `urlToConfig` doesn't exist on `api.utils` type?

bcakmakoglu (changes requested):

packages/nc-gui-v2/pages/index/index/create-external.vue
mertmit: Yes but this actually breaks onSSLModeChange, as it manipulates ssl property by creating or removing, so it had to run in both cases

bcakmakoglu: Ah then our structure is somewhat bad / type unsafe 
For now you can just type cast it to `Record<string, any>` I guess 😅

packages/nc-gui-v2/pages/index/index/create-external.vue
mertmit: This should be fixed by rebuilding nocodb-sdk

bcakmakoglu: Ah ok, my bad 😄

packages/nc-gui-v2/utils/projectCreateUtils.ts
```diff
@@ -209,4 +144,19 @@ export const getDefaultConnectionConfig = (client: ClientType): ProjectCreateFor
   }
 }
 
-export const sslUsage = ['No', 'Preferred', 'Required', 'Required-CA', 'Required-IDENTITY']
+enum sslUsage {
```
bcakmakoglu, line 147: Enums should be Capitalized 😅 -> `SSLUsage` or `SslUsage`
So it's obvious they're not regular variables.

packages/nc-gui-v2/utils/projectCreateUtils.ts
```diff
@@ -209,4 +144,19 @@ export const getDefaultConnectionConfig = (client: ClientType): ProjectCreateFor
...
+  No = 'No',
+  Allowed = 'Allowed',
+  Preferred = 'Preferred',
+  Required = 'Required',
+  RequiredWithCa = 'Required-CA',
+  RequiredWithIdentity = 'Required-Identity',
+}
+
+enum certTypes {
```
bcakmakoglu, line 156: Same as above - Capitalize enum

packages/nc-gui-v2/utils/validation.ts
```diff
@@ -127,3 +127,19 @@ export const importExcelUrlValidator = {
     })
   },
 }
+
+export const extraParameterValidator = {
+  validator: (rule: any, value: any) => {
```
bcakmakoglu, line 132: Can be updated to `validator: (_: unknown, value: Record<string, string>[])`

packages/nc-gui-v2/utils/projectCreateUtils.ts
```diff
@@ -4,30 +4,15 @@ export interface ProjectCreateForm {
...
-        }
-      | {
-          client?: ClientType.SQLITE
-          database: string
-          connection?: {
-            filename?: string
-          }
-          useNullAsDefault?: boolean
-        }
+    connection: Record<string, any>
```
bcakmakoglu, line 7: Now all these properties are basically "unknown" to us and I'd have to guess which properties belong to one another.
Instead of removing all this - which we could still use to improve the actual typing that can differentiate better between each one

For example
```ts
enum ClientType {
  NONE = 'none',
  MYSQL = 'mysql2',
  MSSQL = 'mssql',
  PG = 'pg',
  SQLITE = 'sqlite3',
  VITESS = 'vitess',
}

interface DefaultConnection {
	client?: ClientType.NONE,
	host: string
        database: string
        user: string
        password: string
        port: number | string
        ssl?: Record<certTypes, string> | 'true'
}

interface SQLiteConnection {
	client?: ClientType.SQLITE
        database: string
        connection?: {
           filename?: string
        }
        useNullAsDefault?: boolean
}

type Connections = DefaultConnection | SQLiteConnection 


...

	dataSource: {
		connection: Connections
		...
	}
```

Now if each Connection shares the fact that they have the `client` property, we could always differentiate between them by asking what client type is currently assigned.

```ts
if (formState.dataSource.connection.client === ClientType.None) { ... }
else if (formState.dataSource.connection.client === ClientType.SQLITE) { ... }
```

If it's about the `onSSLModeChange` issue - if `.ssl` is supposed to be writable for *any* type of Connection, it should also be included in each type that is possible - otherwise it's incorrect and TypeScript rightfully complains about it 😄 

Otherwise with the suggested changes this would work
```ts
const onSSLModeChange: SelectHandler = (mode: sslUsage) => {
  if (formState.dataSource.connection.client === ClientType.NONE) {
    switch (mode) {
      case sslUsage.No:
        delete formState.dataSource.connection.ssl
        break
      case sslUsage.Allowed:
        formState.dataSource.connection.ssl = 'true'
        break
      default:
        formState.dataSource.connection.ssl = {
          ca: '',
          cert: '',
          key: '',
        }
        break
    }
  }
}
```

bcakmakoglu (changes requested):

packages/nc-gui-v2/utils/projectCreateUtils.ts
bcakmakoglu: You could also, if necessary, type cast `formState.dataSource.connection` to `DefaultConnection` - at least we have some typesafety then 😅

o1lab:
@mertmit : if you can help resolve above things - lets get this in soon :D

packages/nc-gui-v2/utils/validation.ts
```diff
@@ -127,3 +127,19 @@ export const importExcelUrlValidator = {
     })
   },
 }
+
+export const extraParameterValidator = {
+  validator: (_: unknown, value: any) => {
```
bcakmakoglu, line 132: should at least be `any[]` but doesn't `{ key: string }[]` work here as well?

bcakmakoglu (approved):

Issue #3340, opened 2022-08-25
vue3 : While loading tables a spinner at the center is required.


o1lab:
like in v1

Issue #3089, opened 2022-08-11
vue3 : When tab is clicked - needs a reload.


o1lab:

dstala:
@o1lab check if you can still see the issue in `develop`? I couldn't reproduce any issue with tab usage.

o1lab:
This is an optimisation issue that @pranavxc knows.

PR #3342, opened 2022-08-25
vue3: Forced meta loading when tab changes

mustafapc19:
Ref: https://github.com/nocodb/nocodb/issues/3089 

re #3340

Known issue:
- On tab change api is called twice and state related to tab is changed multiple times.

image

Here we are logging on changes on activeTab state in page(`index.vue`) and changes on the `tabs.value?.[activeTabIndex.value]` (activeTab is just a computed of this) (`useTabs.ts`)

### DIFF
--- a/packages/nc-gui-v2/composables/useViewData.ts
+++ b/packages/nc-gui-v2/composables/useViewData.ts
@@ -10,6 +10,7 @@ import {
   getHTMLEncodedText,
   useProject,
   useUIPermission,
+  useApi
 } from '#imports'
 
 const formatData = (list: Record<string, any>[]) =>
@@ -38,8 +39,7 @@ export function useViewData(
     throw new Error('Table meta is not available')
   }
 
-  const loading = ref(false)
-  const error = ref<any>()
+  const { api, isLoading, error } = useApi()
   const _paginationData = ref<PaginatedType>({ page: 1, pageSize: 25 })
   const aggCommentCount = ref<{ row_id: string; count: number }[]>([])
   const galleryData = ref<GalleryType>()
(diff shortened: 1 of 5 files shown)


packages/nc-gui-v2/composables/useViewData.ts
```diff
@@ -38,6 +38,8 @@ export function useViewData(
     throw new Error('Table meta is not available')
   }
 
+  const loading = ref(false)
```
bcakmakoglu, line 41: instead of using `$api` from the global state you can just use `const { api, isLoading } = useApi()` :)
Keeps your composable free of unnecessary refs.

`useApi` also provides an error ref that you can use.

bcakmakoglu (changes requested):

packages/nc-gui-v2/composables/useViewData.ts
```diff
@@ -114,14 +116,13 @@ export function useViewData(
...
     if ((!project?.value?.id || !meta?.value?.id || !viewMeta?.value?.id) && !isPublic.value) return
-
     const response = !isPublic.value
-      ? await $api.dbViewRow.list('noco', project.value.id!, meta.value.id!, viewMeta!.value.id, {
-          ...params,
-          ...(isUIAllowed('sortSync') ? {} : { sortArrJson: JSON.stringify(sorts.value) }),
-          ...(isUIAllowed('filterSync') ? {} : { filterArrJson: JSON.stringify(nestedFilters.value) }),
-          where: where?.value,
-        })
+      ? await api.dbViewRow.list('noco', project.value.id!, meta.value.id!, viewMeta!.value.id, {
```
wingkwong, line 120: if we are using `api` from `useApi`, I believe all `$api` from `useNuxtApp` should also be replaced as well?

bcakmakoglu: If the api call is supposed to trigger `isLoading`, then yea, `$api` should be replaced. :)

pranavxc: Only for the list API we need the spinner

bcakmakoglu (approved):

PR #3349, opened 2022-08-25
fix: shared URL displayed in shared list

dstala:
Signed-off-by: Raju Udava <86527202+dstala@users.noreply.github.com>

## Change Summary
Remove additional `/`
  
Screenshot 2022-08-25 at 7 03 22 PM

## Change type

## Test/ Verification
Locally verified

### DIFF
--- a/packages/nc-gui-v2/components/smartsheet-toolbar/SharedViewList.vue
+++ b/packages/nc-gui-v2/components/smartsheet-toolbar/SharedViewList.vue
@@ -60,7 +60,7 @@ const sharedViewUrl = (view: SharedViewType) => {
     default:
       viewType = 'view'
   }
-  return `nc/${viewType}/${view.uuid}`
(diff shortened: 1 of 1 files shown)


packages/nc-gui-v2/components/smartsheet-toolbar/SharedViewList.vue
```diff
@@ -60,7 +60,7 @@ const sharedViewUrl = (view: SharedViewType) => {
     default:
       viewType = 'view'
   }
-  return `/nc/${viewType}/${view.uuid}`
+  return `nc/${viewType}/${view.uuid}`
```
pranavxc, line 63: You can combine `${dashboardUrl?.value as string}/....` here

dstala: @pranavxc check if the changes are as you had intended.

PR #3358, opened 2022-08-26
API unit tests

mustafapc19:
Ref: #3579

By default tries to connect mysql with (user: `root` and password: `password`) and fall backs to sqlite if errors out. But some test will fail under sqlite.


packages/nocodb/tests/unit/factory/project.ts
```diff
@@ -0,0 +1,67 @@
...
+    });
+};
+
+const createSakilaProject = async (context) => {
+  const response = await request(context.app)
+    .post('/api/v1/db/meta/projects/')
+    .set('xc-auth', context.token)
+    .send(externalProjectConfig);
+
+  const project: Project = await Project.getByTitleOrId(response.body.id);
```
wingkwong, line 52: `project` will not be used. I think you can just write

```ts
return await Project.getByTitleOrId(response.body.id);
```

packages/nocodb/tests/unit/factory/project.ts
```diff
@@ -0,0 +1,67 @@
...
+  return project;
+};
+
+const createProject = async (context, projectArgs = defaultProjectValue) => {
+  const response = await request(context.app)
+    .post('/api/v1/db/meta/projects/')
+    .set('xc-auth', context.token)
+    .send(projectArgs);
+
+  const project: Project = await Project.getByTitleOrId(response.body.id);
```
wingkwong, line 63: same as above

packages/nocodb/src/lib/Noco.ts
```diff
@@ -81,10 +81,12 @@ export default class Noco {
       return Noco._this.router;
     }
     Noco._this = new Noco();
+    Noco.initialized = true;
```
wingkwong, line 84: where will we use this value?

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -143,7 +144,6 @@ class BaseModelSqlv2 {
           is_group: true,
           logical_op: 'and',
         }),
-        ...(args.filterArr || []),
```
wingkwong, line 146: what is the reason to remove this line?

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -1775,21 +1777,25 @@ class BaseModelSqlv2 {
...
           this.dbDriver
         );
+
         qb.update(updateData);
-        res = ((await qb) as any).count;
+        queryResponse = (await qb) as any;
       }
-      return res;
+
+      const count = queryResponse.count || queryResponse;
```
wingkwong, line 1789: what would `queryResponse` look like if `count` is not in it?

wingkwong (commented):
Here's my comments.

- tests/mysql-dump/* looks like exactly same as tests/mysql-sakila-db/*. any differences?
- when to write tests under `src/__tests__/unit` and when to write tests under `src/tests/unit`?
- prefer to include some explanations on the structure of `src/tests/unit` and some general steps for a beginner to write tests.

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
pranavxc: @wingkwong - which is not required since it's a repetition, copy of ```new Filter({
          children: args.filterArr || [],
          is_group: true,
          logical_op: 'and',
        })```

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -1667,8 +1666,10 @@ class BaseModelSqlv2 {
     datas: any[],
     {
       chunkSize: _chunkSize = 100,
+      cookie,
```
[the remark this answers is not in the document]
mustafapc19: @wingkwong Cookie stores the `req` object. Its used for getting user info for creating an audit.

mustafapc19:
@wingkwong Moving on, we will be writing unit tests under `tests/unit` and will be removing the `__tests__` after a while. 

Will be adding a how to guide and how test env working in our docs.

I am not sure why there is `tests/mysql-dump/`, but sql files added in`tests/mysql-sakila-db/*`  in this PR are different. It seeds `test_sakila` db instead of `Sakila` db

o1lab:
Looks like this is close to complete. Any more comments @mertmit @pranavxc ? We should get this in.

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -1861,7 +1875,7 @@ class BaseModelSqlv2 {
     await this.handleHooks('After.insert', data, req);
     // if (req?.headers?.['xc-gui']) {
     const id = this._extractPksValues(data);
-    Audit.insert({
+    await Audit.insert({
```
pranavxc, line 1878: you can remove the await here since we don't want to wait for the audit insert

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -1907,7 +1975,7 @@ class BaseModelSqlv2 {
   public async afterDelete(data: any, _trx: any, req): Promise<void> {
     // if (req?.headers?.['xc-gui']) {
     const id = req?.params?.id;
-    Audit.insert({
+    await Audit.insert({
```
pranavxc, line 1978: Same as above

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
mustafapc19: @pranavxc It will fail some test, since we are resetting sakila in test env based on this, at least in the case of test we need the await.

pranavxc: See if you can manage it with a delay

mustafapc19: @pranavxc We can to await only if env is test maybe? But will clutter the code base

pranavxc: If it's using for data verification then you can directly check it in the database

mertmit (approved):

Issue #3357, opened 2022-08-26
vue3: formview - unable to set `E-mail me at` option


dstala:
Even after SMPT configurations are made

PR #3364, opened 2022-08-27
fix(gui-v2): email me in form view

wingkwong:
## Change Summary

- frontend part : ref: #3357
- backend part: fix broken logic in v1

## Change type

## Test/ Verification

image


dstala (changes requested):
@wingkwong `email me` option works even after we reset SMTP configuration. should work only when we have an active SMTP configuration.

pranavxc (approved):

packages/nocodb/src/lib/meta/helpers/webhookHelpers.ts
```diff
@@ -268,16 +267,11 @@ export function _transformSubmittedFormDataForEmail(
   // @ts-ignore
   formView,
   // @ts-ignore
-  columns: Column[]
+  columns: Record<string, any>[]
```
pranavxc, line 270: The type could be ColumnType.

wingkwong: `columns` includes `Column[]` and some extra fields.`ColumnType` shouldn't be used as well.

pranavxc: Then you can do `(Column &  { additionalTypes : Type })[]`

dstala (changes requested):
broken. unable to enable `email me` option after configuring SMTP
I tried over SQLite xcdb project

wingkwong:
@dstala Got the cache problem. Should be good now.

image

dstala (approved):

Issue #3305, opened 2022-08-23
vue3: gallery view re-work items


dstala:
Use the following template import (MySQL XCDB project)
https://airtable.com/shr2Hyeoyz7gt1daE
keyn1MR87qgyUsYg4

Open `Video tracker` table `video thumbnails` gallery view

Observations
- [ ] Limit to only 3 fields + 1 image by default; hide rest of columns
- [ ] Gallery card width should have minimum width configured. They become too narrow with text overflowing outside their box space
  
![Screenshot 2022-08-22 at 11 24 08 AM](https://user-images.githubusercontent.com/86527202/185848976-cbecc2f7-b5f9-4a2a-a837-c7fef514bad7.png)

- [ ] Column type collaborator, wrap text not enabled (wider view)
  
![Screenshot 2022-08-22 at 11 25 30 AM](https://user-images.githubusercontent.com/86527202/185849182-c8e36cb4-455b-47d3-913b-379043de0316.png)

- [ ] Error observed on click for expanding the row; similar error observed in grid view when trying to open expand-row for m2m columns
  
![Screenshot 2022-08-22 at 11 26 56 AM](https://user-images.githubusercontent.com/86527202/185849335-c43b469b-bb53-467a-b6ef-a611fbd020da.png)

- [ ] Empty cell, retain it as empty (until we figure out another alternative, as discussed over the call)
  
![Screenshot 2022-08-22 at 11 28 48 AM](https://user-images.githubusercontent.com/86527202/185849625-a16e9fd8-a985-49cc-87ee-134bf56e5970.png)

- [ ] `Add new row` broken. need to rebase?
- [ ] Fields hide/unhide, reorder took considerable amount of time

PR #3379, opened 2022-08-28
feat: gallery rework

mertmit:
## Change Summary

Re #3305

## Change type

image


mustafapc19:
@mertmit I don't see a setting  cover image option. In v1 it was under fields option in gallery view which imo also doesn't feel intuitive. Took me a while to find that out. 

Also after a setting multiple images in the cover image column and clicking on the next image in the carousal, seems to bug out(outputs same image 4 times in this case) but only for the first time. 
image

Issue #3386, opened 2022-08-29
vue3 : theme returns to default when modals used from left nav drawer


mertmit:
- [x] Returning to default project color
- [x] Badge and theme color should be the same.

PR #3390, opened 2022-08-29
fix: project theme

mertmit:
## Change Summary

Re #3386 + Pick project theme color from badge

## Change type

image


o1lab:
Works. Fixes issue that were there previously.

o1lab:
Having a hover effect on the color picker button in project list would make this better. It is hard now to realise - that color picker could be clicked to change colors.

bcakmakoglu (approved):

Issue #3459, opened 2022-09-03
vue3: theme palette doesn't work at times


dstala:
The custom theme color picker worked. 
in the color palette, does first row represent `primary` color and second `accent`? or it's just all `primary`? this isn't very clear currently

wingkwong:
both row is for primary color.

mertmit:
Changed theme handling on PR, for ux part:
Color palette actually works as predefined themes, we can use something like this to make it clear that it is going to change both primary and accent
![image](https://user-images.githubusercontent.com/59797957/188288384-872c52f0-1920-4a96-b4ef-22259815312f.png)
**Or like this (accent is easier to recognize)**
![image](https://user-images.githubusercontent.com/59797957/188288481-3527c128-619b-4811-aa77-f154cdc817d5.png)

Waiting for feedback to apply this

dstala:
I understand now. Themes are with a pre-defined combination of primary and accent colors. The current representation is ok IMO.

PR #3481, opened 2022-09-03
fix: theme palette

mertmit:
## Change Summary

Re #3459
- Wrapper component for Chrome (ChromeWrapper)
  - This emits input on color change and usable without v-model (broken for vue-color library we use)
  - Allows returning of any color mode vue-color supports (hex, hex8, hsl, hsv, rgba) instead of full object
- Handle theme colors using events emitted and functions instead of watchers
- Change ColorPicker to use ChromeWrapper (returns hex8 easier to handle / cleaner code)
- SaveTheme allows either only primaryColor or accentColor and uses theme color for missing prop

## Change type


dstala (approved):

dstala:
@pranavxc you can merge post review

Issue #1970, opened 2022-05-04
Synologoy NAS : the 0.90.10 docker container crash


xavier-GitHub76:
Hello

the 0.90.10 docker container crash

docker run nocodb/nocodb

Digest: sha256:5cee7583b5bbe93d79b4ee83d73160480daa697521f2c94bf06a14d45f3e6187
Status: Downloaded newer image for nocodb/nocodb:latest
Knex: run
$ npm install sqlite3 --save
Error relocating /usr/src/app/node_modules/sqlite3/lib/binding/napi-v6-linux-x64/node_sqlite3.node: fcntl64: symbol not found
Error: Error relocating /usr/src/app/node_modules/sqlite3/lib/binding/napi-v6-linux-x64/node_sqlite3.node: fcntl64: symbol not found
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:1057:18)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/usr/src/app/node_modules/sqlite3/lib/sqlite3-binding.js:4:15)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
App started successfully.
Visit -> http://localhost:8080undefined

Regards

o1lab:
Can you please provide more details - like is it on windows or linux ?

and try `docker run -d -p 8080:8080 nocodb/nocodb`

xavier-GitHub76:
I have this error on my Synology NAS and on my personnel computer (Linux)

I try docker run -d -p 8080:8080 nocodb/nocodb
http://localhost:8080/
=> I have a blanck page with just "Cannot GET /"

o1lab:
We do not have access to NAS you are mentioning. Some of users have used it without a problem with NAS before. See [if this helps](https://github.com/nocodb/nocodb/tree/develop/scripts/setups/TrueNAS).

o1lab:
Please also paste your NAS details (version etc)

mertmit:
[...]

PR #3570, opened 2022-09-09
feat: upgrade knex

mertmit:
## Change Summary

Re #3582 
- Upgrade knex to v2.2.0
- Install tedious as mssql client for knex
- Upgrade sqlite3 to v5.1.12
  - Dockerfile
    - upgraded builder to node:16.17.0-alpine3.15
    - upgraded runner to alpine:3.15
    - install node-gyp dependencies (required with upgrades)
- SQLite drop column -> use new alter column instead of alter table
- MSSQL fixed custom schema handling

## Change type


packages/nocodb/src/lib/db/sql-client/lib/mssql/MssqlClient.ts
```diff
@@ -356,9 +356,7 @@ class MssqlClient extends KnexClient {
 
     try {
       /** ************** START : create _evolution table if not exists *************** */
-      const exists = await this.sqlClient.schema.hasTable(
-        this.getTnPath(args.tn)
-      );
+      const exists = await this.sqlClient.schema.hasTable(args.tn);
```
wingkwong, line 359: what's the reason to remove `getTnPath` here?

packages/nocodb/src/lib/db/sql-client/lib/mssql/MssqlClient.ts
```diff
@@ -396,9 +394,7 @@ class MssqlClient extends KnexClient {
     log.api(`${_func}:args:`, args);
 
     try {
-      result.data.value = await this.sqlClient.schema.hasTable(
-        this.getTnPath(args.tn)
-      );
+      result.data.value = await this.sqlClient.schema.hasTable(args.tn);
```
wingkwong, line 397: what's the reason to remove getTnPath here?

packages/nocodb/src/lib/db/sql-client/lib/sqlite/SqliteClient.ts
```diff
@@ -1527,14 +1528,12 @@ class SqliteClient extends KnexClient {
...
-          //   oldColumn,
-          //   downQuery,
-          //             this.sqlClient
-          // );
+          downQuery += this.alterTableRemoveColumn(
+            args.table,
+            args.columns[i],
+            oldColumn,
+            downQuery
+          );
```
o1lab, line 1536: SQLite down query should not be possible.

packages/nocodb/src/lib/db/sql-client/lib/sqlite/SqliteClient.ts
```diff
@@ -1956,19 +1945,31 @@ class SqliteClient extends KnexClient {
...
+      let q2 = ''
+      q2 += this.genQuery(` ADD ?? ${n.dt}`, [n.cn], shouldSanitize);
+      q2 += n.dtxp && n.dt !== 'text' ? `(${n.dtxp})` : '';
+      q2 += n.cdf ? (n.cdf.includes(',') ? ` DEFAULT ('${n.cdf}')` : ` DEFAULT ${n.cdf}`) : ' ';
+      q2 += n.rqd ? ` NOT NULL` : ' ';
+      q2 = this.genQuery(`ALTER TABLE ?? ${q2};`, [t], shouldSanitize);
+
+      let q3 = this.genQuery(`UPDATE ?? SET ?? = ??;`, [t, n.cn, `${o.cno}_old`], shouldSanitize);
+
+      let q4 = this.genQuery(`ALTER TABLE ?? DROP COLUMN ??;`, [t, `${o.cno}_old`], shouldSanitize);
```
o1lab, line 1963: What does this `_old` hold ? was this part of our meta earlier. If not, we need to see its documented and uniform across other clients. Can discuss offline.

mertmit: Because of the limitations with alter on SQLite on first query we rename original column to `colName`_old then we recreate column and update all data using this backup column on next query (q4) we drop it.

Also I realized there should be a pragma foreign keys on | off surrounding the queries to support foreign keys, I will handle this situation as well.

o1lab:
@mertmit : pls lets get this in tomorrow after you resolve the comments.

packages/nocodb/src/lib/db/sql-client/lib/sqlite/SqliteClient.ts
```diff
@@ -1544,22 +1543,18 @@ class SqliteClient extends KnexClient {
...
       );
-      //downQuery += alterTablePK(args.originalColumns, args.columns, downQuery);
-
-      if (upQuery) {
-        //upQuery = `ALTER TABLE ${args.columns[0].tn} ${upQuery};`;
-        //downQuery = `ALTER TABLE ${args.columns[0].tn} ${downQuery};`;
-      }
 
-      await Promise.all(
+      if (synchronous) {
```
pranavxc, line 1547: Since `await Promise.all(` is removed the catch part won't handle the exceptions that happened within the map callback function.

On exception it may print `Uncaught Error`.


You can try the following to reproduce the  bug.


```js

(async () => {

  // without promise.await
  try {
    [1,2,3].map(async () =>{
      throw new Error('test')
    })
  }catch(e){
    console.log(e)
  }

  
  // with promise.await
  try {
    await Promise.all([1,2,3].map(async () =>{
      throw new Error('test')
    }))
  }catch(e){
    console.log(e)
  }
```

mertmit: I changed this to a for loop which seems to act same with Promise.all

pranavxc: It will run sequentially with for loop

pranavxc: looks good now :+1:

packages/nocodb/src/lib/db/sql-client/lib/mssql/MssqlClient.ts
mertmit: Thank you for pointing out, this looks like a wrong change

mertmit:
@wingkwong I dropped getTnPath for all schema specific queries because schema.hasTable returns wrong result for [schema].[table], other functions seem to work (both syntax) but I dropped them as well to keep a convention so all of the raw queries use getTnPath, none of the schema queries use it.

wingkwong:
@mertmit how does knex know which schema should be used? (let's say users define a new one)

mertmit:
Passed all the tests, builds are working.
Ready for review: @pranavxc @wingkwong (bumping as this left behind)

packages/nocodb/src/lib/db/sql-client/lib/mssql/MssqlClient.ts
```diff
@@ -1599,9 +1595,9 @@ class MssqlClient extends KnexClient {
     const result = new Result();
     log.api(`${func}:args:`, args);
     try {
-      const query = `CREATE TRIGGER ${args.trigger_name} on ${this.getTnPath(
+      const query = `CREATE TRIGGER [${args.trigger_name}] on [${this.getTnPath(
         args.tn
-      )} \n${args.timing} ${args.event}\n as\n${args.statement}`;
+      )}] \n${args.timing} ${args.event}\n as\n${args.statement}`;
```
pranavxc, line 1600: You can replace it with `genQuery`

```js
this.genQuery(`CREATE TRIGGER ?? on ?? \n${args.timing} ${args.event}\n as\n${args.statement}`, [args.trigger_name, this.getTnPath(
        args.tn
      )])
```

packages/nocodb/src/lib/db/sql-client/lib/mssql/MssqlClient.ts
```diff
@@ -1853,11 +1849,11 @@ class MssqlClient extends KnexClient {
         const triggerName = `xc_trigger_${args.table_name}_${column.column_name}`;
         const triggerCreateQuery =
           this.querySeparator() +
-          `CREATE TRIGGER ${this.schema}.${triggerName} ON ${this.schema}.${args.table_name} AFTER UPDATE
+          `CREATE TRIGGER [${this.schema}].[${triggerName}] ON [${this.schema}].[${args.table_name}] AFTER UPDATE
                   AS
                   BEGIN
                       SET NOCOUNT ON;
-                      UPDATE ${this.schema}.${args.table_name} Set ${column.column_name} = GetDate() where ${pk.column_name} in (SELECT ${pk.column_name} FROM Inserted)
+                      UPDATE [${this.schema}].[${args.table_name}] Set [${column.column_name}] = GetDate() where [${pk.column_name}] in (SELECT [${pk.column_name}] FROM Inserted)
```
pranavxc, line 1856: Use genQuery

packages/nocodb/src/lib/db/sql-client/lib/pg/PgClient.ts
```diff
@@ -2102,7 +2102,7 @@ class PGClient extends KnexClient {
           this.genQuery(`DROP FUNCTION IF EXISTS ??()`, [triggerFnName]);
       }
     }
-    await this.sqlClient.raw(upQuery);
+    if (upQuery != '') await this.sqlClient.raw(upQuery);
```
pranavxc, line 2105: Replace `!=` with `!==`

pranavxc (changes requested):

pranavxc:
@mertmit please resolve conflicts

packages/nocodb/package.json
```diff
@@ -121,11 +121,13 @@
     "passport-jwt": "^4.0.0",
     "passport-local": "^1.0.0",
     "pg": "^8.3.0",
+    "request": "^2.88.2",
     "request-ip": "^2.1.3",
     "rmdir": "^1.2.0",
     "slash": "^3.0.0",
     "socket.io": "^4.4.1",
-    "sqlite3": "5.0.0",
+    "sqlite3": "5.1.2",
```
pranavxc, line 129: Issue related to sqlite upgrade https://github.com/nocodb/nocodb/issues/1970 , will test the docker build later in linux and confirm thee same :+1:

pranavxc: #### Verified docker in:

- [ ] Ubuntu
- [ ] MacOS
- [ ] Windows 

#### Verified executable in:

- [ ] Ubuntu
- [ ] MacOS
- [ ] Windows

pranavxc: Executable failed in windows with the following error 

image

pranavxc: Probably we need to add binaries for the platform here - https://github.com/nocodb/nocodb-timely/tree/main/binaries/binding

pranavxc: Executable failing in mac as well

Screenshot 2022-11-04 at 12 58 30 PM

mertmit: Should be resolved with merge of @pranavxc 's PR

pranavxc: yes 👍🏽

packages/nocodb/src/lib/db/sql-client/lib/mssql/MssqlClient.ts
```diff
@@ -2144,6 +2143,132 @@ class MssqlClient extends KnexClient {
...
+  /**
+   *
+   * @param {Object} - args
+   * @param {String} - args.parentTable
+   * @param {String} - args.parentColumn
+   * @param {String} - args.childColumn
+   * @param {String} - args.childTable
+   * @returns {Promise<{upStatement, downStatement}>}
+   */
+   async relationCreate(args) {
```
bcakmakoglu, line 2155: Let me chime in real quick.

These JSDocs are incorrect as shown by Webstorm:
image

`Parameter args is not described in JSDoc `

And it's kind of unnecessary to use JSDocs for typing when we're already using TypeScript that handles that.
It even limits the typing because you describe return values as `Object` when in reality they aren't just `Object` but something else, for example `Result`.

Imo: Remove the JSDocs comments.

mertmit: Thank you for comment. I agree those feel like duplicates with ts types.
I derived JSDocs from KnexClient. As those are mostly implemented by Pranav, @pranavxc what do you think on this?

Note: even if we decide removing/updating them, it is better we create a separate chore PR and get this one in as this PR is already big enough

pranavxc: We added them when the entire project was in Javascript, since we migrated to Typescript we can remove them. We can do the cleanup part in a separate PR.

pranavxc (approved):

wingkwong (approved):

PR #3606, opened 2022-09-13
refactor(nc-gui): defer loading project to subpage

bcakmakoglu:
## Change Summary

* defer loading to subpage
* unblock smartsheet page and show loading spinner while content is loading
* defer adding tabs until project was loaded

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/composables/useSharedView.ts
+++ b/packages/nc-gui/composables/useSharedView.ts
@@ -1,4 +1,4 @@
-import type { ExportTypes, FilterType, PaginatedType, SortType, TableType, ViewType } from 'nocodb-sdk'
+import type { ExportTypes, FilterType, PaginatedType, RequestParams, SortType, TableType, ViewType } from 'nocodb-sdk'
 import { UITypes } from 'nocodb-sdk'
 import { useNuxtApp } from '#app'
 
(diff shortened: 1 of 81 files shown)


packages/nc-gui/composables/useSharedView.ts
```diff
@@ -58,7 +60,7 @@ export function useSharedView() {
     const pageSize = paginationData.value.pageSize || 25
 
     const { data } = await $api.public.dataList(
-      sharedView?.value?.uuid,
+      (sharedView.value as any)?.uuid,
```
pranavxc, line 63: Could it be `as ViewType`?

bcakmakoglu: Nope because ViewType is missing `uuid` otherwise it would work without the type hack :(

pranavxc: Yes it have `uuid`

```ts
export interface ViewType {
  id?: string;
  title: string;
  deleted?: boolean;
  order?: number;
  fk_model_id?: string;
  slug?: string;
  uuid?: string;
  show_system_fields?: boolean;
  lock_type?: 'collaborative' | 'locked' | 'personal';
  type?: number;
  view?: FormType | GridType | GalleryType;
}
```

Rebuild `nocodb-sdk` and check it out.

packages/nc-gui/pages/index/index/index.vue
```diff
@@ -175,18 +175,18 @@ const getProjectPrimary = (project: ProjectType) => {
     </div>
 
     <TransitionGroup name="layout" mode="out-in">
-      <div v-if="isLoading">
+      <div v-if="isLoading" key="skeleton">
```
pranavxc, line 178: Why there is a static key for these elements and any specific reason to keep them?

pranavxc (changes requested):

packages/nc-gui/pages/index/index/index.vue
bcakmakoglu: Yeah Vue's `transition-group` requires a key on elements :)

packages/nc-gui/composables/useSharedView.ts
bcakmakoglu: Ah okay, I missed the update then, will revert on those changes ^^

pranavxc (approved):

bcakmakoglu:
@dstala I am not sure if the test fail because of an issue here :s
Could you possibly check if that is the case or maybe even do small manual check if everything seems fine?
On my end everything works well, that's why I am a bit confused

wingkwong (approved):

dstala (approved):

Issue #3599, opened 2022-09-13
Photos in gallery view are "squeezed"


WeT-Klb:
**Please enter the following details**
Node: v12.22.6
Arch: x64
Platform: linux (QNAP 5.0.0.2131)
Docker: true
Database: mysql2 (MariaDB10.5)
ProjectOnRootDB: true
RootDB: mysql2 (MariaDB 10.5)
PackageVersion: 0.96.4```

**Steps To Reproduce**

1. Go to gallery view
2. The picture on top of the column has more length than height
3. So all pictures looks quite distorted

**Expected Behavior**

The pictures should be square

**Screenshots**

![Nocodb](https://user-images.githubusercontent.com/20600342/189836948-8af3cdc3-d5ff-4267-82f4-ff72fd2df88e.jpg)

Join our discord : https://discord.gg/5RgZmkW for realtime help.

WeT-Klb:
Hi,

with 0.97.0 still the picture looks not right. i suggest to make the pics square - just like in the table itself.
Otherwise always something will be cut off the pic.
Before 0.96.0 i think it was quite perfect - but only my opinion.

![Gallery View neu](https://user-images.githubusercontent.com/20600342/191691297-7fd76142-ad56-470e-86e1-0905b7d4fe96.jpg)

o1lab:
Ah good.

PR #3618, opened 2022-09-13
fix: gallery card cover image distortion

mertmit:
## Change Summary

Re #3599 
Use object-cover for gallery card cover image

## Change type

Screenshot_2022-09-14_01-43-30

As we use object-cover some parts of the image might be missed, an alternative approach would have been using object-contain but it results with extra space for most aspect ratios.

Example:
Screenshot_2022-09-14_01-44-06

Open for any suggestion / alternative


o1lab:
Will test it with few more sizes

PR #3637, opened 2022-09-14
fix(nc-gui): project state not properly reset 

### LINKED ISSUE #3635: new project displays deleted project table name's
Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
Database: **sqlite3**
ProjectOnRootDB: **true**
RootDB: **sqlite3**
PackageVersion: **0.96.4**

also observed in pg/pg
also observed with disabled cache

- create project
- add table `Sheet1`
- delete project
- create a new project
- table tree view displays `Sheet1`

bcakmakoglu:
## Change Summary

* fixes #3635 
* only load project if no request is running already
* move fetching project to parent page to prevent multiple loads
* defer subpages until project is loaded
   * show spinner when loading
* reset state when page is unmounted

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/composables/useProject.ts
+++ b/packages/nc-gui/composables/useProject.ts
@@ -1,31 +1,53 @@
 import type { MaybeRef } from '@vueuse/core'
-import { SqlUiFactory } from 'nocodb-sdk'
 import type { OracleUi, ProjectType, TableType } from 'nocodb-sdk'
-import { useNuxtApp, useRoute } from '#app'
-import type { ProjectMetaInfo } from '~/lib'
+import { SqlUiFactory } from 'nocodb-sdk'
+import { isString } from '@vueuse/core'
+import {
+  USER_PROJECT_ROLES,
+  computed,
+  createEventHook,
+  ref,
+  useApi,
+  useGlobal,
+  useInjectionState,
+  useNuxtApp,
+  useRoute,
+  useState,
+  useTheme,
+  watch,
+} from '#imports'
+import type { ProjectMetaInfo, Roles } from '~/lib'
 import type { ThemeConfig } from '@/composables/useTheme'
-import { createEventHook, useInjectionState } from '#imports'
 
 const [setup, use] = useInjectionState((_projectId?: MaybeRef<string>) => {
-  const { $api, $e } = useNuxtApp()
+  const { $e } = useNuxtApp()
+
+  const { api, isLoading } = useApi()
+
   const route = useRoute()
+
   const { includeM2M } = useGlobal()
+
   const { setTheme, theme } = useTheme()
 
   const projectLoadedHook = createEventHook<ProjectType>()
 
-  const projectId = computed(() => (_projectId ? unref(_projectId) : (route.params.projectId as string)))
   const project = ref<ProjectType>({})
+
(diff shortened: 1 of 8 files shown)


packages/nc-gui/composables/useProject.ts
```diff
@@ -1,31 +1,53 @@
...
+} from '#imports'
+import type { ProjectMetaInfo, Roles } from '~/lib'
 import type { ThemeConfig } from '@/composables/useTheme'
-import { createEventHook, useInjectionState } from '#imports'
 
 const [setup, use] = useInjectionState((_projectId?: MaybeRef<string>) => {
-  const { $api, $e } = useNuxtApp()
+  const { $e } = useNuxtApp()
+
+  const { api, isLoading } = useApi()
```
pranavxc, line 25: You are using the same `api` reference for loading the table list as well, so if you try to reload the table list(context menu in the tree view) it may lead to some issues I think.

bcakmakoglu: What's the difference between using `$api` and `api` in this case though?
It would still be only one api instance that's used to fetch the data, regardless if it's the global one or a new one locally.

If there is/was an issue with the reload, it was a pre-existing condition 😄

pranavxc: > What's the difference between using `$api` and `api` in this case though? It would still be only one api instance that's used to fetch the data, regardless if it's the global one or a new one locally.
> 
> If there is/was an issue with the reload, it was a pre-existing condition 😄

We are using the `isLoading` variable for toggling between loader and spinner.

```vue
 <div class="w-full min-h-[300px] flex-auto">
        <NuxtPage v-if="!isLoadingProject" />
        <div v-else class="w-full h-full flex justify-center items-center">
          <a-spin size="large" />

```

bcakmakoglu: global isLoading is *always* triggered if an api instance does a request.

So if `const { isLoading } = useApi()` is `true` then `const { isLoading } = useGlobal()` is true as well, at least that is the intended way ^^

mertmit: I can also confirm I used useNuxtApp -> $api to not trigger isLoading state while using API. (on project list page, color picker api, as pageLoading for every call was not ideal)

pranavxc: I think earlier I'd used `useApi()` only for loading project and for rest of them the global `$api` is used

bcakmakoglu: As I suspected, the issue(s) come from something else. Will push changes in a bit ^^

dstala (approved):

packages/nc-gui/composables/useProject.ts
bcakmakoglu: Fix pushed, I had to play around a bit to get it right
* Subpages should now wait (and show a spinner) until the parent pages are done fetching the project/tables etc.
* The reload issue in the video came from using `v-if` on a NuxtPage component, which doesn't work, when the condition turns from true to false, it crashes the app

PR #3641, opened 2022-09-14
fix(nc-gui): remove injection usage in callback

bcakmakoglu:
## Change Summary

* update EditOrAdd form style
* remove injection use on callbacks * fixes errors thrown while changing column type
* use `unref`

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/composables/useColumnCreateStore.ts
+++ b/packages/nc-gui/composables/useColumnCreateStore.ts
@@ -82,8 +82,6 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
     const { resetFields, validate, validateInfos } = useForm(formState, validators)
 
     const onUidtOrIdTypeChange = () => {
-      const { isCurrency } = useColumn(ref(formState.value as ColumnType))
-
       const colProp = sqlUi.value.getDataTypeForUiType(formState.value as { uidt: UITypes }, idType ?? undefined)
       formState.value = {
         ...formState.value,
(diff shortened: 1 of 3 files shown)


packages/nc-gui/composables/useColumnCreateStore.ts
```diff
@@ -82,8 +82,6 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
     const { resetFields, validate, validateInfos } = useForm(formState, validators)
 
     const onUidtOrIdTypeChange = () => {
-      const { isCurrency } = useColumn(ref(formState.value as ColumnType))
```
bcakmakoglu, line 85: This causes problems as `useColumn` uses `inject` which it can't trigger in a callback - `provide` and `inject` can only be used inside the body of a setup function

So `isCurrency.value` probably is always false currently (also there's a double check for it below

packages/nc-gui/composables/useColumnCreateStore.ts
```diff
@@ -111,14 +109,12 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
         }
       }
 
-      if (isCurrency.value) {
-        if (column.value?.uidt === UITypes.Currency) {
```
bcakmakoglu, line 115: These check for the same value - so the second if clause can never be true

packages/nc-gui/composables/useColumnCreateStore.ts
```diff
@@ -111,14 +109,12 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
...
-          formState.value.dtxp = column.value.dtxp
-          formState.value.dtxs = column.value.dtxs
-        } else {
-          formState.value.dtxp = 19
-          formState.value.dtxs = 2
```
bcakmakoglu, line 120: This can never be true

mustafapc19 (approved):

PR #3680, opened 2022-09-16
refactor(nc-gui): remove keys from erd views

bcakmakoglu:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.


packages/nc-gui/components/smartsheet-toolbar/Erd.vue
```diff
@@ -22,6 +22,7 @@ const selectedView = inject(ActiveViewInj)
     width="max(900px,60vw)"
     :closable="false"
     wrap-class-name="erd-single-table-modal"
+    transition-name="fade"
```
bcakmakoglu, line 25: @mustafapc19 Using the fade transition instead of the default `zoom` solves the issue of inconsistent values for `getBoundingClientRect` on initial render as fade animation does not use any transform values which cause the problem

bcakmakoglu: Here are the different transitions btw: https://github.com/vueComponent/ant-design-vue/tree/main/components/style/core/motion

mustafapc19 (approved):

Issue #3702, opened 2022-09-19
Table view is empty on initial render


mustafapc19:
Node: **v16.15.1**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
Database: **sqlite3**
ProjectOnRootDB: **true**
RootDB: **sqlite3**
PackageVersion: **0.96.4**

**Steps To Reproduce**

1. Open a table.
2. Refresh the page
3. Table view will be empty

**Screenshots**

<img width="1420" alt="image" src="https://user-images.githubusercontent.com/35809690/190991899-914591e1-663c-4c8c-88d3-a554f3a3743b.png">

dstala:
@mustafapc19 
Try if you can reproduce still after cleanup cache files
from packages/nc-gui, delete folders `.nuxt` and `node_modules/.vite`

Restart browser

mustafapc19:
@dstala I have deleted `.nuxt`, `node_modules/.vite`, cleared local storage, cookies and also have restarted the browser but I can still observe this issue. Its happening 50% of the time I refresh the page.

mustafapc19:
<img width="448" alt="image" src="https://user-images.githubusercontent.com/35809690/191012723-c1adfd62-e96c-43d2-b10d-34e0d3811bc5.png">

Seems like state like `ActiveView` and `metas` is undefined

PR #3708, opened 2022-09-19
Fix: Random blank screen bug

pranavxc:
## Change Summary

re #3702

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/components/general/Overlay.vue
+++ b/packages/nc-gui/components/general/Overlay.vue
@@ -11,7 +11,7 @@ interface Props {
   target?: TeleportProps['to']
   teleportDisabled?: TeleportProps['disabled']
   transition?: boolean
-  lightBackground?: boolean
+  className?: string
 }
 
 interface Emits {
@@ -50,10 +50,9 @@ export default {
         vModel ? 'opacity-100' : 'opacity-0 pointer-events-none',
         inline ? 'absolute' : 'fixed',
         transition ? 'transition-opacity duration-200 ease-in-out' : '',
-        lightBackground ? 'bg-gray-100/25' : '',
-        !lightBackground ? 'bg-gray-700/75' : '',
+        className,
       ]"
-      class="z-100 top-0 left-0 bottom-0 right-0 bg-gray-100/25"
+      class="z-100 top-0 left-0 bottom-0 right-0 bg-gray-700/75"
     >
       <slot :is-open="vModel" />
     </div>
--- a/packages/nc-gui/components/general/Overlay.vue
+++ b/packages/nc-gui/components/general/Overlay.vue
@@ -11,6 +11,7 @@ interface Props {
   target?: TeleportProps['to']
   teleportDisabled?: TeleportProps['disabled']
   transition?: boolean
+  lightBackground?: boolean
 }
 
 interface Emits {
@@ -49,8 +50,10 @@ export default {
         vModel ? 'opacity-100' : 'opacity-0 pointer-events-none',
         inline ? 'absolute' : 'fixed',
         transition ? 'transition-opacity duration-200 ease-in-out' : '',
+        lightBackground ? 'bg-gray-100/25' : '',
+        !lightBackground ? 'bg-gray-700/75' : '',
       ]"
-      class="z-100 top-0 left-0 bottom-0 right-0 bg-gray-700/75"
+      class="z-100 top-0 left-0 bottom-0 right-0 bg-gray-100/25"
     >
       <slot :is-open="vModel" />
     </div>
--- a/packages/nc-gui/components/general/Overlay.vue
+++ b/packages/nc-gui/components/general/Overlay.vue
@@ -11,7 +11,6 @@ interface Props {
   target?: TeleportProps['to']
   teleportDisabled?: TeleportProps['disabled']
   transition?: boolean
-  className?: string
 }
 
 interface Emits {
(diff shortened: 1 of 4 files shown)


mustafapc19 (approved):

packages/nc-gui/components/general/Overlay.vue
```diff
@@ -11,6 +11,7 @@ interface Props {
   target?: TeleportProps['to']
   teleportDisabled?: TeleportProps['disabled']
   transition?: boolean
+  lightBackground?: boolean
```
wingkwong, line 14: You've added `lightBackground` in interface but didn't add it in L23 (defineProps). It may get `Unresolved variable or type`.

packages/nc-gui/components/general/Overlay.vue
```diff
@@ -49,8 +50,10 @@ export default {
         vModel ? 'opacity-100' : 'opacity-0 pointer-events-none',
         inline ? 'absolute' : 'fixed',
         transition ? 'transition-opacity duration-200 ease-in-out' : '',
+        lightBackground ? 'bg-gray-100/25' : '',
+        !lightBackground ? 'bg-gray-700/75' : '',
       ]"
-      class="z-100 top-0 left-0 bottom-0 right-0 bg-gray-700/75"
+      class="z-100 top-0 left-0 bottom-0 right-0 bg-gray-100/25"
```
wingkwong, line 56: 1. I'd suggest to use `bg-gray-700/75` by default

packages/nc-gui/components/general/Overlay.vue
```diff
@@ -49,8 +50,10 @@ export default {
         vModel ? 'opacity-100' : 'opacity-0 pointer-events-none',
         inline ? 'absolute' : 'fixed',
         transition ? 'transition-opacity duration-200 ease-in-out' : '',
+        lightBackground ? 'bg-gray-100/25' : '',
```
wingkwong, line 53: 2. so that we can just use `lightBackground?  '!bg-gray-100/25': '',` to override the dark overlay.

packages/nc-gui/components/general/Overlay.vue
bcakmakoglu, line 14: This is unnecessary, you can just add class names to the component already.

If you need a lighter background you can use `!bg-opacity-75` as `class` on the component.

packages/nc-gui/components/general/Overlay.vue
bcakmakoglu, line 56: See comment above

packages/nc-gui/pages/[projectType]/[projectId]/index/index/[type]/[title]/[[viewTitle]].vue
```diff
@@ -15,14 +15,17 @@ const activeTab = inject(
...
-if (!project.value.id) {
-  projectLoadedHook(async () => {
-    await getMeta(route.params.title as string, true)
+/** wait until table list loads since meta load requires table list **/
+watch(
+  () => tables.value.length,
+  (nextVal) => {
+    if (!nextVal) return
+
+    getMeta(route.params.title as string, true).finally(() => (loading.value = false))
```
Comment (bcakmakoglu) [inline packages/nc-gui/pages/[projectType]/[projectId]/index/index/[type]/[title]/[[viewTitle]].vue:24]: This watcher will keep watching - is this intended?
The comment says "wait until" which suggests we only wait once.
If so: You can use the `until` watcher to watch once until tables have loaded :)

bcakmakoglu (changes requested):

packages/nc-gui/components/general/Overlay.vue
wingkwong: Resolved my convos. This approach is more clean. However, I think it would be `!bg-opacity-25` instead assuming the default background is `bg-gray-700/75`.

bcakmakoglu: Depends on the color
if you use a white background and lower the opacity to 25, it will appear like a dark background because white with very low opacity is not super light but rather dark 😄

packages/nc-gui/pages/[projectType]/[projectId]/index/index/[type]/[title]/[[viewTitle]].vue
Comment (pranavxc) [reply packages/nc-gui/pages/[projectType]/[projectId]/index/index/[type]/[title]/[[viewTitle]].vue:24]: What about something like following which will only trigger once

```ts
if (tables.value.length) {
  getMeta(route.params.title as string, true).finally(() => (loading.value = false))
  loading.value = false
} else {
  /** wait until table list loads since meta load requires table list **/
  watchOnce(
    () => tables.value.length,
    (nextVal) => {
      if (!nextVal) return
      getMeta(route.params.title as string, true).finally(() => (loading.value = false))
      loading.value = false
    },
  )
}
```

Comment (bcakmakoglu) [reply packages/nc-gui/pages/[projectType]/[projectId]/index/index/[type]/[title]/[[viewTitle]].vue:24]: That will only watch "once" so even if tables.value.length returns undefined in the watch, it will stop the watcher fully.
Instead using `until` you could check if the length is not 0 and then trigger the watcher once and be done with it =)

https://vueuse.org/shared/until/

Comment (pranavxc) [reply packages/nc-gui/pages/[projectType]/[projectId]/index/index/[type]/[title]/[[viewTitle]].vue]: length will never be `undefined`

packages/nc-gui/components/general/Overlay.vue
```diff
@@ -11,6 +11,7 @@ interface Props {
   target?: TeleportProps['to']
   teleportDisabled?: TeleportProps['disabled']
   transition?: boolean
+  className?: string
```
bcakmakoglu, line 14: You don't need an extra prop for class 😄 

```html
<div>
	 <div v-bind="$attrs">Foo</div>
</div>
```

$attrs will contain the `class` attribute already so they're bound to the div in the overlay.

Technically the prop you added does nothing right now ^^

bcakmakoglu (changes requested):

packages/nc-gui/components/general/Overlay.vue
mustafapc19: Habits from react 😅

PR #3715, opened 2022-09-20
Fix(nocodb): Audit insert will always await in test env while in normal environments will by default not await

mustafapc19:
Behavior:
- All `await Audit.insert` will be blocking/awaiting in test environment.
- All `await Audit.insert` will be non blocking in non test environment.

Rational being in test environment, data reset is being done using Audit table. Thus we need to await there but for other environments, this awaiting will slow things down.

### DIFF
--- a/packages/nocodb/src/lib/models/Audit.ts
+++ b/packages/nocodb/src/lib/models/Audit.ts
@@ -69,8 +69,7 @@ export default class Audit implements AuditType {
     };
 
     if (forceAwait) {
-      const auditRec = await insertAudit();
-      return auditRec;
+      return await insertAudit();
     } else {
       insertAudit();
     }
--- a/packages/nocodb/src/lib/models/Audit.ts
+++ b/packages/nocodb/src/lib/models/Audit.ts
@@ -44,13 +44,13 @@ export default class Audit implements AuditType {
       forceAwait: process.env['TEST'] === 'true',
     }
   ) {
-    if (!audit.project_id && audit.fk_model_id) {
-      audit.project_id = (
-        await Model.getByIdOrName({ id: audit.fk_model_id }, ncMeta)
-      ).project_id;
-    }
-
     const insertAudit = async () => {
+      if (!audit.project_id && audit.fk_model_id) {
+        audit.project_id = (
+          await Model.getByIdOrName({ id: audit.fk_model_id }, ncMeta)
+        ).project_id;
+      }
+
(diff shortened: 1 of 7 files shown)


o1lab:
Better to have some some wait in tests. We dont want to slow the updates.

mustafapc19:
Will add delay and remove all the awaits if its an issue

o1lab:
Its in the path of every write

packages/nocodb/src/lib/models/Audit.ts
```diff
@@ -31,37 +31,49 @@ export default class Audit implements AuditType {
...
   public static async insert(
     audit: Partial<
       Audit & {
         created_at?;
         updated_at?;
       }
     >,
-    ncMeta = Noco.ncMeta
+    ncMeta = Noco.ncMeta,
+    { forceAwait }: { forceAwait: boolean } = {
```
pranavxc, line 43: could be just  `forceAwait = process.env['TEST'] === 'true'`

packages/nocodb/src/lib/models/Audit.ts
```diff
@@ -31,37 +31,49 @@ export default class Audit implements AuditType {
...
+      forceAwait: process.env['TEST'] === 'true',
+    }
   ) {
     if (!audit.project_id && audit.fk_model_id) {
       audit.project_id = (
         await Model.getByIdOrName({ id: audit.fk_model_id }, ncMeta)
       ).project_id;
     }
```
pranavxc, line 51: Move this to `insertAudit` function

pranavxc (changes requested):

packages/nocodb/src/lib/models/Audit.ts
mustafapc19: @pranavxc I have kept it for future use. Since its an optional param I thought it won't hurt to keep it here.

packages/nocodb/src/lib/models/Audit.ts
```diff
@@ -31,37 +31,49 @@ export default class Audit implements AuditType {
...
+        details: audit.details,
+        created_at: audit.created_at,
+        updated_at: audit.updated_at,
+      });
+    };
 
-    return auditRec;
+    if (forceAwait) {
+      const auditRec = await insertAudit();
+      return auditRec;
```
pranavxc, line 73: this could be just `return insertAudit()`

pranavxc (changes requested):

pranavxc (approved):

Issue #3781, opened 2022-09-23
Restrict access to app store


o1lab:
- [x] To only super admin

App store : 
- can be brought to root level under user menu.
- and should be only visible to super admin within the project.

PR #3796, opened 2022-09-24
Feat: Limit app store access to super admin

pranavxc:
## Change Summary

- Allow app store access only for super admin
- Add a global app store page
- Only show app store for super admin in project settings

re #3781

## Change type

## Test/ Verification

Provide summary of changes.


packages/nc-gui/composables/useUIPermission/index.ts
```diff
@@ -11,6 +11,14 @@ const hasPermission = (role: Role | ProjectRole, hasRole: boolean, permission: P
 
   if (isString(rolePermission) && rolePermission === '*') return true
 
+  // todo: type correction
+  if ('include' in rolePermission && rolePermission.include) {
+    return rolePermission.include[permission]
+  }
+  if ('exclude' in rolePermission && rolePermission.exclude) {
+    return !rolePermission.exclude[permission]
+  }
```
bcakmakoglu, line 20: I'm not sure if this logic really works - 

Previously, if we checked a permission and the user had the role `Creator`, it would give access to any permission as it had the `*` wildcard permission assigned.
Now it has *no* permissions assigned, except for permissions that are *excluded*.

So shouldn't a creator now have no permissions at all?
Same for the `Owner` role.

bcakmakoglu (changes requested):

packages/nc-gui/composables/useUIPermission/index.ts
pranavxc: The `excluded` means everything excluding those specified permissions, it's like a wildcard with some limitation.

packages/nc-gui/composables/useUIPermission/rolePermissions.ts
```diff
@@ -1,51 +1,65 @@
 import { ProjectRole, Role } from '~/lib'
 
-const rolePermissions = {
+const rolePermissions: Record<Role | ProjectRole, '*' | Partial<Record<'include' | 'exclude', Record<string, boolean>>>> = {
```
bcakmakoglu, line 3: Please remove this typing.
`rolePermissions` is typed as a constant (hence the `as const` cast at the end)

That means the object itself is the interface of the object and adding an additional type as you did just ruins the intention of `as const`

packages/nc-gui/composables/useUIPermission/index.ts
bcakmakoglu: I see.

Could you add a comment explaining the logic behind that so it's obvious to idiots like me? 😄

packages/nc-gui/composables/useUIPermission/index.ts
```diff
@@ -11,6 +11,14 @@ const hasPermission = (role: Role | ProjectRole, hasRole: boolean, permission: P
...
+
```
bcakmakoglu, line 21: Change this to:

```ts
const hasPermission = (role: Role | ProjectRole, hasRole: boolean, permission: Permission | string) => {
  const rolePermission = rolePermissions[role]

  if (!hasRole || !rolePermission) return false

  if (isString(rolePermission) && rolePermission === '*') return true

  if ('include' in rolePermission && rolePermission.include) {
    return !!rolePermission.include[permission as keyof typeof rolePermission.include]
  }

  if ('exclude' in rolePermission && rolePermission.exclude) {
    return !rolePermission.exclude[permission as keyof typeof rolePermission.exclude]
  }

  return rolePermission[permission as keyof typeof rolePermission]
}
```

packages/nc-gui/composables/useUIPermission/rolePermissions.ts
```diff
@@ -1,51 +1,65 @@
...
-    smartSheet: true,
-    column: true,
-    projectSettings: true,
+    include: {
+      smartSheet: true,
+      column: true,
+      projectSettings: true,
+    },
   },
 } as const
```
bcakmakoglu, line 65: Update the below types to so the typing works properly again, otherwise `Permission` will be resolved to `never` with your changes.

```ts
type RolePermissions = Omit<typeof rolePermissions, 'guest' | 'admin' | 'super'>

type GetKeys<T> = T extends Record<any, Record<infer Key, boolean>> ? Key : never

export type Permission<K extends keyof RolePermissions = keyof RolePermissions> = RolePermissions[K] extends Record<any, any>
  ? GetKeys<RolePermissions[K]>
  : never

export default rolePermissions
```

bcakmakoglu (changes requested):

package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "build:common": "cd ./packages/nocodb-sdk; npm install; npm run build",
     "install:common": "cd ./packages/nocodb; npm install ../nocodb-sdk; cd ../nc-gui; npm install ../nocodb-sdk",
-    "start:api": "npm run build:common ; cd ./packages/nocodb; npm install ../nocodb-sdk; npm install; NC_DISABLE_CACHE=true NC_DISABLE_TELE=true npm run watch:run:cypress",
+    "start:api": "npm run build :common ; cd ./packages/nocodb; npm install ../nocodb-sdk; npm install; NC_DISABLE_CACHE=true NC_DISABLE_TELE=true npm run watch:run:cypress",
```
wingkwong, line 18: build:common?

scripts/cypress/integration/common/5a_user_role.js
```diff
@@ -177,48 +177,64 @@ export const genTest = (apiType, dbType) => {
...
         //      Editor/Viewer/Commenter : can only view 'existing' views
         //      Rest: can create/edit
-        _viewMenu(roleType, "userRole");
+        if (roleType !== "owner") {
+          _viewMenu(roleType, "userRole");
+        }
       });
 
       it(`[${roles[roleType].name}] Download files`, () => {
         // to be fixed
```
wingkwong, line 232: what is to be fixed? or already handled? better elaborate more with a TODO tag.

pranavxc: @dstala - Can you add more details with comment?

bcakmakoglu (approved):

scripts/cypress/integration/common/5a_user_role.js
dstala: skipped tests for role `owner` as we do all sort of tests with that credentials already. will update details in comment

PR #3926, opened 2022-10-04
Fix: add singletext number-based sort support

piromsurang:
## Change Summary

To support number-based sort on varchar, normally, it is achieved by `ORDER BY LENGTH(column_name), column_name`

## Change type

## Test/ Verification

Provide summary of changes.

Screen Shot 2565-10-04 at 10 06 39 AM

### DIFF
--- a/packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/sortV2.ts
+++ b/packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/sortV2.ts
@@ -232,8 +232,19 @@ export default async function sortV2(
           break;
         }
       default:
-        qb.orderBy(sanitize(column.column_name), sort.direction || 'asc');
+      {
+        const clientType = knex.clientType();
+        if (clientType === 'mssql') {
+          qb
+            .orderBy(sanitize(knex.raw('LEN(??)', [column.column_name])))
+            .orderBy(sanitize(column.column_name), sort.direction || 'asc');
+        } else {
+          qb
+            .orderBy(sanitize(knex.raw('LENGTH(??)', [column.column_name])))
+            .orderBy(sanitize(column.column_name), sort.direction || 'asc');
+        }
         break;
+      }
     }
   }
 }


packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/sortV2.ts
```diff
@@ -232,8 +232,19 @@ export default async function sortV2(
...
-        qb.orderBy(sanitize(column.column_name), sort.direction || 'asc');
+      {
+        const clientType = knex.clientType();
+        if (clientType === 'mssql') {
+          qb
+            .orderBy(sanitize(knex.raw('LEN(??)', [column.column_name])))
+            .orderBy(sanitize(column.column_name), sort.direction || 'asc');
+        } else {
+          qb
+            .orderBy(sanitize(knex.raw('LENGTH(??)', [column.column_name])))
```
mateonunez, line 243: Using this kind of sort we can generate inconsistencies with data.

image

The issue touches all `varchar` datatypes. (the sort is **asc**, but considering the length we got this)

image

mateonunez (commented):
Check that most common databases has their own type to handle **IP Addresses**.

- [PG](https://www.postgresql.org/docs/current/datatype-net-types.html)
- [MySQL](https://stackoverflow.com/a/4590224/11578351)
- Mssql and SQLite3 has a different behavior with IP Addresses.

Have you consider to create a _custom_ type to handle this case? So we can move the `sort` logic there.

piromsurang:
> Check that most common databases has their own type to handle **IP Addresses**.
> 
> * [PG](https://www.postgresql.org/docs/current/datatype-net-types.html)
> * [MySQL](https://stackoverflow.com/a/4590224/11578351)
> * Mssql and SQLite3 has a different behavior with IP Addresses.
> 
> Have you consider to create a _custom_ type to handle this case? So we can move the `sort` logic there.

could you please help give an example? I am very beginner here. when you mentioned custom type, do you mean custom custom datatype?

o1lab:
Agree with Mateo on using of length is not a good idea here. 

If underlying database supports ipaddress type - a snapshot how the database handles the sort will be helpful to understand and then we can take a call on this. 

However, at the moment I feel the issue raised is an overkill for the usecase.

piromsurang:
@o1lab @mateonunez what should be the next step? should I close this PR?

o1lab:
I will close this for now and mark the issue as won't fix. Appreciate your effort @piromsurang and thoughts from @mateonunez.

Please see if any other [issues](https://github.com/nocodb/nocodb/issues?page=2&q=is%3Aissue+is%3Aopen+label%3A%22%F0%9F%91%8B+For+%3A+Community+or+Good+First+Issue%22) appeal to you & I will be happy to follow up.

PR #3937, opened 2022-10-04
fix: at import failing if more than 64 select options provided

mertmit:
## Change Summary

MySQL: we use enum for single select and set for multi select but [set only allow 64 distinct members](https://dev.mysql.com/doc/refman/8.0/en/set.html). So we will fallback using text for multi-select when importing a multi-select column with options more than 64.
- MySQL use find_in_set only if dt is set (this logic was missing from column apis)
- AT import:
  - match select options case-insensitive (as this leads duplicate entries on db level)
  - use text as dt if 64 or more options provided for multiselect on MySQL
- use text as dt if 64 or more options provided for multiselect on MySQL (Extended this fix to column api level)

## Change type

### DIFF
--- a/packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
+++ b/packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
@@ -577,11 +577,15 @@ export default async (
             ncCol.colOptions = {
               options: [...colOptions.data],
             };
-            // if options are empty, configure '' as default option
-            ncCol.dtxp =
-              colOptions.data
-                .map((el) => `'${el.title.replace(/'/gi, "''")}'`)
-                .join(',') || "''";
+
+            if (['mysql', 'mysql2'].includes(getRootDbType())) {
+              // if options are empty, configure '' as an option
+              ncCol.dtxp =
+                colOptions.data
+                  .map((el) => `'${el.title.replace(/'/gi, "''")}'`)
+                  .join(',') || "''";
+            }
+            
             break;
           case undefined:
             break;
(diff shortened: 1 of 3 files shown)


wingkwong:
Let's sync. I'm revising similar stuff for csv import :)

packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
```diff
@@ -577,9 +577,16 @@ export default async (
             ncCol.colOptions = {
               options: [...colOptions.data],
             };
-            // if options are empty, configure '' as default option
-            ncCol.dtxp =
-              colOptions.data.map((el) => `'${el.title.replace(/'/gi, "''")}'`).join(',') || "''";
+
+            if (['mysql', 'mysql2'].includes(getRootDbType())) {
+              if (colOptions.data.length > 63) {
```
wingkwong, line 582: enum can have more than 64 options. Need discussion on the upper bound. (taking how we display in UI into consideration as well)

e.g.

enum('51','7','35','19','26','4','18','20','16','61','30','22','37','38','14','50','23','48','34','17','32','5','9','15','45','42','54','53','52','28','29','43','40','0','56','33','55','39','58','1','3','31','62','49','6','2','57','63','60','36','44','65','64','8','25','47','46','59','13','41','10','21','12','11','24','27','66')


cc: @o1lab

packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
```diff
@@ -577,11 +577,19 @@ export default async (
...
               options: [...colOptions.data],
             };
-            // if options are empty, configure '' as default option
-            ncCol.dtxp =
-              colOptions.data
-                .map((el) => `'${el.title.replace(/'/gi, "''")}'`)
-                .join(',') || "''";
+
+            if (['mysql', 'mysql2'].includes(getRootDbType())) {
+              if (colOptions.type === 'multiSelect' && colOptions.data.length > 63) {
```
wingkwong, line 582: more than 64 so it should be > 64

wingkwong (approved):

pranavxc (approved):

Issue #3696, opened 2022-09-17
Save a new row when only all required fields are entered


o1lab:
Please [see here](https://github.com/nocodb/nocodb/discussions/3695#discussioncomment-3671877).

PrasanthKumar77:
can we trigger webhook also when all values are rows are added? currently its being triggered when one column value is added also

dstala:
> can we trigger webhook also when all values are rows are added? currently its being triggered when one column value is added also

@PrasanthKumar77 : You can use `expand row` option to update record. This will trigger webhook only when you click `save` button. On similar lines, you can use `Add new row` button in right top tool bar to insert record. This triggers webhook only on clicking `Save` button. Hope this helps.

PR #3953, opened 2022-10-06
Fix: Grid view - save new row only when all required cell values provided

pranavxc:
## Change Summary

 re #3696 

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/composables/useExpandedFormStore.ts
+++ b/packages/nc-gui/composables/useExpandedFormStore.ts
@@ -147,6 +147,7 @@ const [useProvideExpandedFormStore, useExpandedFormStore] = useInjectionState((m
           ltarState,
           getMeta,
           row: row.value.row,
+          throwError: true,
         })
 
         if (missingRequiredColumns.size) return
--- a/packages/nc-gui/composables/useExpandedFormStore.ts
+++ b/packages/nc-gui/composables/useExpandedFormStore.ts
@@ -19,7 +19,9 @@ import {
   useProvideSmartsheetRowStore,
   useSharedView,
 } from '#imports'
+import { useMetas } from '~/composables/useMetas'
 import type { Row } from '~/lib'
+import { populateInsertObject } from '~/utils'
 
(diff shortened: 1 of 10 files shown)


wingkwong (commented):
Let's say I've column A, B, C, and D where C & B are required cells.

Case 1. Fill in A & B only. C & D are highlighted. Cell values are not saved which is expected.
Case 2. Fill in A, B, C & D. Cell values are saved which is expected.
Case 3. After Case 2, remove values from C & D. Cell values are saved. (Is it expected in this case? or the value becomes empty rather than NULL so they are considered as filled?)

pranavxc:
> Let's say I've column A, B, C, and D where C & B are required cells.
> 
> Case 1. Fill in A & B only. C & D are highlighted. Cell values are not saved which is expected. Case 2. Fill in A, B, C & D. Cell values are saved which is expected. Case 3. After Case 2, remove values from C & D. Cell values are saved. (Is it expected in this case? or the value becomes empty rather than NULL so they are considered as filled?)


We are deciding the required field based on the database table field's `not null` attribute, if it's true need to be a non-null value and it can be an empty string as well.

packages/nc-gui/composables/useExpandedFormStore.ts
```diff
@@ -19,7 +19,9 @@ import {
   useProvideSmartsheetRowStore,
   useSharedView,
 } from '#imports'
+import { useMetas } from '~/composables/useMetas'
 import type { Row } from '~/lib'
+import { populateInsertObject } from '~/utils'
```
bcakmakoglu, line 24: -> `'#imports'` 😉

packages/nc-gui/composables/useViewData.ts
```diff
@@ -19,6 +20,7 @@ import {
   useSmartsheetStoreOrThrow,
   useUIPermission,
 } from '#imports'
+import { useMetas } from '~/composables/useMetas'
```
bcakmakoglu, line 23: -> `'#imports'`

bcakmakoglu (changes requested):

wingkwong (approved):

bcakmakoglu (approved):

PR #3979, opened 2022-10-08
feat: OpenID Connect authentication

### LINKED ISSUE #742: Features: OpenID Connect Auth
OpenID Connect isn't a long step from Google Auth, but would add support for e.g. Azure AD, Github, Gitlab, KeyCloak, ...

kobim:
## Change Summary

Add the option to configure an OpenID Connect identity provider as a sign-in option.

Resolves #742

## Change type

### Without any modifications

- [x] Sign in with email & password
- [x] Sign up with email & password

### Defining an OIDC provider

- [x] Login screen without a display name

Login without a display name

- [x] Signup screen without a display name

- [x] Login screen with a display name

Login with a display name

- [x] Signup screen with a display name
- [x] The first signup assigns the *Owner* role to the new user.
- [x] Invited users with the same email defined by the IdP can accept invitations.

### Forcing the IdP

When the provider is set, it overrides the option to use another identity method (such as email & password or Google Login).
Is this the expected behavior? I would say organizations want to restrict access to their installations, which is the appropriate way to ensure only the IdP will be able to dictate who should be able to login.

### DIFF
--- a/packages/nc-gui/composables/useGlobal/state.ts
+++ b/packages/nc-gui/composables/useGlobal/state.ts
@@ -82,7 +82,7 @@ export function useGlobalState(storageKey = 'nocodb-gui-v2'): State {
     firstUser: true,
     githubAuthEnabled: false,
     googleAuthEnabled: true,
-    oidcAuthName: '',
+    oidcAuthName: null,
     ncMin: false,
     oneClick: false,
     projectHasAdmin: false,
--- a/packages/nc-gui/composables/useGlobal/state.ts
+++ b/packages/nc-gui/composables/useGlobal/state.ts
@@ -82,6 +82,7 @@ export function useGlobalState(storageKey = 'nocodb-gui-v2'): State {
     firstUser: true,
     githubAuthEnabled: false,
     googleAuthEnabled: true,
+    oidcAuthName: '',
     ncMin: false,
     oneClick: false,
     projectHasAdmin: false,
--- a/packages/nc-gui/pages/signup/[[token]].vue
+++ b/packages/nc-gui/pages/signup/[[token]].vue
@@ -97,7 +97,7 @@ function resetError() {
           {{ $t('msg.info.signUp.superAdmin') }}
         </h2>
 
-        <a-form ref="formValidator" :model="form" layout="vertical" no-style @finish="signUp">
+        <a-form v-if="appInfo.oidcAuthName == ''" ref="formValidator" :model="form" layout="vertical" no-style @finish="signUp">
           <Transition name="layout">
             <div v-if="error" class="self-center mb-4 bg-red-500 text-white rounded-lg w-3/4 mx-auto p-1">
               <div class="flex items-center gap-2 justify-center">
@@ -138,7 +138,7 @@ function resetError() {
               <span class="flex items-center gap-2">
                 <LogosGoogleGmail />
 
-                {{ $t('labels.signUpWithGoogle') }}
+                {{ $t('labels.signUpWithProvider', { provider: 'Google' }) }}
               </span>
             </a>
 
@@ -158,6 +158,23 @@ function resetError() {
             </div>
           </div>
         </a-form>
+
+        <div
+          v-if="appInfo.oidcAuthName != ''"
+          class="self-center flex flex-col flex-wrap gap-4 items-center mt-4 justify-center">
+            <a
+              :href="`${appInfo.ncSiteUrl}/auth/oidc`"
+              class=" !text-primary !no-underline"
+            >
+            <button type="button" class="scaling-btn bg-opacity-100">
+              <span class="flex items-center gap-2">
+                <MdiLogin />
+
+                {{ $t('labels.signUpWithProvider', { provider: appInfo.oidcAuthName }) }}
+              </span>
+            </button>
+          </a>
+        </div>
       </div>
 
       <div class="prose-sm mt-4 text-gray-500">
--- a/packages/nc-gui/pages/signup/[[token]].vue
+++ b/packages/nc-gui/pages/signup/[[token]].vue
@@ -97,7 +97,7 @@ function resetError() {
           {{ $t('msg.info.signUp.superAdmin') }}
         </h2>
 
-        <a-form v-if="appInfo.oidcAuthName == ''" ref="formValidator" :model="form" layout="vertical" no-style @finish="signUp">
+        <a-form v-if="!appInfo.oidcAuthName" ref="formValidator" :model="form" layout="vertical" no-style @finish="signUp">
           <Transition name="layout">
             <div v-if="error" class="self-center mb-4 bg-red-500 text-white rounded-lg w-3/4 mx-auto p-1">
               <div class="flex items-center gap-2 justify-center">
@@ -160,7 +160,7 @@ function resetError() {
         </a-form>
 
         <div
-          v-if="appInfo.oidcAuthName != ''"
+          v-if="appInfo.oidcAuthName"
           class="self-center flex flex-col flex-wrap gap-4 items-center mt-4 justify-center">
             <a
               :href="`${appInfo.ncSiteUrl}/auth/oidc`"
--- a/packages/nocodb/src/lib/meta/api/utilApis.ts
+++ b/packages/nocodb/src/lib/meta/api/utilApis.ts
@@ -35,7 +35,7 @@ export async function appInfo(req: Request, res: Response) {
     process.env.NC_OIDC_CLIENT_ID &&
     process.env.NC_OIDC_CLIENT_SECRET) ?
       ( process.env.NC_OIDC_DISPLAY_NAME ?? 'Your Identity Provider' ) :
-      '';
+      null;
   const result = {
     authType: 'jwt',
     projectHasAdmin,
--- a/packages/nocodb/src/lib/meta/api/utilApis.ts
+++ b/packages/nocodb/src/lib/meta/api/utilApis.ts
@@ -28,6 +28,14 @@ export async function testConnection(req: Request, res: Response) {
 
 export async function appInfo(req: Request, res: Response) {
   const projectHasAdmin = !(await User.isFirst());
+  const oidcAuthName = !!(process.env.NC_OIDC_ISSUER &&
+    process.env.NC_OIDC_AUTH_URL &&
+    process.env.NC_OIDC_TOKEN_URL &&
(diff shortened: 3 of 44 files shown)


bcakmakoglu:
Thanks for your contribution :)
We'll take a look but so far this looks nice 👍

packages/nc-gui/composables/useGlobal/state.ts
```diff
@@ -82,6 +82,7 @@ export function useGlobalState(storageKey = 'nocodb-gui-v2'): State {
     firstUser: true,
     githubAuthEnabled: false,
     googleAuthEnabled: true,
+    oidcAuthName: '',
```
bcakmakoglu, line 85: Use `null` instead of an empty string as initial value, that way we don't need to check against empty strings in the template later on.

packages/nc-gui/pages/signup/[[token]].vue
```diff
@@ -97,7 +97,7 @@ function resetError() {
           {{ $t('msg.info.signUp.superAdmin') }}
         </h2>
 
-        <a-form ref="formValidator" :model="form" layout="vertical" no-style @finish="signUp">
+        <a-form v-if="appInfo.oidcAuthName == ''" ref="formValidator" :model="form" layout="vertical" no-style @finish="signUp">
```
Comment (bcakmakoglu) [inline packages/nc-gui/pages/signup/[[token]].vue:100]: Relating to the comment above, if you use `null` as the default you can replace this with `v-if="!appInfo.oidcAuthName"`

packages/nc-gui/pages/signup/[[token]].vue
```diff
@@ -158,6 +158,23 @@ function resetError() {
             </div>
           </div>
         </a-form>
+
+        <div
+          v-if="appInfo.oidcAuthName != ''"
```
Comment (bcakmakoglu) [inline packages/nc-gui/pages/signup/[[token]].vue:163]: This could be changed to just be `v-if="appInfo.oidcAuthName"`

packages/nocodb/src/lib/meta/api/utilApis.ts
```diff
@@ -23,6 +23,14 @@ export async function testConnection(req: Request, res: Response) {
...
 export async function appInfo(req: Request, res: Response) {
   const projectHasAdmin = !(await User.isFirst());
+  const oidcAuthName = !!(process.env.NC_OIDC_ISSUER &&
+    process.env.NC_OIDC_AUTH_URL &&
+    process.env.NC_OIDC_TOKEN_URL &&
+    process.env.NC_OIDC_USERINFO_URL &&
+    process.env.NC_OIDC_CLIENT_ID &&
+    process.env.NC_OIDC_CLIENT_SECRET) ?
+      ( process.env.NC_OIDC_DISPLAY_NAME ?? 'Your Identity Provider' ) :
+      '';
```
bcakmakoglu, line 33: Return `null` as auth name instead of empty string so we can enforce the checks as mentioned above.

bcakmakoglu (changes requested):

syji35:
Thanks for the PR. Is it possible to merge for the next release ?  
We are waiting for this 👍

syji35:
@o1lab : What is missing to move the pull request forward ?

o1lab:
First up, thank you @kobim for this PR. Most products offer SSO & SAML as paid features. We are offering google-auth SSO for free already. Hence we are discussing whether this should be in our free version or not. @kobim @syji35 - I understand SAML is not part of openid auth. Could you please confirm.

syji35:
> First up, thank you @kobim for this PR. Most products offer SSO & SAML as paid features. We are offering google-auth SSO for free already. Hence we are discussing whether this should be in our free version or not. @kobim @syji35 - I understand SAML is not part of openid auth. Could you please confirm.

I Confirm

timmwille:
> SAML and OpenID Connect are two completely different approaches to single sign on. OIDC would not only allow developers and home users to integrate with existing user stores but also, if implemented correctly, would allow developers to build plugins and integrations that leverage OIDC to deliver a smooth experience for users as, for example, a Nextcloud plugin.
> 
> Generally speaking, SSO is not a business only feature, it's a core security feature just as much as users and RBAC are.

I agree much, if there is any chance to include this in the near future I would appreciate it very much and recommend within our non-profit association it requires OIDC. If there is anything we can do to help, I would reach out. 
This is kind of the current discussion between alternatives and I would love to promote NocoDB instead.

ja0nz:
Came here to boost this PR as well❤️ I believe OIDC offers some compelling benefits regarding service integration and potential extensibility and should be treated on par with password auth. Advanced SAML and RBAC features could instead fit well into the paid version.

almereyda:
The last comment expresses very well how SAML and OIDC are very often approached in other projects of the open source world. This is to differentiate between Free Software, which intents to liberate all aspects of an application, and open-core business models, where the open core is a working, tested and standards compliant implementation, and enterprise features, like SSO, are sold atop. Please consider supporting OIDC for secure identity federation also for small-scale users.

As a small not-for-profit organisation, we also have to second @timmwille's comment above.

AlexM4H:
Who takes care of the Playwright test errors?

moedn:
Good afternoon nocodb folks (@o1lab) :) It would be really nice to have a decision here: Does nocodb want this PR or not? Currently, I'm seeing a lot of people waiting, not only for the PR, but for a general statement how open source nocodb wants to be (and why it wanted to be open source in the first place). This piece of software would become a critical part of our administrative infrastructure, so it would be important to know whether we can improve it, fix issues etc. as a communal effort …or will be essentially paying for a service.

A general comment regarding business models: Since there seem to be a few non-profit organisations here using and contributing to nocodb, we can also talk about public funding, financially supporting the development efforts (as a project hosted by these non-profits). This wouldn't be possible (or at least far more difficult) for proprietary technologies – especially when you'd want funding from a country you're not necessarily located in.

Hope that helps :)

anselmoshim:
Is there a way I can disable the NocoDB login page entirely so I can just set up my own login page with whichever SSO/SAML provider I want?

flaktrooper:
I have to admit it is a little concerning to see this being considered as "value add".
SSO is first and foremost a security feature, and I would argue a core one at that. 

Managing separate accounts and permission levels for dozens of volunteers on every single application we deploy is not feasible, and carries a significantly increased risk of mistakes. Paywalling SSO simply means we will have to use another product entirely, because we are not in a position to pay monthly subscription fees for every application we use. This is a feature that is _required_ by more than just enterprise users - I think @ja0nz has suggested an approach that is much more reasonable.

timmwille:
> I have to admit it is a little concerning to see this being considered as "value add". SSO is first and foremost a security feature, and I would argue a core one at that.
> 
> Managing separate accounts and permission levels for dozens of volunteers on every single application we deploy is not feasible, and carries a significantly increased risk of mistakes. Paywalling SSO simply means we will have to use another product entirely, because we are not in a position to pay monthly subscription fees for every application we use.
> 
> This is a feature that is _required_ by more than just enterprise users - I think @ja0nz has suggested an approach that is much more reasonable.

yes, in most cases it is "I rather pay for good service, then for lock-in or lock-out" - I consider a separate Patch module to be funded already if this won't move but I also do not want provoke endless policy discussions. So maybe it is time to start looking for alternatives. Live and let live. Hopefully it does not end like this (probably will): https://forgejo.org/faq/

o1lab:
We have been rearchitecting a large part of our application which was released as [beta](https://github.com/nocodb/nocodb/releases/tag/0.101.0-beta.0). So apologies for late reply & thanks for your patience. 

We did consider this request a **FAIR BIT** however we have clearly stated that SSO will be [part of our enterprise edition](https://docs.nocodb.com/FAQs#do-you-plan-to-have-enterprise-edition-) since [long time ago](https://github.com/nocodb/nocodb/blob/develop/packages/noco-docs/content/en/FAQs.md#do-you-plan-to-have-enterprise-edition-). We have a [public roadmap](https://github.com/orgs/nocodb/projects/13/views/2) available here now which also gives a clear picture on what features are (CE) and what are in enterprise edition (EE). And all PRs submitted to relevant edition will be accepted in that edition (similar to Gitlab).

I can understand it is tempting to request most of the features in CE but please do understand this is a fairly large project which involves lot of effort and it is important for us to keep it sustainable. And please do also note our community edition (CE) is much **MORE GIVING** than anything out there in this space (feature list below). 

More than basic features of all no-code database features.
- Email authentication.
- Google auth. 
- All of the roles (super-admin, creator, editor, viewer, commenter).
- Kanban, Survey view / Wizard view
- Locked views
- Row comments. 
- UI access control.
- CSV/JSON download and upload.
- Audit logs. 
- Support for existing databases.
- Support any underlying SQL database to run on.

If @kobim is in agreement  - happy to accept the PR into EE.

ja0nz:
@o1lab Thank you for the thorough clarification. Not the answer I was hoping for but I still love NocoDB for what it is🙏

timmwille:
Thank you for clarifying @o1lab → I guess your legal department will have to do a lot of work if you implement AGPL PR from @kobim but for us this is too much taking from the community but not giving back. At some point in the future this might fall on your feet. I still hope the project will survive. For us it is time to move on and look for alternatives...

timmwille:
@kobim what about a nice community fork? :robot:

syji35:
> Thank you for clarifying @o1lab → I guess your legal department will have to do a lot of work if you implement AGPL PR from @kobim but for us this is too much taking from the community but not giving back. At some point in the future this might fall on your feet. I still hope the project will survive. For us it is time to move on and look for alternatives...

We are migrating to Grist, which offers SAML authentication (works with Keycloak for example)

IcedQuinn:
It's always sad to see ex-open source software start gatekeeping patches out for corpo editions.

jogueber:
Well free and open source is a wish that you only get if you have it in a foundation setup I guess. 
In apache projects this is more or less standard so quiet sad to see that even PRs from the community are not accepted. 
Usually how these projects then go is that they go to die. 

Because in the end what you see is: people pay for services and are willing to if they feel like they can also control the product. 
Here that seems to be not really the spirit

samuk:
That's a shame. What's the best alternative at this point? Is it worth forking NocoDB? Or a people going to https://github.com/gristlabs/grist-core instead?

baltazartroisville:
@o1lab i wish nocodb would cancel google auth, audit log, support for all databases and instead put oidc in for the community edition. that is no enterprise feature at all, even for my family and friends deployment its useful and makes life a lot easier. you could offer enterprise customers still a lot other features like rbac, saml, ldap, customer support, etc.

PR #4039, opened 2022-10-13
fix(ng-gui): find unique neighbours

### LINKED ISSUE #4024: [Cy]: Circular reference in Formula column

### Steps to reproduce ?

- create new table
- create new formula field 'One': {Title}
- create new formula field 'Two': {One}
- create new formula field 'Three': CONCAT({Two},{Two})
- edit definition of either column One or Two
- add a single space at the end of the formula
- error "Can’t save field because it causes a circular reference" displayed

### Desired Behavior

This should not be flagged as a circular reference and it should be possible to edit the formula fields. Note: The formulas work fine (the expected calculated values are displayed) but further editing of the formulas is impossible.

### Project Details

Node: **v18.10.0**
Arch: **x64**
Platform: **darwin**
Docker: **false**
Database: **sqlite3**
ProjectOnRootDB: **true**
RootDB: **sqlite3**
PackageVersion: **0.97.0**

### Attachements

_No response_

jbbn:
## Change Summary

Find unique neighbours on formula paths, therefore avoiding "circular
reference" incorrect error validation while editing formula columns.

## Change type

- [X] fix: (bug fix for the user, not a fix to a build script)

## Test/ Verification

Use `new Set` to create a array of graph dependencies without dupications.

### Before

https://user-images.githubusercontent.com/3846247/195473594-0dbc4807-1c05-40d0-8e93-fb2b6fbc7191.mov

### After

https://user-images.githubusercontent.com/3846247/195473738-810be0db-0183-4c00-bed8-acea09daf498.mov

(IMO, I think this part of the code could use some unit testing - adding the duplicated graph depedencies scenario would be straightforward and it would be easier to "see" the coverage of use cases)

ref: #4024

### DIFF
--- a/packages/nc-gui/components/smartsheet/column/FormulaOptions.vue
+++ b/packages/nc-gui/components/smartsheet/column/FormulaOptions.vue
@@ -252,11 +252,15 @@ function validateAgainstMeta(parsedTree: any, errors = new Set(), typeErrors = n
     const formulaPaths = columns.value
       .filter((c: Record<string, any>) => c.id !== column.value?.id && c.uidt === UITypes.Formula)
       .reduce((res: Record<string, any>[], c: Record<string, any>) => {
-        // in `formula`, get all the target neighbours
+        // in `formula`, get all the (unique) target neighbours
         // i.e. all column id (e.g. cl_xxxxxxxxxxxxxx) with formula type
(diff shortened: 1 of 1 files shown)


wingkwong (approved):
As spoken in the issue, please check with the assignee first next time.

PR #4101, opened 2022-10-15
feat: better scroll for borders and sticky header

mertmit:
## Change Summary

- Avoid overlapping with sticky header
- If first column include # column in view as well
- If last row include 'Add New Row' in view as well

## Change type

## Test/ Verification

https://user-images.githubusercontent.com/59797957/195978223-74d7104d-4390-49c3-ab56-db78e7414083.mp4

### DIFF
--- a/packages/nc-gui/components/smartsheet/Grid.vue
+++ b/packages/nc-gui/components/smartsheet/Grid.vue
@@ -88,6 +88,8 @@ const expandedFormDlg = ref(false)
 const expandedFormRow = ref<Row>()
 const expandedFormRowState = ref<Record<string, any>>()
 const tbodyEl = ref<HTMLElement>()
+const gridWrapper = ref<HTMLElement>()
+const tableHead = ref<HTMLElement>()
 
 const {
   isLoading,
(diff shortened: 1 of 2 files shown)


packages/nc-gui/components/smartsheet/Grid.vue
```diff
@@ -114,16 +116,57 @@ const { selectCell, selectBlock, selectedRange, clearRangeRows, startSelectRange
...
+            ? gridWrapper.value.scrollTop + relativePos.bottom + 9 // 9 is for border
+            : relativePos.top - headerHeight < 0
+            ? gridWrapper.value.scrollTop + relativePos.top - headerHeight
+            : gridWrapper.value.scrollTop,
+        left:
+          relativePos.right > 0
+            ? gridWrapper.value.scrollLeft + relativePos.right + 9 // 9 is for border
+            : relativePos.left < 0
+            ? gridWrapper.value.scrollLeft + relativePos.left
+            : gridWrapper.value.scrollLeft,
```
bcakmakoglu, line 168: It would be nice if these sort of blocks had some explaining comments so other users/contributors understand what it exactly does.
Just by looking at it, it would be hard to discern what these blocks actually calculate 😄

bcakmakoglu: Might be even better so outsource the logic into a function instead of inlining these conditions :)

mertmit: Implemented `getContainerScrollForElement` function, tried to comment as well (mostly geometrical calculations so don't know how to better comment for them)

bcakmakoglu (approved):

pranavxc (approved):

PR #4137, opened 2022-10-19
fix(nc-gui): gallery cover image is clipped

mayacoda:
## Change Summary

ref: #3599

## Change type

## Test/ Verification

No test changes

This PR changes the styling of cover images from `object-cover`, which scales down the image to the smallest size so it fills the entire container, to `object-contain` which scales the image so it fits entirely within the container. 

**Old:** `object-cover` looks better in terms of design, since each card is uniformly filled with the image. However, information is lost in this case.
image

**New:** `object-contain` displays the full image, albeit much smaller in the case of wider or longer images. 
image

Since the cards are responsive and their width varies, it's unadvisable to fix their size to the size of the image.

Ideally, the user would be able to choose how the image is displayed, similar to this UI from [Craft](https://www.craft.do/). I can see both use cases being valid for users.
image

### DIFF
--- a/packages/nc-gui/components/smartsheet/Gallery.vue
+++ b/packages/nc-gui/components/smartsheet/Gallery.vue
@@ -200,7 +200,7 @@ watch(view, async (nextView) => {
                   :key="`carousel-${record.row.id}-${index}`"
                   quality="90"
                   placeholder
-                  class="h-52 object-cover"
+                  class="h-52 object-contain"
                   :src="attachment.url"
                 />
               </a-carousel>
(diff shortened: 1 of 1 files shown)


wingkwong (approved):
🐕

chrisLusty:
I've hit this exact issue in my testing of nocodb gallery view, i.e. thumbnail images being clipped/zoomed inappropriately. This fix looks like the perfect solution from the images shown above. Has it actually been implemented into a nocodb release yet, or is it scheduled to be?

wingkwong:
@chrisLusty Please create a new issue with the reference to this PR. This fix was applied to an old version and probably got changed in latest version.

Issue #4213, opened 2022-10-27
🐛 Bug: Error `no such file or directory` when Enabling Backblaze B2


3699n:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

When trying to enable Backblaze B2 through the appstore on a fresh NocoDB instance without any projects, testing the configuration fails even when entering the correct credentials and region: `no such file or directory, open '/app/code/temp.txt`

Logs:
```
Oct 27 12:56:24 errno: -2,
Oct 27 12:56:24 code: 'ENOENT',
Oct 27 12:56:24 syscall: 'open',
Oct 27 12:56:24 path: '/app/code/temp.txt'
Oct 27 12:56:24 }
Oct 27 12:56:24 Error [Error: ENOENT: no such file or directory, open '/app/code/temp.txt'] {
Oct 27 12:56:24 errno: -2,
Oct 27 12:56:24 code: 'ENOENT',
Oct 27 12:56:24 syscall: 'open',
Oct 27 12:56:24 path: '/app/code/temp.txt'
Oct 27 12:56:24 }
Oct 27 12:56:24 p :: [Error: ENOENT: no such file or directory, open '/app/code/temp.txt'] {
Oct 27 12:56:24 errno: -2,
Oct 27 12:56:24 code: 'ENOENT',
Oct 27 12:56:24 syscall: 'open',
Oct 27 12:56:24 path: '/app/code/temp.txt'
Oct 27 12:56:24 }
Oct 27 12:56:25 POST /api/v1/db/meta/plugins/test 400 70 - 50.493 ms
Oct 27 12:56:30 GET /dashboard 301 185 - 2.259 ms
Oct 27 12:56:32 File Error [Error: ENOENT: no such file or directory, open '/app/code/temp.txt'] {
Oct 27 12:56:32 errno: -2,
Oct 27 12:56:32 code: 'ENOENT',
Oct 27 12:56:32 syscall: 'open',
Oct 27 12:56:32 path: '/app/code/temp.txt'
Oct 27 12:56:32 }
Oct 27 12:56:32 Error [Error: ENOENT: no such file or directory, open '/app/code/temp.txt'] {
Oct 27 12:56:32 errno: -2,
Oct 27 12:56:32 code: 'ENOENT',
Oct 27 12:56:32 syscall: 'open',
Oct 27 12:56:32 path: '/app/code/temp.txt'
Oct 27 12:56:32 }
[...]

PR #4219, opened 2022-10-28
fix: app store plugin test

wingkwong:
## Change Summary

ref: #4213

Error `no such file or directory` shows sometimes when testing the plugin. The reason is the file `temp.txt` isn't ready when we need it. We should have waited for the stream emitting `finish` before the futher process.

## Change type

## Test/ Verification

image

### DIFF
--- a/packages/nocodb/src/lib/utils/pluginUtils.ts
+++ b/packages/nocodb/src/lib/utils/pluginUtils.ts
@@ -3,10 +3,14 @@ import fs from 'fs';
 export async function waitForStreamClose(
   stream: fs.WriteStream
 ): Promise<void> {
-  stream.end();
-  return new Promise((resolve) => {
-    stream.once('finish', () => {
-      resolve();
-    });
+  return new Promise((resolve, reject) => {
+    stream
+      .once('finish', () => {
+        resolve();
+      })
+      .once('error', () => {
+        reject();
+      });
+    stream.end();
   });
 }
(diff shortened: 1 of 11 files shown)


packages/nocodb/src/lib/utils/pluginUtils.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export async function waitForStreamClose(
+  stream: fs.WriteStream
+): Promise<void> {
+  stream.end();
+  return new Promise((resolve) => {
+    stream.once('finish', () => {
+      resolve();
+    });
+  });
```
pranavxc, line 11: We need a `stream.once('error',` as well to `reject` in case some error occurs

packages/nocodb/src/lib/utils/pluginUtils.ts
```diff
@@ -0,0 +1,12 @@
+import fs from 'fs';
+
+export async function waitForStreamClose(
+  stream: fs.WriteStream
+): Promise<void> {
+  stream.end();
```
pranavxc, line 6: Hope we need to move it after adding the `finish` handler

```js
  return new Promise((resolve) => {
    stream.once('finish', () => {
      resolve();
    });
    stream.end()
  });
```

packages/nocodb/src/lib/utils/pluginUtils.ts
```diff
@@ -3,10 +3,14 @@ import fs from 'fs';
...
-      resolve();
-    });
+  return new Promise((resolve, reject) => {
+    stream
+      .once('finish', () => {
+        resolve();
+      })
+      .once('error', () => {
+        reject();
+      });
```
pranavxc, line 13: when rejecting you can pass the error as well

```js
      .once('error', (err) => {
        reject(err);
      });
```

pranavxc (approved):

Issue #4230, opened 2022-10-28
🐛 Bug: AT import fails if other tables present in project


mertmit:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Create a project
Create a table
Try to import a project
Fetching data blocked at non-imported table

### Desired Behavior

We should only fetch data for imported tables

### Project Details

develop

wingkwong:
Fixed in `0.99.0`

PR #4231, opened 2022-10-28
fix: try to fetch data for only imported tables

mertmit:
## Change Summary

Re #4230

## Change type

### DIFF
--- a/packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
+++ b/packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
@@ -2176,7 +2176,7 @@ export default async (
 
     logBasic('Importing Tables...');
     // prepare table schema (base)
-    await nocoCreateBaseSchema(aTblSchema);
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/lib/meta/api/sync/helpers/job.ts
```diff
@@ -2235,9 +2235,9 @@ export default async (
 
         const recordsMap = {};
 
-        for (let i = 0; i < ncTblList.list.length; i++) {
+        for await (const table of ncTblList.list.filter((ncTbl) => tables.find((table) => table.title === ncTbl.title))) {
```
pranavxc, line 2238: Why `for await...of` here, is it required or just ``for await...of` is enough?

mertmit: Yes right also I missed that we only miss that check for link part, so I revised that commit, thank you for pointing out.

PR #4360, opened 2022-11-10
fix(test): Maximum timeout for each instance is 30 minutes and improved quick test CI configuration

mustafapc19:
-  Made maximum timeout to 30 minutes for a CI instance
- Quick test's backend is now waited on till its ready
- Handling playwright report of quick test as a different report. Quick tests playwright reports were overwriting the actual tests report.
- Added backend log of quick test server as GH artifact

### DIFF
--- a/.github/workflows/playwright-test-workflow.yml
+++ b/.github/workflows/playwright-test-workflow.yml
@@ -14,6 +14,7 @@ on:
 jobs:
   playwright:
     runs-on: ubuntu-20.04
+    timeout-minutes: 20
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/playwright-test-workflow.yml
```diff
@@ -14,6 +14,7 @@ on:
 jobs:
   playwright:
     runs-on: ubuntu-20.04
+    timeout-minutes: 20
```
wingkwong, line 17: PR title says `30` but here's `20`. Please double check the number.

wingkwong (approved):

PR #4379, opened 2022-11-13
feat: add filter types for multiselect field

### LINKED ISSUE #534: 🐛 Bug: Filtering MultiSelect in more than basic way
## Environment

NocoDB used as docker : true
NocoDB version : 0.11.29
Database used in NC_DB URL  : pg
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : pg
OS on which NocoDB is running : Linux
Node.js version if running as node : v12.22.1
Database version : 13.4

## Steps To Reproduce

1. Add a column named `tags` of type "MultiSelect" to a table, add options `Foo`, `Bar` and `Quux`
2. Add rows with the following `tags` values:
    - no values (row ❶)
    - only `Foo` (row ❷)
    - only `Bar` (row ❸)
    - only `Quux` (row ❹)
    - `Foo` and `Bar` (row ❺)
    - `Foo` and `Bar` and `Quux` (row ❻)
3. Try to add a filter so that only rows which include `Foo` appear (rows ❷, ❺ and ❻).

## Expected behavior

There should be an easy way to filter rows based on MultiSelect fields. Using the provided filter operators `is equal` or `is like`, this does not seem possible, or at least it's not obvious how to do that.

## Suggested solution

The possible filter operators should change when I select a MultiSelect field:
- `contains all of`
- `contains any of`
- `does not contain any of` (alternative phrasing: `contains none of`)
- `does not contain all of` (included for completeness, maybe not useful/desirable)

The existing filter operators `is (not) null` and `is (not) equal` should be kept.

However, the following operator families should be hidden for MultiSelect fields. They do not make much sense for such fields and distract from the "correct" operators.
- `is (not) like` - given that LIKE is string-based, how would one specify filter values?
    - option IDs? not even exposed in the UI
    - option labels? my option label could be a `?` or `_`, breaking the usual LIKE syntax
- `less/greater than (or equal)` - comparing by inherent ordering is impractical as the ordering of MultiSelect options seems to be fixed as "creation order"
    - Full disclosure: I _could_ imagine a use case for these operators if the filter criterion is a number: "all rows that have more than one tag, no matter which". But that is probably covered more intuitively by listing a separate filter field option (a "pseudo field") called `number of tags`, which would also allow things like `number of tags` `is equal` `1`.

Example filters:

- where `tags` `contains all of` `Foo`
    - result: rows ❷, ❺ and ❻
    - solution for step 3 above
- where `tags` `contains any of` `Foo, Bar`
    - result: rows ❷, ❸, ❺ and ❻
- where `tags` `contains all of` `Foo, Bar`
    - result: rows ❺ and ❻
- where `tags` `is equal` `Foo, Bar`
    - result: row ❺
- where `tags` `is equal` `Bar`
    - result: row ❸
- where `tags` `is not null`
    - result: rows ❷, ❸, ❹, ❺ and ❻
- where `tags` `is null`
    - result: row ❶
- where `tags` `does not contain any of` `Quux`
    - result: rows ❶, ❷, ❸ and ❺

southball:
## Change Summary

Add filters "contains all of", "contains any of", "does not contain all of" and "does not contain any of" for multiselect field (although they are enabled for all field types).
Resolves #534.

## Change type

- [X] feat: (new feature for the user, not a new feature for build script)

## Test/ Verification

Tested locally with multiselect field.

- I did a quick search for `'eq'` in the entire codebase and added the new operators to where it seems is needed. If there seems to be missing implementations for the new operators, please kindly let me know.


wingkwong (changes requested):
Trying the examples in the issue post. The following cases are failing.

Test Case 2:

where `tags` `contains any of` `Foo, Bar`
Expected Result: rows ❷, ❸, ❺ and ❻
Your Result: rows ❷, ❺ and ❻

Test Case 3:

where `tags` `contains all of` `Foo, Bar`
Expected Result: rows ❺ and ❻
Your Result: empty

Test Case 4:

where `tags` `is equal` `Foo, Bar`
Expected Result: rows ❺
Your Result: empty

Test Case 8:

where `tags` `does not contain any of` `Quux`
Expected Result: rows ❶, ❷, ❸ and ❺
Your Result: rows ❷, ❸ and ❺

wingkwong (commented):
besides, these new operators should be only available when filtering MultiSelect.

southball:
Test case 8 is a problem. I will take a look at it later.

Regarding test case 2, 3, and 4, there is a problem that there is a space after the `,`. A sensible solution would be to trim each part of the query, however it is possible that one of the value of multiselect can contain leading and/or trailing whitespaces. Is it alright to ignore such cases and pick the solution of trimming the split query?

wingkwong:
We may also think about when a tag contains `,`, e.g. `Foo, Bar` and the query is like `Foo, Bar`.

packages/nocodb/src/lib/meta/helpers/webhookHelpers.ts
```diff
@@ -72,6 +72,26 @@ export async function validateCondition(filters: Filter[], data: any) {
...
+        break;
+      case 'anyof':
+        res = (filter.value?.split(',') ?? []).some((item) =>
+          (data[field]?.split(',') ?? []).includes(item)
+        );
+        break;
+      case 'nallof':
+        res = !(filter.value?.split(',') ?? []).every((item) =>
+          (data[field]?.split(',') ?? []).includes(item)
+        );
```
pranavxc, line 88: `notallof` - I guess it should check all of them are not matching, right now it will be true even if one is not matching.

```
  res = (filter.value?.split(',') ?? []).every((item) =>
          !(data[field]?.split(',') ?? []).includes(item)
        );

```

The same applies for `nanyof`

```
res = (filter.value?.split(',') ?? []).some((item) =>
          !(data[field]?.split(',') ?? []).includes(item)
        );
```

Correct me if I'm wrong

southball: Let us take a look at "does not contain any of" (`nanyof`) first. Consider the filter `nanyof Foo,Bar` and two tag values:

1. Foo,Bar,Quux
2. Foo,Quux

Both 1, 2 should *not* be shown, because they contain some of the items in the filter.

Let us then take a look at the implementation.

```ts
      case 'nanyof':
        res = !(filter.value?.split(',').map((item) => item.trim()) ?? []).some((item) =>
          (data[field]?.split(',') ?? []).includes(item)
        );
        break;
```

In this case, `(filter.value?.split(',').map((item) => item.trim()) ?? [])` evaluates to `["Foo", "Bar"]` and so the `some` evaluates to `true` in both case. Which means that `res` is false, and the rows will not be shown.

So I *think* the implementation of `nanyof` is correct, but I am not 100% sure. What do you think?

southball:
It has taken me some time to get to this PR, but I have rebased the PR and added some fixes.

- Between the time that this PR is made and the rebase, it became impossible to type in spaces in the filter. However, trimming of string is still added. It should pass all test cases now.
- The filters are only now shown if the field `uidt` is `MultiSelect`.

packages/nc-gui/components.d.ts
```diff
@@ -90,6 +90,7 @@ declare module '@vue/runtime-core' {
     LogosMysqlIcon: typeof import('~icons/logos/mysql-icon')['default']
     LogosPostgresql: typeof import('~icons/logos/postgresql')['default']
     LogosRedditIcon: typeof import('~icons/logos/reddit-icon')['default']
+    LogosSnowflakeIcon: typeof import('~icons/logos/snowflake-icon')['default']
```
southball, line 93: I think ESLint added this line, which should be alright?

packages/nc-gui/utils/filterUtils.ts
```diff
@@ -35,6 +35,26 @@ export const comparisonOpList = [
     value: 'notnull',
     ignoreVal: true,
   },
+  {
+    text: 'contains all of',
+    value: 'allof',
+    types: ['MultiSelect'],
```
southball, line 41: Added a field here, so we can have more uidt-specific filters in the future.

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/conditionV2.ts
```diff
@@ -306,6 +306,46 @@ const parseConditionV2 = async (
...
+                }
+              }
+            };
+            if (
+              filter.comparison_op === 'allof' ||
+              filter.comparison_op === 'anyof'
+            ) {
+              qb = qb.where(condition);
+            } else {
+              qb = qb.whereNot(condition).orWhereNull(field);
```
southball, line 346: Added `orWhereNull(field)` for the last test. Things behave unexpectedly for our usecase when `null` is involved.

packages/nc-gui/components.d.ts
wingkwong: Yes. This was introduced from another PR. Maybe the author didn't commit this change.

packages/nocodb/src/lib/meta/helpers/webhookHelpers.ts
pranavxc: Now it makes sense :+1: , you can update the docs part as well to make it clear - https://github.com/nocodb/nocodb/blob/develop/packages/noco-docs/content/en/developer-resources/rest-apis.md#comparison-operators

southball:
Seems that CI is broken in this PR. Is rebasing this commit enough to fix the problem?

wingkwong:
@southball Please rebase. Seems the tests failed not because of your changes.

southball:
Took a bit of time, but rebased. Before rebase the develop branch seems to have passing tests, so if there are no problems this should pass I guess.

packages/nocodb/src/lib/meta/helpers/webhookHelpers.ts
```diff
@@ -72,6 +72,26 @@ export async function validateCondition(filters: Filter[], data: any) {
...
+      case 'nallof':
+        res = !(filter.value?.split(',').map((item) => item.trim()) ?? []).every((item) =>
+          (data[field]?.split(',') ?? []).includes(item)
+        );
+        break;
+      case 'nanyof':
+        res = !(filter.value?.split(',').map((item) => item.trim()) ?? []).some((item) =>
+          (data[field]?.split(',') ?? []).includes(item)
+        );
+        break;
```
wingkwong, line 94: may move (data[field]?.split(',') ?? []) out of .map since the value won't be changed in each iteration.

packages/nc-gui/components/smartsheet/toolbar/ColumnFilter.vue
```diff
@@ -213,7 +218,12 @@ defineExpose({
               dropdown-class-name="nc-dropdown-filter-comp-op"
               @change="filterUpdateCondition(filter, i)"
             >
-              <a-select-option v-for="compOp in comparisonOpList" :key="compOp.value" :value="compOp.value" class="">
+              <a-select-option
+                v-for="compOp in getApplicableFilters(filter.fk_column_id)"
+                :key="compOp.value"
+                :value="compOp.value"
+                class=""
```
wingkwong, line 225: may remove it if there's no class name

packages/nc-gui/components/smartsheet/toolbar/ColumnFilter.vue
```diff
@@ -100,6 +100,11 @@ watch(
   },
 )
 
+const getApplicableFilters = (id?: string) => {
+  const colType = (meta.value?.columnsById as Record<string, ColumnType>)?.[id ?? '']?.uidt
```
wingkwong, line 104: in which case `id` would be not required?

packages/nc-gui/utils/filterUtils.ts
wingkwong: may use `UITypes.MultiSelect`

packages/nc-gui/components/smartsheet/toolbar/ColumnFilter.vue
pranavxc: if user is not selected a column id will be undefined/null

pranavxc (approved):

PR #4425, opened 2022-11-18
Use cell inputs for filters input type

ketewan:
## Change Summary

It seems to be a good idea to replace the text input cell-type input for almost all data types. Cell components can be reused.

## Change type

## Test/ Verification

Provide summary of changes.


packages/nc-gui/components/smartsheet/toolbar/FilterInput.vue
```diff
@@ -0,0 +1,90 @@
+<script setup lang="ts">
+import type { ColumnType } from 'nocodb-sdk'
+import { useColumn } from '~/composables/useColumn'
+import { ColumnInj, EditModeInj, computed, provide, toRef } from '#imports'
+import type { Filter } from '~/lib'
+import SingleSelect from '~/components/cell/SingleSelect'
```
ketewan, line 6: For some reason explicit imports from '#components' doesn't work here, so had to list that ugly

wingkwong (changes requested):
Please update corresponding playwright tests.

See `nocodb/tests/playwright/README.md` for more info.

mustafapc19:
There is some UI/UX issues with directly using `Cells` component for Filter input

Here we can create options right from the filter view, which is not the best way UX wise and also since modal is small the text is really cramped in. 
image
also if click out side, with option modal opened, it stays on
image

Styling issues
image

This minor but better would be to have no padding on `selection` border
image

wingkwong (commented):
FYI. We're not using Cypress now. Just tested a bit. Here's my comments.

1. Select option creation should be disabled in this case
2. LTAR columns are failing
image
3. hard to read the value of single select

image

4. after clicking outside, the dropdown is still here

image

wingkwong:
from my previous comment, item 1, 3, 4 have been resolved but item 2 is still reproducible.

packages/nc-gui/components/smartsheet/toolbar/FilterInput.vue
```diff
@@ -0,0 +1,186 @@
...
+  isInt,
+  isFloat,
+  isTextArea,
+}
+
+type FilterType = keyof typeof checkTypeFunctions
+
+const { sqlUi } = $(useProject())
+
+const abstractType = $computed(() => (column.value && sqlUi ? sqlUi.getAbstractType(column.value) : null))
```
wingkwong, line 84: The root cause to the issue that I've posted is that `column.value.dt` is `null` for LTAR columns because they don't exist in the database and they are virtual. That's why in `getAbstractType(col)` (e.g. MysqlUI.ts in nocodb-sdk) would throw the error in the switch case because `col.dt` is `null` and `null` doesn't have  `toLowerCase()`.

https://github.com/nocodb/nocodb/blob/159908b103592040981af3b2f706f182f896f2c4/packages/nocodb-sdk/src/lib/sqlUi/MysqlUi.ts#L874-L875

wingkwong (dismissed):

wingkwong:
Will be excluded in the upcoming release. Need to test thoroughly.

dstala:
Thanks @ketewan @wingkwong 

Here are my observations
- [ ] For some of data fields (that either have a dropdown select / picker), we wont be able to partially configure filter data when filter operation configured as `is like`. An example is paste below with existing official version of NocoDB
  - So, we now have to decide if it's required to continue support for such partial configuration?
  - How to migrate such existing filter usage?
  - I would propose we falling back to existing official version model when filter operation chosen is `is like`. Any other alternatives, we can discuss
Screenshot 2022-12-06 at 12 26 00 PM

  
Below listed items might fall into a common category  
- [ ] Checkbox `is like` `true/false/unset` : results doesn't match with official image
- [ ] Number `is like` some number : results doesn't match with official image (even for exact match)
- [ ] `Percent` `Duration` : how to use `is like`?

ketewan:
@dstala Thanks for fair feedback!

> So, we now have to decide if it's required to continue support for such partial configuration?
Yes, I would rather return text input for all cases other that 'is equal' and 'is not equal'
Another option would be using this new inputs only for select/multiselect + maybe for a couple of other fields, not for the majority

> Checkbox is like true/false/unset : results doesn't match with official image
I guess false and unset should be grouped into one request? I found current behavior rather unpredictable, since there is equally a difference in what is returned for null and for false, but no visual difference for data in table

> Number is like some number : results doesn't match with official image (even for exact match)
Didn't notice it, could you provide any example, please?

ketewan:
@dstala @wingkwong what do you think?

wingkwong:
We need to cater different cases with different filter type. Better to have a list for the breakdown.

wingkwong:
I'll handle the rest of the items on [nocodb:enhancement/filters](https://github.com/nocodb/nocodb/tree/enhancement/filters).

Issue #4430, opened 2022-11-18
🐛 Bug: Need to click on "Cancel" after saving in order to close record editor


djmaze:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Open the editor modal for a record
2. Change any field
3. Click on "Save row" in order to save your changes
4. Click on "Cancel" in order to close the editor modal

### Desired Behavior

In my opinion, the UX is quite bad and unintuitive. 

1. When the row is saved, the modal should close immediately. That's what I would expect as a user.
2. Clicking "Cancel" after saving in order to close the modal feels wrong. I know you can also click outside of the modal in order to close it, but that is also not very intuitive to me. If you want the user to be explicit about closing the modal, there should be an "X" or "Close" button in the modal instead.

In my opinion the current behaviour is so unintuitive that I call this a bug.

### Project Details

NocoDB used as docker: true
NocoDB version: 0.98.4
Database used in NC_DC URL: pg
Project was created by clicking: New Project
Database on which spreadsheet is created: pg
OS on which NocoDB is running: Docker host is Ubuntu 22.04
Database version: Postgres 15.0

### Attachements

![editor](https://user-images.githubusercontent.com/7229/202750443-216dceef-8ca1-41bc-a814-e63c145adea9.png)

wingkwong:
The current expanded form doesn't save the data automatically. If there are lots of fields, users may need to click save row from time to time so that the data won't be lost (like accidentally click outside the modal) rather than saving the whole form in one go. 

However, I do agree it's a bit tedious if I only update a field and click 2 buttons to save and leave the modal. We'll have some internal discussions first on this matter. Thanks for reporting.

djmaze:
Just tried the new version. Really like the solution. Thank you! :100:

PR #4436, opened 2022-11-19
feat: add Save & Exit and Save & Stay buttons in Expanded Form

wingkwong:
## Change Summary

ref: #4430

By default, the button is using Save & Exit which means the form will be saved and exited after clicking. If users prefer to save the form from time to time without closing the form, they can choose Save & Stay instead.

## Change type

## Test/ Verification

image


packages/nc-gui/components/smartsheet/expanded-form/Header.vue
```diff
@@ -25,6 +25,8 @@ const { isUIAllowed } = useUIPermission()
 
 const reloadTrigger = inject(ReloadRowDataHookInj, createEventHook())
 
+const saveRowAndStay = ref(0)
```
pranavxc, line 28: We can persist the state in localstorage

pranavxc (approved):

packages/nc-gui/components/smartsheet/expanded-form/Header.vue
wingkwong: Agree. May do it later.

Issue #4470, opened 2022-11-23
🐛 Bug: NC_AUTH_JWT_SECRET is not inserted to nc_store


wingkwong:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Originally reported from https://github.com/nocodb/nocodb/discussions/4338#discussion-4552509

### Desired Behavior

`NC_AUTH_JWT_SECRET` in environment can be read and inserted in `nc_store`

### Project Details

Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
Database: **mysql2**
ProjectOnRootDB: **true**
RootDB: **mysql2**
PackageVersion: **0.99.0**

wingkwong:
Closing. The existing logic is correct.

PR #4471, opened 2022-11-23
fix(nocodb): add missing insert logic & replace hardcoded db name

wingkwong:
## Change Summary

ref: #4470

- add missing logic to handle insertion if `NC_AUTH_JWT_SECRET` is provided
- replace handcoded `nc_store` by `MetaTable.STORE`

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/lib/Noco.ts
+++ b/packages/nocodb/src/lib/Noco.ts
@@ -18,6 +18,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { NcConfig } from '../interface/config';
 import Migrator from './db/sql-migrator/lib/KnexMigrator';
 import NcConfigFactory from './utils/NcConfigFactory';
+import { MetaTable } from './utils/globals';
 import { Tele } from 'nc-help';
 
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/lib/Noco.ts
```diff
@@ -496,34 +497,34 @@ export default class Noco {
...
+        await Noco._ncMeta.metaGet('', '', MetaTable.STORE, {
+          key: 'nc_auth_jwt_secret',
+        })
+      )?.value;
+
+      if (!secret) {
+        await Noco._ncMeta.metaInsert('', '', MetaTable.STORE, {
+          key: 'nc_auth_jwt_secret',
+          value: this.config.auth.jwt.secret ?? (secret = uuidv4()),
+        });
```
pranavxc, line 510: I think storing the secret key from env is not a good idea. If someone got the DB access they could easily decrypt the project config and password, etc. By default, we are generating and storing the key if env is missing but if env is defined we should use it and ignore saving it in my opinion. @o1lab  - any thoughts

wingkwong:
Closing this PR as I misunderstood the original issue. The current logic is correct.

Issue #4277, opened 2022-11-04
Playwright improvments


mustafapc19:
Tasks

- [ ] Playwright CI : 
  - [x] Integrate Postgres  
  - [x] Integrate sqlite 
  - [x] Integrate quick test
  - [x] Will have to make xcdb project external otherwise too much load on one sqlite file
  - [ ] Reduce flakiness : 
    - [x] Disable page transition for CI nc-gui build
    - [ ] Add flaky tag to PR if any test is flaky on CI
    - [ ] Break test if un-expected network error message received
    - [ ] Have tests added in a PR run multiple times(maybe 6 times) to confirm that the added tests are not flaky (Should not increase the test time by much)
    - [x] Fix the infinitely grid loading issue. In test(ci) grid loading spinner gets stuck in loading even though the data is loaded and is visible behind the loader, two tests has been disabled because of this
  - [x] Production CI build readiness : if its feasible 
  - [ ] Document 
    - [x] Tutorial on writing test, creating and working with playwright and avoiding flakiness
    - [ ] Add test writing standards in docs and cleanup page objects and tests accordingly
  - [ ] Blog
  - [ ] Integrate turbo pack to speed up build time on CI

PR #4476, opened 2022-11-23
Added docs for playwright

mustafapc19:
Ref: #4277

- Added docs for playwright
- Removed docs for cypress
- Renamed `Testing` sub tub in `Engineering` section, to `Unit testing`, since it only has unit test docs

image


packages/noco-docs/content/en/engineering/playwright.md
```diff
@@ -0,0 +1,224 @@
+---
+title: "Playwright E2E testing"
```
wingkwong, line 2: prefer `Playwright E2E Testing`

packages/noco-docs/content/en/engineering/playwright.md
```diff
@@ -0,0 +1,224 @@
...
+description: "Overview to playwright based e2e tests"
+position: 3260
+category: "Engineering"
+menuTitle: "Playwright E2E testing"
```
wingkwong, line 6: prefer `Playwright E2E Testing`

wingkwong (commented):
I think we no longer need `nocodb/tests/playwright/README.md` as I can see the documentation shows more details or we can just point the link to the documentation in README.md. Besides, we don't need to update both if there are changes.

Issue #3296, opened 2022-08-23
case-sensitive collation for database causing issues for MS SQL


AshleyJackson:
**Please enter the following details**

```
NocoDB used as docker : true
NocoDB version : 0.92.4
Database used in NC_DB URL  : mssql
Project was created  by clicking : New Project by connecting to external database
Database on which spreadsheet is created : sqlite3
OS on which NocoDB is running : Docker but Ubuntu 18.04
Node.js version if running as node : 
Database version : 
```

**Steps To Reproduce**

1. Have a MSSQL DB that has a case-sensitive collation.
2. Try to create a new project with an existing database.
3. Enter in the relevant correct information.
4. Create and Save Project.
5. Project fails to create with error: **Invalid object name 'information_schema.columns'**

**Expected Behavior**

Project Creation

**Screenshots**

![Screenshot from 2022-08-23 14-38-51](https://user-images.githubusercontent.com/12793989/186071328-90f07eef-7232-46b8-b382-a28ae00a27ba.png)
![Screenshot from 2022-08-23 14-38-56](https://user-images.githubusercontent.com/12793989/186071352-a991b830-9887-4aa6-8f69-36551394456e.png)
![Screenshot from 2022-08-23 14-39-47](https://user-images.githubusercontent.com/12793989/186071363-da8721c1-e567-4a02-a7cb-a70685b82a29.png)
![Screenshot from 2022-08-23 14-40-06](https://user-images.githubusercontent.com/12793989/186071375-09d71c45-381b-4208-ac34-e6b23cba3ed1.png)

nith2001:
Unless this has already been addressed, I'll try to get a start on this one.

nith2001:
Is this issue resolved?

PR #4477, opened 2022-11-23
Update MssqlClient.ts

Naude555:
## Change Summary
Changed SQL reserved words to be uppercase along with the object names of INFORMATION_SCHEMA.XYZ 
This should fix the issue
"case-sensitive collation for database causing issues for MS SQL #3296"

## Change type


o1lab:
Thank you @Naude555

Issue #4480, opened 2022-11-23
🐛 Bug: Webhook payload includes attachment json as string


djmaze:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Create a table with an attachment field
2. Create a webhook and look at the "Sample Payload". For the attachment field, you can see that the array is encoded as a json string instead of a json object. See this snippet:

```json
    "title": "test",
    "file": "[{\"url\":\"https://nocodb.com/dummy.png\",\"title\":\"image.png\",\"mimetype\":\"image/png\",\"size\":0}]",
```

### Desired Behavior

The attachment should be a _decoded_ json array including multiple objects, like this:

```json
    "title": "test",
    "file": [{"url":"https://nocodb.com/dummy.png","title":"image.jpg","mimetype":"image/jpeg","size":0}]
```

### Project Details

NocoDB used as docker: true
NocoDB version: 0.99.0
Database used in NC_DC URL: pg
Project was created by clicking: New Project
Database on which spreadsheet is created: pg
OS on which NocoDB is running: Docker host is Ubuntu 22.04
Database version: Postgres 15.0

PR #4488, opened 2022-11-25
fix(nocodb): convert attachment type in api response

wingkwong:
## Change Summary

ref: #4480

currently attachment data is stored in TEXT and parse to object in UI level. This PR is to convert it to array of objects so that users don't need to convert themselves while using our APIs.

## Change type

## Test/ Verification

webhook response

```json
[
  {
    "Id": 10,
    "Title": "a2",
    "CreatedAt": "2022-11-25 12:30:23",
    "UpdatedAt": "2022-11-25 12:31:56",
    "title72": null,
    "nc_deak__Sheet-2_id": null,
    "title7": null,
    "title6": null,
    "title8": [
      {
        "url": "http://localhost:8080/download/noco/xcdb/Sheet-1/title8/iDii5f.jpeg",
        "title": "haha.jpeg",
        "mimetype": "image/jpeg",
        "size": 6494
      }
    ]
  }
]
```

list response

```
{
    ...
    "title8": [
        {
            "url": "http://localhost:8080/download/noco/xcdb/Sheet-1/title8/44Z_bA.jpeg",
            "title": "haha.jpeg",
            "mimetype": "image/jpeg",
            "size": 6494
        }
    ]
}
```

### DIFF
--- a/packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
+++ b/packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
@@ -99,9 +99,10 @@ class BaseModelSqlv2 {
 
     qb.where(_wherePk(this.model.primaryKeys, id));
 
-    const data = (await this.extractRawQueryAndExec(qb))?.[0];
+    let data = (await this.extractRawQueryAndExec(qb))?.[0];
 
     if (data) {
+      data = this.convertAttachmentType(data);
       const proto = await this.getProto();
       data.__proto__ = proto;
     }
@@ -251,7 +252,8 @@ class BaseModelSqlv2 {
 
     if (!ignoreViewFilterAndSort) applyPaginate(qb, rest);
     const proto = await this.getProto();
-    const data = await this.extractRawQueryAndExec(qb);
+    let data = await this.extractRawQueryAndExec(qb);
+    data = this.convertAttachmentType(data);
 
     return data?.map((d) => {
       d.__proto__ = proto;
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/BaseModelSqlv2.ts
```diff
@@ -2726,6 +2734,30 @@ class BaseModelSqlv2 {
...
+      const attachmentColumns = this.model.columns.filter(
+        (c) => c.uidt === UITypes.Attachment
+      );
+      if (attachmentColumns.length) {
+        if (!Array.isArray(data)) {
+          data = [data];
+        }
+        data = data.map((d) => {
+          attachmentColumns.forEach((col) => {
+            if (d[col.column_name] && typeof d[col.column_name] === 'string') {
```
pranavxc, line 2751: I hope we have to use `column.title` here ?

djmaze:
The [sample payload](https://github.com/nocodb/nocodb/blob/5302bef7a3fe2674f5d4f368d3bb4a6967e76376/packages/nocodb/src/lib/meta/helpers/populateSamplePayload.ts#L97) is still wrong.

Also, I am unable to destructure the array in the webhook template. When `file` is an attachment field and an attachment has been uploaded, the following template

```
data.file.[0].url
```

does return an empty text when sending via the Mattermost plugin.

Or is just the Handlebar syntax not supported for destructuring of arrays?

wingkwong:
@djmaze Thanks for spotting out. The syntax is correct but there is another issue in the underlying logic. Both will be handled in this [PR](https://github.com/nocodb/nocodb/pull/4502).

djmaze:
Thanks, in 0.99.2 it works now! :+1:

PR #4504, opened 2022-11-26
Feat: Grid - Add advanced options in column header menu

pranavxc:
## Change Summary

- Advanced options in the fields menu
     - Sort option
     - Hide field
     - Duplicate column
     - Add column before/after option
 - Badge with count 

## Change type

## Test/ Verification

Screenshot 2022-11-29 at 4 43 16 PM

Screenshot 2022-11-29 at 4 43 29 PM

### TODO

- [ ] Playwright test


dstala:
@pranavxc 

- [ ] Insert a column into sort list only once, even if applied multiple times. 
  - Sort `Col1` `A->Z`, even if applied 'n' times, should have only one entry in Sort list
   
Screenshot 2022-11-29 at 8 45 55 PM

- [ ] Newly selected sort option from columns menu should be inserted to top of the list, unlike its behaviour now where new options are always added to bottom of the list. If the column was already part of sort list, it will be removed from its current position & re-inserted at the top
- [ ] Do not show sort options for LTAR fields
- [ ] Show newly added menu on right clicking column header as well
- [ ] Skip showing `Duplicate` option for LTAR, Lookup instead of notification `Not available at the moment`
- [ ] Duplicate didn't work for `ncRecordId` column (import primary key); subsequent attempt threw duplicate column error
- [ ] Duplicate failed for single-select and multi-select (SQLite root DB)
```
columnAdd :: TypeError: Cannot read properties of undefined (reading 'options')
    at columnAdd (/Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/lib/meta/api/columnApis.ts:535:51)
    at /Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/lib/meta/helpers/catchError.ts:6:14
```
 - [ ] Insert before didn't work for first column.

dstala (approved):

wingkwong (commented):
Not sure if it is by design. If I sort a column asc, then desc, supposedly the previous one should be removed rather than adding on top of it.

image

pranavxc:
> Not sure if it is by design. If I sort a column asc, then desc, supposedly the previous one should be removed rather than adding on top of it.
> 
> image

Will do that in. a follow-up PR :+1:

wingkwong (approved):

PR #4521, opened 2022-11-28
Added stress testing of newly added test to CI

mustafapc19:
Ref: https://github.com/nocodb/nocodb/issues/4277

- Run newly added/modified tests on a PR 4 times per dbtype on CI
- Detecting what tests to stress test is a bit primitive, where we use `git diff` to detect `test("foo", ` on added changes and run foo

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -10,6 +10,7 @@ on:
             - "packages/nc-gui/**"
             - "packages/nocodb/**"
             - ".github/workflows/ci-cd.yml"
+            - "tests/playwright/**"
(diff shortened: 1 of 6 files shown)


wingkwong (commented):
any reason to fetch from `develop` instead of the branch that the action is running? Let's say I added new tests in a new branch, in this case, those tests won't be executed.

mustafapc19:
@wingkwong Actually fetching `develop` because we git diff on the given PR's branch and `develop`. Fetching just to be safe, if by chance in the CI instance, `develop` branch is not there locally.

wingkwong (approved):

Issue #4528, opened 2022-11-29
🐛 Bug: Attachment photo file size issue


geromyx:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Create table with one column file attachment. create a form from this table. Open the form on a smartphone. Try uploading a photo in the attachment field. The error message that this file is too big appears.

### Desired Behavior

With today's smartphones, the resolution of the photos cannot be reduced. Therefore larger files up to at least 10 mb should be supported.

### Project Details

NocoDB installed on Ubuntu Server 20.04
NocoDB Version: 0.99.2
Database: mysql

### Attachements

![Attachment photo file size issue](https://user-images.githubusercontent.com/119406652/204491711-3de8919d-524a-43fd-9a2b-070af97a4a9a.jpg)

RusackN1:
The problem appeared again in v0.106.0

Another problem is that on the iPhone you can't add the file, it immediately opens camera image and you have to take a picture. Can it be fixed?

PR #4538, opened 2022-11-30
fix(test): Made default file field size limit to 20MB and is configurable using NC_ATTACHMENT_FIELD_SIZE env variable

mustafapc19:
## Change Summary

Ref: #4528

- Made default file field size limit to 20MB and is configurable using NC_ATTACHMENT_FIELD_SIZE env variable
- Added playwright tests for it
- Mysql database resets every time a tests starts instead of relying on the audits
- Added in docs 
image

## Change type


o1lab:
* Lets default to 20 MB and users can increase if they need to.
* Lets have column name `attachment` part of env variable. NC_FILE_FIELD_SIZE -> NC_ATTACHMENT_FIELD_SIZE

packages/nocodb/src/lib/constants/index.ts
```diff
@@ -1,2 +1,5 @@
 export const NC_LICENSE_KEY = 'nc-license-key';
 export const NC_APP_SETTINGS = 'nc-app-settings';
+export const NC_FILE_FIELD_SIZE = process.env['NC_ATTACHMENT_FIELD_SIZE']
+  ? Number(process.env['NC_ATTACHMENT_FIELD_SIZE'])
+  : 20 * 1024 * 1024; // 20 MB
```
wingkwong, line 5: If `process.env['NC_ATTACHMENT_FIELD_SIZE']` is not a number, says `abc`, then  `NC_ATTACHMENT_FIELD_SIZE` would be `NaN`.

You can change this to the following so that any invalid value would turn to the default one.

```
+process.env['NC_ATTACHMENT_FIELD_SIZE'] || 20 * 1024 * 1024
```

packages/nocodb/src/lib/constants/index.ts
```diff
@@ -1,2 +1,5 @@
 export const NC_LICENSE_KEY = 'nc-license-key';
 export const NC_APP_SETTINGS = 'nc-app-settings';
+export const NC_FILE_FIELD_SIZE = process.env['NC_ATTACHMENT_FIELD_SIZE']
```
wingkwong, line 3: I prefer to keep the same name to avoid confusion.

wingkwong (commented):
please also update the doc to show that we have such option.

wingkwong (approved):

Issue #4529, opened 2022-11-29
🐛 Bug: On going issues in develop ( >0.99.2 )


o1lab:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

- [ ] Quick tab opening. After 7th or 8th tab switching breaks.
- [x] Opening project sometimes - doesn't open the page it completely.

### Desired Behavior

.

### Project Details

Develop

mertmit:
Second one resolved with https://github.com/nocodb/nocodb/commit/6f0deb5e657a3214da61f9228fa1fa5e00441dd9
This commit is in multiple-source branch but that should be fine as it should be in soon.

o1lab:
@mertmit : cool. We plan to have multi source thing after 0.100 release.

PR #4545, opened 2022-11-30
fix: improved refresh sidebar logic

mertmit:
## Change Summary

Re #4529
- Refresh sidebar only if next view doesn't have sidebar

## Change type


dstala (approved):

pranavxc (approved):

dstala:
tried few times; looked good. will continue to be on this branch. lets see if we see this issue.

Issue #4570, opened 2022-12-02
🔦 Feature: change the error prompt for forgot password


whyteserious:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

After many people forget the password, it will be prompted with “Some error occured”.  They always ask me for how to change the password.
“forgot password”  in latest version,  the password need  "At least 8 letters".
in previous version,  the password need  "Atleast 8 letters. One Uppercase Letter. One special letter".

### Suggested Solution

Maybe it can have an audit at the reset page or give a complete prompt of the error when change the passwd error

PR #4571, opened 2022-12-02
feat(nocodb): include error message for resetting password

wingkwong:
## Change Summary

ref: #4570 

- include error message

## Change type

## Test/ Verification

image

### DIFF
--- a/packages/nocodb/src/lib/meta/api/userApi/ui/auth/resetPassword.ts
+++ b/packages/nocodb/src/lib/meta/api/userApi/ui/auth/resetPassword.ts
@@ -86,7 +86,11 @@ export default `<!DOCTYPE html>
             });
             this.success = true;
           } catch (e) {
-            alert('Failed to reset password: ' + e.response.data.msg)
+            if (e.response && e.response.data && e.response.data.msg) {
+              alert('Failed to reset password: ' + e.response.data.msg)
+            } else {
+              alert('Some error occurred')
+            }
           }
         }
       }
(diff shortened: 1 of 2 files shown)


packages/nocodb/src/lib/meta/api/userApi/ui/auth/resetPassword.ts
```diff
@@ -86,7 +86,7 @@ export default `<!DOCTYPE html>
             });
             this.success = true;
           } catch (e) {
-            alert('Some error occured')
+            alert('Failed to reset password: ' + e.response.data.msg)
```
mustafapc19, line 89: Exception here will only be in the form of `e.response.data` r8? Otherwise might cause issues

mustafapc19, line 89: Exception here will only be in the form of `e.response.data` r8? Otherwise might cause issues

wingkwong: `e.message` would return a general error message, i.e. `Request failed with status code 400` which is not a good message. We need to catch the error message thrown from API. That's why `e.response.data.msg` is used.

mustafapc19: What I meant was if there is an exception not from API call `e.response.data.msg` will error out. Probably won't happen

mustafapc19 (approved):

PR #4631, opened 2022-12-07
Fix getColVal in parserHelpers.ts -> col index can be 0 -> so check only that its not undefined

semmjon:
## Change Summary

Excel / CSV Import not working with auto type detection.
The getColVal returns the whole row for the first col, because the index is null.

So i do instead of:
```
row && col ? row[col] : row
```
this
```
row && col !== undefined ? row[col] : row
```

## Change type

- [X ] fix: (bug fix for the user, not a fix to a build script)

## Test/ Verification

Tested locally.

## Error - Image:
image


wingkwong (changes requested):
please don't commit unrelated files.

semmjon:
sorry, it was probably unintentional .

wingkwong:
The expected file changes should be just 1. You are removing the entire package file now.

semmjon:
ok got it I'll be more careful next time ;)

wingkwong:
Now you removed the fix :D

wingkwong:
Looks fine now.

PR #4633, opened 2022-12-07
Possible fix for API requests with wrong baseId ("undefined").

semmjon:
## Change Summary

For some reason the baseId parameter is undefined in the frontend when calling the tableCreate function.
As a result, the request-url is not correct and the backend parses the baseId as an "undefined" string, but cannot find it in the list of bases.

I don't know exactly where to fix the bug because I couldn't find out exactly where the frontend gets the baseId from. However, this is also a possible (quicker) fix.

## Change type

## Test/ Verification

- locally tested

image


wingkwong (changes requested):
Your changes only prevent the error from being thrown but not resolve the root cause. The reason causing `undefined` is due to the missing parameter `baseId` in `packages/nc-gui/pages/[projectType]/[projectId]/index/index/index.vue`

https://github.com/nocodb/nocodb/blob/65a646c1d48c975cd561d10a3d7a7b2102ca897a/packages/nc-gui/pages/%5BprojectType%5D/%5BprojectId%5D/index/index/index.vue#L126-L131

In UI, that is the page when there is no active tab.

image

Table Creation from `Add new table` would cause the error. The fix is simple. As we don't have the selected `baseId`, we have to use the default one, i.e. `base.value[0].id`.

In short,

- revert your change
- go to the above file path
- change L22 to `const { bases, isSharedBase } = useProject()`
- pass `baseId': bases.value[0].id` after L130 
- test it

semmjon:
@wingkwong sorry for the day's delay. Is it okay now?

wingkwong (approved):

semmjon:
Sorry, I didn't check that earlier, but the baseId is still undefined. I guess a default is never initialized.

PR #4634, opened 2022-12-07
feat: create internal project config inflection

willnewii:
## Change Summary

image

## Change type

- [x] feat: create internal project  add inflection config


wingkwong:
By the design of Inflection, it only applies to the existing tables. That's why we only added it to external projects. For internal one, the inflection doesn't change anything. It simply takes whatever you input. Thanks for your effort but sorry to say we have to close this PR. Next time you may create an issue first before implementing.

PR #4656, opened 2022-12-14
fix multiselect

ketewan:
## Change Summary

Locked cells break multiselect behavior since they stop click event.

https://user-images.githubusercontent.com/31967149/207723328-24e05599-5f03-4774-8396-f79c4cf9b2e0.mov

## Change type

## Test/ Verification

Provide summary of changes.

"By sense" it would be much better, if activeCell was computed upon selectedRange, but since selectedRange should be updated straightaway, and cell should become active only after child clicks are handled (to maintain old behavior, e.g. to edit date field, two clicks had to be done -- one to select cell, and another one to open date dropdown). The same could be achieved with async computed, but I didn't want to install additional library.

A little more "oddness" adds the fact, that
a) we want cell become active BEFORE dblclick handler
b) for MultiSelect data cell mouseDown event is prevented inside the component, so it's necessary to handle range both on click and on mouse down

### DIFF
--- a/packages/nc-gui/components/smartsheet/Grid.vue
+++ b/packages/nc-gui/components/smartsheet/Grid.vue
@@ -172,7 +172,7 @@ const getContainerScrollForElement = (
   return scroll
 }
 
-const { selectCell, startSelectRange, endSelectRange, clearSelectedRange, copyValue, isCellSelected, selectedCell } =
+const { isCellSelected, activeCell, handleMouseDown, handleMouseOver, handleCellClick, clearSelectedRange, copyValue } =
   useMultiSelect(
     meta,
     fields,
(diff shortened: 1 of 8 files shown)


packages/nc-gui/components/smartsheet/Grid.vue
```diff
@@ -227,29 +228,34 @@ const { selectCell, startSelectRange, endSelectRange, clearSelectedRange, copyVa
         switch (e.key) {
           case 'ArrowUp':
             e.preventDefault()
-            selectedCell.row = 0
-            selectedCell.col = selectedCell.col ?? 0
+            clearSelectedRange()
+            $e('c:shortcut', { key: 'CTRL + ArrowUp' })
```
wingkwong, line 232: please remove this line

wingkwong (changes requested):

Issue #1544, opened 2022-03-29
[Feature] Helm chart


just1689:
**Need for this feature**
A helm chart makes it easy to install in Kubernetes clusters

**Is your feature request related to a problem? Please describe.**
Running in Kubernetes using Helm requires a Helm chart.

**Describe the solution you'd like**
I'd like to contribute the initial Helm chart. I'm sure the team and the community will help shape it into what it should be.

**Describe alternatives you've considered**
Someone else could do it if you would prefer. 

**Additional context**
NA

just1689:
Let me know. Would be happy to contribute the chart

o1lab:
Available [here](https://github.com/nocodb/nocodb/tree/develop/charts/nocodb)

erikalfthan:
Is it packaged somewhere? I.e. is there a publicly available path? `helm repo add nocodb https://<path>`
(Or maybe I am just a helm newbie - is it possible to install the chart directly from git?)

debugger24:
@erikalfthan you can download the repo run helm install.

debugger24:
We should publish this helm chart so it is easy to install and extend.

vjm:
has this been published anywhere?

koalalorenzo:
Any update on this?

Serubin:
Bumping this - this chart should be published. It cannot be easily integrated into GitOps systems.

PR #4693, opened 2022-12-23
feat: Adds helm chart

RuriRyan:
Initial work was done by https://github.com/one-acre-fund/oaf-public-charts/tree/main/charts/nocodb

The changes include mostly fixes and the addition of missing features.

It currently requires a ReadWriteMany volume to support multiple replicas, as uploaded files are stored there.

I don't have much experience with NocoDB itself. Additional information are very welcome.

## Change Summary

Implements #1544

## Change type

- [x ] feat: (new feature for the user, not a new feature for build script)

### DIFF
--- a/charts/nocodb/Chart.yaml
+++ b/charts/nocodb/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+appVersion: 0.100.2
+dependencies:
+- condition: postgresql.enabled
+  name: postgresql
+  repository: https://charts.bitnami.com/bitnami
+  version: ~11.6.6
+description: A Helm chart for Kubernetes
+maintainers: []
+name: nocodb
+type: application
+version: 0.3.0
--- a/charts/nocodb/.gitignore
+++ b/charts/nocodb/.gitignore
@@ -1,0 +1,2 @@
+charts/
+Chart.lock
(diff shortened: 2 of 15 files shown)


starbirdtech383:
@RuriRyan - I had a quick glance and it looks like a good start and nice work on the charts. It would be good if you can port [README](https://github.com/one-acre-fund/oaf-public-charts/blob/main/charts/nocodb/README.md) as well with notes covering purpose and how-to

charts/nocodb/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+appVersion: 0.100.2
```
wingkwong, line 2: can this value be dynamic or point to latest version?

RuriRyan: what exactly do you mean by "dynamic"? It has to have some value of course, you cannot let helm dynamically fetch this. You could do "latest", but then if something changes in the application it may break older chart versions.
You can of course use github actions to replace this value before publishing the chart somewhere.

o1lab: I've very basic idea about helm. Possibly wing's question was to send the version from command line or could it be wired to be default set image.tag ? like `helm install noco-helm path/to/chart --set image.tag=0.xyz` ?

Issue #3927, opened 2022-10-04
Quick Import - AIRTABLE (Cannot read property 'length' of undefined)


o1lab:
### Discussed in https://github.com/nocodb/nocodb/discussions/3925

<div type='discussions-op-text'>

<sup>Originally posted by **kmataeh23** October  4, 2022</sup>
Hi Nocodb team. I am trying to import an air table to noco using its api key and shared base ID but am getting a "cannot read property 'length' of undefined" error in the logs section. Any advice on a workaround for this error?
<img width="1045" alt="read property length undefined" src="https://user-images.githubusercontent.com/70729030/193707852-11b14a29-c632-4f6b-9e67-36d7b07da960.png">
</div>

dstala:
@kmataeh23
You can DM me additional details on discord (@dstala). If you can share the redacted schema, that would be helpful for us to investigate

KevinGutowski:
I believe this is caused by not using the Airtable **_Base_** URL but rather trying to import a shared Airtable **_View_**

PR #4705, opened 2022-12-27
fix: proper error if provided url is not shared base

mertmit:
## Change Summary

Re #3927
If baseId is not found in given URL (shared view etc.) raise an error including instructions to get proper link.

## Change type


dstala (approved):
worked as expected for shared view. see if we can also introduce appropriate error message for invalid shared base ID

dstala (approved):

Issue #4722, opened 2022-12-29
🐛 Bug: inconsistent project tree view after project rename


dstala:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

https://www.loom.com/share/b24f8e1915344a6995559f7e9989d528

### Desired Behavior

project tree should remain consistent after rename

### Project Details

Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
RootDB: **sqlite3**
PackageVersion: **0.101.0-beta.0**

PR #4726, opened 2022-12-30
fix: route problems after editing project name

mertmit:
## Change Summary

Re #4722 
- Change useRoute in useProject composable to vue-router one (https://github.com/nuxt/framework/issues/6646)
- Disable lazy load in TreeView as it makes injectionState inconsistent for useProject

## Change type


dstala (changes requested):
https://www.loom.com/share/195f4af880734b878f5cfe9901bfdaf9

dstala (approved):

PR #4781, opened 2023-01-08
Undo select option removal

ketewan:
## Change Summary

Since option removal is a destructive operation in can be a good idea to warn a user beforehead

## Change type

## Test/ Verification

Provide summary of changes.

image

### DIFF
--- a/packages/nc-gui/components.d.ts
+++ b/packages/nc-gui/components.d.ts
@@ -124,6 +124,7 @@ declare module '@vue/runtime-core' {
     MdiArrowDownDropCircleOutline: typeof import('~icons/mdi/arrow-down-drop-circle-outline')['default']
     MdiArrowExpand: typeof import('~icons/mdi/arrow-expand')['default']
     MdiArrowLeftBold: typeof import('~icons/mdi/arrow-left-bold')['default']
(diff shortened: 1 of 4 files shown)


o1lab:
Clicking cancel wont save it. Having a confirm for a leaf level option will come in between good ux.

ketewan:
@o1lab You suggest saving/cancelling option removal at once on a leaf level, not waiting for a main "save" button to be clicked, or?

ketewan:
@o1lab 
As we discussed, I removed the popup solution and changed it to have temporarily deleted options with the possibility to undo before save

https://user-images.githubusercontent.com/31967149/212680221-7d302c4f-f69f-437d-8f79-bf9b292b0789.mov

note: I don't like the solution much because of 2 sources of true (renderedOptions and options) and the need to sync them manually, but otherwise it will end up either in deep watch for renderedOptions (if we add watch to change the original options), which is non-performant, or big refactoring (a better option would be to normalize the data before saving, but with the current code it would either make the code messier or require more general refactoring, and I don't think the task is worth it)

o1lab:
@pranavxc

pranavxc (approved):

Issue #4793, opened 2023-01-10
🐛 Bug: Unexpected project_id in created filters


wingkwong:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Reported by @ketewan 

Steps to reproduce:

- run with NC_DISABLE_CACHE=true
- login with any role but super user
- make sure there are at least 2 tables
- go to the “later created” table (which will NOT be returned with query.first). make sure user has “filter access” to this table and not the “query.first()” one
- create first filter
- add second filter and change operation before doing anything else
- try to delete or change newly added filter

### Desired Behavior

No such error being thrown

### Project Details

Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
RootDB: **pg**
PackageVersion: **0.101.0-beta.0**

PR #4795, opened 2023-01-10
fix: filters logic

wingkwong:
## Change Summary

ref: #4793 

`project_id` in `nc_filter_exp_v2` is incorrect if users input the filter value first (since `fk_column_id` is missing which causes picking the first project id). This PR does the following changes

- upgrader to correct the `project_id` based on either `fk_view_id`, `fk_hook_id` or `fk_column_id` so that users can update / delete the problematic filters.
- disable filter value input if filter field is not chosen (in frontend)
- if `fk_column_id` is not provided, `project_id` should be retrieved from either `fk_view_id` or `fk_hook_id` (in backend).

## Change type

## Test/ Verification

the first 3 filters can be updated / deleted successfully

image

### DIFF
--- a/packages/nocodb/src/lib/meta/helpers/extractProjectIdAndAuthenticate.ts
+++ b/packages/nocodb/src/lib/meta/helpers/extractProjectIdAndAuthenticate.ts
@@ -73,9 +73,14 @@ export default async (req, res, next) => {
       const column = await Column.get({ colId: params.columnId });
       req.ncProjectId = column?.project_id;
     } else if (params.filterId) {
-      const filter = await Filter.get(params.filterId).then((f) =>
-        f?.getColumn()
-      );
+      const filter = await Filter.get(params.filterId).then(async (f) => {
+        if (f.fk_view_id) {
+          return await View.get(f.fk_view_id);
+        } else if (f.fk_hook_id) {
+          return await Hook.get(f.fk_hook_id);
+        }
+        return await Column.get({ colId: filter.fk_column_id });
+      });
       req.ncProjectId = filter?.project_id;
     } else if (params.filterParentId) {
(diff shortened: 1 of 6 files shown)


packages/nocodb/src/lib/meta/helpers/extractProjectIdAndAuthenticate.ts
```diff
@@ -73,7 +73,14 @@ export default async (req, res, next) => {
...
     } else if (params.filterId) {
-      const filter = await Filter.get(params.filterId);
+      const filter = await Filter.get(params.filterId).then(async (f) => {
+        if (f.fk_view_id) {
+          return await View.get(f.fk_view_id);
+        } else if (f.fk_hook_id) {
+          return await Hook.get(f.fk_hook_id);
+        }
+        return await Column.get({ colId: filter.fk_column_id });
+      });
```
pranavxc, line 83: this change is not required since we are updating filters with the correct project id

packages/nocodb/src/lib/version-upgrader/ncFilterUpgrader.ts
```diff
@@ -0,0 +1,34 @@
...
+      model = await Column.get({ colId: filter.fk_column_id }, ncMeta);
+    } else {
+      continue;
+    }
+    if (filter.project_id != model.project_id) {
+      await ncMeta.metaUpdate(
+        null,
+        null,
+        MetaTable.FILTER_EXP,
+        { project_id: model.project_id },
```
pranavxc, line 29: Include the base id as well

pranavxc (changes requested):

pranavxc (approved):

Issue #4185, opened 2022-10-26
🐛 Bug: Attachment URLs are stored with base_url prefixed


mertmit:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Upload any attachment
This causes undesired behavior when reached behind a proxy and later base_url is changed

### Desired Behavior

We should only keep path (we only allow internal attachments), and use append base_url on front-end / middleware

### Project Details

Develop

wingkwong:
Here's the summary after our discussion.

- use `ncSiteUrl` as base url
- for creating a new attachment, the format would be like

```json
[
    {
        "path": "download/noco/xcdb/Sheet-1/title5/39A410.jpeg",
        "title":"foo.jpeg",
        "mimetype":"image/jpeg",
        "size":6494
    }
]
```

For the existing data, when we read / update an attachment, we do the following

- check if `path` is present in attachment object
  - if path is present, check if `url` field is present, if so, delete `url` (during update)
  - if not, use `url`, else construct the link from `ncSiteUrl` and `path` (during read)

cc: @mertmit @pranavxc

o1lab:
Can we write an upgrade function for this fix which does the following
- For all attachment fields, if URL is present lets remove the `url`

wingkwong:
@o1lab In this case we need to find out all attachment columns in users' meta databases (nc_columns_v2) and make connections to the external databases (start new transactions ..). The entire upgrade is happening within a transaction, which may cause some issues.

o1lab:
Upgrade happens on the boot - so holding the transaction within xcdb does not affect any operations in backend or frontend. And external database updates - we do not require it to be done via transaction.

Issue #3876, opened 2022-09-30
🔦 Feature: Configure path and filename when uploading files


PrasanthKumar77:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

i want to give custom path for each project and custom name for each attachment when uploading attachments to s3.
i cannot identify what is  in a link from the random name attached to file in s3.

### Suggested Solution

for project
i should be able to update s3 path in project settings
for attachments
when uploading attachments give a option to enter manual name or suggest to go with random name and the name should be added with a random suffix to ensure uniqueness.

PrasanthKumar77:
Hi @wingkwong, May i know if this is in your roadmap for future releases?

wingkwong:
The latter one has been implemented in a PR. The first one is pending for re-enginnering.

graphixillusion:
Interesting feature. Is it possible to add an extra option to keep the attachment's name untouched?

o1lab:
We do not envision to support this.

Issue #2905, opened 2022-07-31
[Feature] Limit attachments


LucBerge:
## Recommending Features

**Need for this feature**
I would like to be able to limite the attachments in a column:

Possibility to limit 
- [ ] The number of attachments to a given number.
- [ ] To allow only specified file formats
- [ ] To refuse only specified file formats
- [ ] To limit the file size

**Describe the solution you'd like**
When editing an attachment column, the limits can be defined.

LucBerge:
Thanks @djmaze. I have added this to the original post

wingkwong:
> To allow only specified file formats
> To refuse only specified file formats

@LucBerge What is the use case for showing both at the same time? Supposedly `Allowed` and `Not Allowed` are two disjoint sets. If one is allowed, then it won't be not allowed, and vice versa. I think we can just show the list of allowed formats (by default all are allowed) so that all unselected formats will be considered as `Not Allowed`. What do you think?

Issue #335, opened 2021-07-10
Feature: Select multiple files and download attachments.


jreniel:
The only way to download attachments as of 0.9.43 is to right click and save as individually. Having a multiselect tool for attachment downloads would be very useful.

PR #4820, opened 2023-01-13
feat: attachments

wingkwong:
## Change Summary

- ~~(partially) https://github.com/nocodb/nocodb/issues/3876~~
  - ~~Rename:~~
    - ~~When uploading a file open prompt to rename the file~~
    - ~~file name should be unique in cell~~
    - ~~file name must exist~~
    - ~~When rename is canceled, the original file name will be used~~
-  New Attachment Options (EE only): https://github.com/nocodb/nocodb/issues/2905 (will appear under `show more`)
    - mime type validation when uploading an attachment
    - Maximum Number of Attachments per cell
      - default limit for non-EE: N/A; for EE: 50 which can be configured by `NC_MAX_ATTACHMENTS_ALLOWED`
    - Maximum File Size per file
      - default limit for non-EE: N/A; for EE: 20MB which can be configured by `NC_ATTACHMENT_FIELD_SIZE`
- Base url issue : #4185 (see the issue for the proposed solution flow)
- Multiselect for downloading attachments #335
- Display local files even a storage plugin is configured. #4880
- Misc
  - Allow users to update attachment title (not filename)
  - add modal for confirmation when deleting an attachment 
  - the new attachment filename will be a random string (18 characters + extname)

## Change type

## Test/ Verification

Provide summary of changes.

TODO: patch attachment data efficiently


packages/nocodb/src/lib/constants/index.ts
```diff
@@ -1,4 +1,5 @@
 export const NC_LICENSE_KEY = 'nc-license-key';
 export const NC_APP_SETTINGS = 'nc-app-settings';
-export const NC_ATTACHMENT_FIELD_SIZE =
-  +process.env['NC_ATTACHMENT_FIELD_SIZE'] || 20 * 1024 * 1024; // 20 MB
+export const NC_ATTACHMENT_FIELD_SIZE = +(
+  process.env['NC_ATTACHMENT_FIELD_SIZE'] || 20 * 1024 * 1024
+); // 20 MB
```
pranavxc, line 5: Won't work as expected if `NC_ATTACHMENT_FIELD_SIZE` is a wrong value. For example, if the value is `abc` then the final output will be `NaN`. If `NaN` won't cause any issues then it's fine.

wingkwong: I think it should be `+process.env['NC_ATTACHMENT_FIELD_SIZE'] || 20 * 1024 * 1024` instead.

dstala (changes requested):
- [ ] minor change- bulk download, if a file selected was removed - next file in that index post delete shouldn't be auto selected
https://loom.com/share/f93e4a1b26464ff1bd9af0c4bbce8409

- [ ] Shared form attachment submission allows file of not supported type to be attached. 

- [ ] Copy paste also allowed multiple files to be inserted onto a record
https://www.loom.com/share/6e5ae691523d4b8fa1f1f9cb2de38ab5

- [x] expand modal rename of attachment files - two files can have same name. please confirm if its by design
- [x] attachment column,  download as CSV : displays `undefined` instead of path. Earlier sample `4.json(http://localhost:8080/download/')`
Screenshot 2023-01-24 at 12 37 48 PM

dstala (approved):

Issue #4662, opened 2022-12-16
🐛 Bug: Modals/Popup about Table dependencies has some flaws


spaudanjo:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

* Try to delete a table which is a dependency of another table
* Check the Modal/Popup that appears to inform about this dependency and the need to resolve it at first

### Desired Behavior

* It should have some form of `Ok` or `Close` button to get rid of the modal (which otherwise stays for quite a long time, like 6 seconds) 
* The respective table names should ideally also be listed wit the human readable names, not only the cryptic table-ids

### Project Details

Any project that has tables with dependencies between each other

### Attachements

<img width="779" alt="image" src="https://user-images.githubusercontent.com/249210/208087697-5e828085-67e0-4296-9f3a-ffa3bcac0f9b.png">

PR #4866, opened 2023-01-17
fix: hide system relation column from LTAR popup

mertmit:
## Change Summary

Re #4662
- Hide auto-created mm system column from warning

## Change type

### DIFF
--- a/packages/nc-gui/composables/useTable.ts
+++ b/packages/nc-gui/composables/useTable.ts
@@ -1,5 +1,5 @@
 import type { LinkToAnotherRecordType, TableType } from 'nocodb-sdk'
-import { UITypes } from 'nocodb-sdk'
+import { UITypes, isSystemColumn } from 'nocodb-sdk'
 import {
   Modal,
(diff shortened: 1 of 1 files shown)


packages/nc-gui/composables/useTable.ts
```diff
@@ -84,7 +84,7 @@ export function useTable(onTableCreate?: (tableMeta: TableType) => void, baseId?
       async onOk() {
         try {
           const meta = (await getMeta(table.id as string, true)) as TableType
-          const relationColumns = meta?.columns?.filter((c) => c.uidt === UITypes.LinkToAnotherRecord)
+          const relationColumns = meta?.columns?.filter((c) => c.uidt === UITypes.LinkToAnotherRecord && !c.system)
```
wingkwong, line 87: may take a look if `isSystemColumn(col)` is more suitable or not.

mertmit: fixed and rebased

PR #4883, opened 2023-01-19
fix: SQL server column alter and create bugs

pranavxc:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.


packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
```diff
@@ -306,7 +190,7 @@ export class MssqlUi {
         return '';
 
       case 'decimal':
-        return '';
+        return '10';
```
mertmit, line 193: We return number for nvarchar and varchar, if it is more appropriate to return string we might fix others

mertmit (changes requested):
Select options are all lower-case
image

Behavior in develop
image

mertmit (approved):

packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
pranavxc: @mertmit Actually, the type doesn't matter here except `0`(number) since its falsy value may get ignored. I will make it consistent anyway.

Issue #4881, opened 2023-01-19
🐛 Bug: Checkbox sort in PG locates NULL & false differently


dstala:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Extension to 
https://github.com/nocodb/nocodb/pull/4861

1b - false
1d, 1e - null

ascending
![Screenshot 2023-01-19 at 2 25 57 PM](https://user-images.githubusercontent.com/86527202/213414020-9c9f6b53-c2ef-4b8f-b348-4e04433a8a4f.png)

descending
![Screenshot 2023-01-19 at 2 26 05 PM](https://user-images.githubusercontent.com/86527202/213414062-9119d406-9297-4475-8cdf-9cebb46ffb93.png)

### Desired Behavior

From user POV - false & null should be together
For XCDB - we can mark default value for any new checkbox column to False instead of NULL

### Project Details

https://github.com/nocodb/nocodb/pull/4861
PG database

PR #4885, opened 2023-01-19
fix: Set sort order of nulls in postgres

pranavxc:
## Change Summary

- [x] #4881
- [x] Set the default value of the checkbox column as `0` / `false`
- [x] Set the default value of the rating column as `0`
- [x] On clear cell set checkbox value as `false`
- [x] On clear cell set rating value as `0`

#### Changes
- In descending order show `null` values at bottom 
- In ascending order show `null` values at top 

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
+++ b/packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
@@ -1151,6 +1151,7 @@ export class MssqlUi {
         break;
       case 'Rating':
         colProp.dt = 'int';
+        colProp.cdf = '0';
         break;
       case 'Formula':
         colProp.dt = 'varchar';
--- a/packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
+++ b/packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
@@ -1084,6 +1084,7 @@ export class MssqlUi {
       case 'Checkbox':
         colProp.dt = 'tinyint';
         colProp.dtxp = 1;
+        colProp.cdf = '0';
         break;
       case 'MultiSelect':
         colProp.dt = 'text';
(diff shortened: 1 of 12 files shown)


dstala (approved):

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/sortV2.ts
```diff
@@ -29,6 +29,9 @@ export default async function sortV2(
     const column = await sort.getColumn();
     if (!column) continue;
     const model = await column.getModel();
+
+    const nulls = sort.direction === 'desc' ? 'LAST' : 'FIRST';
```
wingkwong, line 33: what is the reason to name it `nulls`

dstala (approved):

dstala (changes requested):
with PG, output of `is like` filter is correct. 
use of `is not like` filter shows same result as `is like`

Tried for rating, decimal, number

packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
```diff
@@ -1084,6 +1084,7 @@ export class MssqlUi {
       case 'Checkbox':
         colProp.dt = 'tinyint';
         colProp.dtxp = 1;
+        colProp.cdf = '0';
```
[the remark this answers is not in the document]
pranavxc: it's the column's default value and it should be a string, at the query level treats it as a number. For providing string it will be like `"'a'"`

packages/nocodb/src/lib/db/sql-data-mapper/lib/sql/sortV2.ts
pranavxc: it's the same argument name used by knex and it's used to define the order of `null` values.

Screenshot 2023-01-20 at 3 04 16 PM

wingkwong: suggest to add a comment

dstala (approved):

Issue #4625, opened 2022-12-07
🐛 Bug: Default value automatically sets as NULL for SingleSelect


folk-connect:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Create a Single Select field with no option and save.
After that, edit the field and add options.
Try to save

Error shows up saying default value NULL is not an option
Same happens for SingleLineText fields also. NULL is automatically set.
If you  are unable to reproduce let me know and I'll provide some screenshots

One more similar issue is initial length type is set as 45 which is too low for most of the cases. Using some environment variable can we increase the default initial length of the column

### Desired Behavior

In my opinion, setting default value as String "NULL" is not meaningful. There is no need of setting default value for enum and varchar fields

### Project Details

Node: **v14.19.0**
Arch: **x64**
Platform: **linux**
Docker: **false**
Database: **mysql2**
ProjectOnRootDB: **true**
RootDB: **mysql2**
PackageVersion: **0.100.2**

wingkwong:
> Create a Single Select field with no option and save.
> After that, edit the field and add options.
> Try to save

I believe you haven't inputted anything? `NULL` is the default value here because when you haven't selected any option, the cell value would be NULL. That's why you cannot leave the option blank. I don't see why there is no need of setting default value for it.

![image](https://user-images.githubusercontent.com/35857179/206120688-0d16867a-8496-4528-bbf6-6d0957b7255a.png)

folk-connect:
NULL is actually not NULL but String "NULL". I am attaching video to make point more clear. This happens for SingleLineText as well when I try to edit that field

[screen-capture.webm](https://user-images.githubusercontent.com/113366228/206135456-78ae90cf-23c4-49e2-96cc-d62686ba7ae1.webm)

wingkwong:
@folk-connect thanks for the video. It's more clear now. However, I couldn't see there is `NULL` under Default value. The project info indicates you're not using Docker. May I know how you setup NocoDB?

![image](https://user-images.githubusercontent.com/35857179/206148949-250ef4f4-5693-42d0-88e1-1e404d3afeef.png)

folk-connect:
[screen-capture (1).webm](https://user-images.githubusercontent.com/113366228/206152040-195deb36-b0e2-4c9c-959b-9b518028dd19.webm)

[...]

PR #4900, opened 2023-01-21
fix: MariaDB default value

mertmit:
## Change Summary

Re #4625
- Change NULL as text to NULL for MariaDB

## Change type


wingkwong (approved):
suggest to add the comment you mentioned in the issue to the code.

Issue #4462, opened 2022-11-22
🐛 Bug: Filtering on single select empty cells are ignored 


davidpfahler:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Create a table with a single select column
2. Create rows with no value selected in this single select column
3. Create a filter for this single select column with one of these comparisons: "is empty" OR "is not equal" to some value
4. Result: No rows where the single select column is empty are returned

### Desired Behavior

The filter should return all rows where the single select column is empty.

### Project Details

NocoDB used in docker: yes

wingkwong:
Can you share the [project info](https://docs.nocodb.com/FAQs#how-to-check-my-project-info-) please?

davidpfahler:
Here is the project info:

Node: **v16.17.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
Database: **pg**
ProjectOnRootDB: **true**
RootDB: **pg**
PackageVersion: **0.98.4**

davidpfahler:
https://user-images.githubusercontent.com/908242/203331327-b5ae48a6-0eb2-4d22-9084-debc476161f5.mp4

wingkwong:
Preliminary findings:

The difference between `isEmpty` and `isNull` is that 

- originally when a new row is created, data is null if there is no input. (`isEmpty = false` and `isNull = true`)
- update the data
- clear the data
- now the data is considered as empty. (`isEmpty = true` and `isNull = false`)

However, when the single select field is cleared, the value turns `NULL`. Therefore, `isEmpty` is never in use for single select case.

PR #4901, opened 2023-01-21
fix: filtering options for select columns

mertmit:
## Change Summary

Re #4462
- Exclude incompatible options for select columns
- Add any of and not any of filtering for single select

## Change type

### DIFF
--- a/packages/nc-gui/utils/filterUtils.ts
+++ b/packages/nc-gui/utils/filterUtils.ts
@@ -41,13 +41,13 @@ export const comparisonOpList: {
     text: 'is empty',
     value: 'empty',
     ignoreVal: true,
-    excludedTypes: [UITypes.Checkbox, UITypes.Rating],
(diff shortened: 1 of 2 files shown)


dstala (approved):

packages/nc-gui/utils/filterUtils.ts
```diff
@@ -41,13 +41,13 @@ export const comparisonOpList: {
...
     ignoreVal: true,
-    excludedTypes: [UITypes.Checkbox, UITypes.Rating],
+    excludedTypes: [UITypes.Checkbox, UITypes.Rating, UITypes.MultiSelect, UITypes.SingleSelect],
   },
   {
     text: 'is not empty',
     value: 'notempty',
     ignoreVal: true,
-    excludedTypes: [UITypes.Checkbox, UITypes.Rating],
+    excludedTypes: [UITypes.Checkbox, UITypes.Rating, UITypes.MultiSelect, UITypes.SingleSelect],
```
pranavxc, line 50: the value could be `''`(empty string) as well right, so let's keep it.

Issue #2409, opened 2022-06-17
[Feature] Visual distinction between NULL and empty string


pauljura:
**Need for this feature**
If I have a table with a text column that can have NULLs, there is nothing to indicate the difference between a NULL and an empty string. It would be nice to have something that explicitly states when a value is NULL. It could be subtle, so that a table full of NULL values doesn't look busy and crowded. It should also look distinct from an actual string, so nobody could confused it with the literal string "NULL". For example, something like the word "null" (or just the letter "N") in a light muted colour, and italicised, could work well.

o1lab:
Light muted color makes sense.

zblesk:
That would still visually clutter the table, though. 

Somewhat related to that: For the common use cases (that primarily involve users looking at tables and working with them, instead of writing sql queries) I think it would work better to tweak the _handling_ of nulls. 

For instance: right now, if I create a filter in my view, select a string column, and filter by 'is null', I receive 300 rows. If I change that to 'is empty', I get 0. 
I presume that's caused by the distinction between "" and null in the DB; but no common user will understand that, and will instead wonder why these clearly empty cells don't fall under 'is empty'. 

(Not to mention that I often want exactly that - 'show me which fields do not have a value'. As it is, I have to make a disjunctive filter to get both, which is annoying, _and_ makes more complicated queries impossible. (Unless there _is_ a way to parenthesize expressions I haven't discovered yet.) 

Eh, I guess it depends on what the desired user base for NocoDB is. Just thinking out loud here.

pauljura:
It's only "clutter" if you don't understand or care about the distinction. Or if you have a random mix of "" and NULLs, in which case I would argue it's good to see the inconsistency so you can correct it.

It comes down to the audience. If your audience is non-technical and only using Nocodb as an enhanced spreadsheet, then sure, they might not care. But they may not have any NULLs anyway.

On the other hand, my use case is to expose views of several pre-existing databases, without having to code a UI specifically for this, and without exposing database credentials. We are are developers and analysts, and the distinction between "" and NULL is meaningful to us, and therefore important to see. I hope I'm not the exception here, because Nocodb seems like the ideal solution for us, and I can imagine more tech-savvy people using it the same way.

In any case, it should be easy to make the display sufficiently subtle as to not be "clutter", while still being useful for those that need it.

Also, it's funny you mention:

[...]

PR #4908, opened 2023-01-22
feat: visual distinction between NULL and empty string

mertmit:
## Change Summary

Re #2409
- Add NULL as placeholder for fields other than checkbox, rating, select  (we use null state on those types)
- Added showNull to global state to control whatever NULL values are shown or not

## Change type

## Test/ Verification

image

### DIFF
--- a/packages/nc-gui/components/dashboard/settings/Misc.vue
+++ b/packages/nc-gui/components/dashboard/settings/Misc.vue
@@ -18,7 +18,7 @@ watch(includeM2M, async () => await loadTables())
       </div>
       <div class="flex flex-row items-center w-full mb-4 gap-2">
         <!--        Show NULL -->
-        <a-checkbox v-model:checked="showNull" v-e="['c:themes:show-null']" class="nc-settings-show-null">Show NULL</a-checkbox>
+        <a-checkbox v-model:checked="showNull" v-e="['c:settings:show-null']" class="nc-settings-show-null">Show NULL</a-checkbox>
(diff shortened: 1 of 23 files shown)


dstala:
@mertmit current solution is applicable only for single line text & long text. (and few of its derivatives like phone number). how about other data types- select, email, url, number, date, checkbox? was this intended by design?

All of the columns show NULL entries when filtered, though there is no visible mark on these columns

dstala (approved):

packages/nc-gui/components/dashboard/settings/Misc.vue
```diff
@@ -16,6 +16,10 @@ watch(includeM2M, async () => await loadTables())
           {{ $t('msg.info.showM2mTables') }}
         </a-checkbox>
       </div>
+      <div class="flex flex-row items-center w-full mb-4 gap-2">
+        <!--        Show NULL -->
+        <a-checkbox v-model:checked="showNull" v-e="['c:themes:show-null']" class="nc-settings-show-null">Show NULL</a-checkbox>
```
wingkwong, line 21: it's more for settings rather than themes

wingkwong (commented):
The following line is used across lots of files. Should we put it to a component instead? (so that we can change one file instead of N files and the styles can be stayed in that component rather defining it globally.) 

```
<span v-else-if="vModel === null && showNull" class="nc-null">NULL</span>
```

Issue #4616, opened 2022-12-06
🔦 Feature: make ERD diagram full screen and exportable 


vijayrathore8492:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

ERD view is very helpful in checking the relations in existing DB, however, it would be more useful if it was possible to export it. This could be a low hanging fruit to make ERD functionality more useful.

### Suggested Solution

Give an option of full-screen and export on ERD view.

PR #4911, opened 2023-01-23
feat(nc-gui): Added fullscreen support to ERD

mustafapc19:
Ref: #4616

- Fixed issue where ERD was being overflowed due to incorrect height
- Fixed close button styling issue on settings modal
image

- Added fullscreen toggle button to the top left corner
image

image

### DIFF
--- a/packages/nc-gui/components/erd/FullScreenToggle.vue
+++ b/packages/nc-gui/components/erd/FullScreenToggle.vue
@@ -8,10 +8,12 @@ const props = defineProps<{
   config: ERDConfig
 }>()
 
-const config = useVModel(props, 'config')
+const emit = defineEmits(['toggleFullScreen'])
+
+const { config } = toRefs(props)
 
 const toggleFullScreen = () => {
-  config.value.isFullScreen = !config.value.isFullScreen
+  emit('toggleFullScreen')
 }
 </script>
 
--- a/packages/nc-gui/components/erd/FullScreenToggle.vue
+++ b/packages/nc-gui/components/erd/FullScreenToggle.vue
@@ -1,0 +1,28 @@
+<script lang="ts" setup>
+import { Panel, PanelPosition } from '@vue-flow/additional-components'
+import type { ERDConfig } from './utils'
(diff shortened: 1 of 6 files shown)


packages/nc-gui/components/erd/FullScreenToggle.vue
```diff
@@ -0,0 +1,28 @@
...
+</script>
+
+<template>
+  <Panel
+    class="text-xs bg-white border-1 rounded-md p-0.5 border-gray-50 z-50 nc-erd-histogram cursor-pointer hover:bg-gray-100"
+    :position="PanelPosition.TopLeft"
+  >
+    <div class="flex">
+      <MiFullscreenExit v-if="config.isFullScreen" class="h-5 w-5" @click="() => toggleFullScreen()" />
+      <MiFullscreen v-else class="h-5 w-5" @click="() => toggleFullScreen()" />
```
wingkwong, line 25: can be further simplified as `@click="toggleFullScreen"`

packages/nc-gui/components/erd/FullScreenToggle.vue
```diff
@@ -0,0 +1,28 @@
...
+
+const props = defineProps<{
+  config: ERDConfig
+}>()
+
+const config = useVModel(props, 'config')
+
+const toggleFullScreen = () => {
+  config.value.isFullScreen = !config.value.isFullScreen
+}
```
pranavxc, line 15: Here it's mutating the component prop directly which is a bad practice. And `useVModel` is not necessary if 2-way binding is not used.

#### Suggestions

- You can use 2 way binding and do something as follows
    ```js
    config.value = { ...config.value, isFullScreen: !config.value.isFullScreen }
     ```
- Or you can emit an event and mutate the prop from the parent.

pranavxc (approved):

packages/nc-gui/components/erd/FullScreenToggle.vue
mustafapc19: @pranavxc  Taken the second approach

dstala (approved):

pranavxc (approved):

Issue #4617, opened 2022-12-06
🔦 Feature: UX improvement: Remove copy invite token button for registered users


vijayrathore8492:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

In the current situation when you click on the actions button in front of a user under the user management screen, It opens up 
<img width="251" alt="Screenshot 2022-12-06 at 18 17 39" src="https://user-images.githubusercontent.com/17380265/205916726-94cef736-85b0-4f60-a6bc-2c524960f8f9.png">
a menu with 2 options (Resend Invite E-mail and Copy Invite URL). 

These options are redundant for already registered users.

### Suggested Solution

 For better UX and to reduce user confusion these options should only be shown for users that are not yet registered successfully.

o1lab:
@vijayrathore8492 : can u please send a PR

wingkwong:
fixed and will be available in the [next release](https://github.com/nocodb/nocodb/milestone/22).

PR #4934, opened 2023-01-26
feat: UX improvement: Remove copy invite token button for register users

gitstart:
## Change Summary
ref: #4617
For better UX and to reduce user confusion these options should only be shown for users that are not yet registered successfully.

## Change type

https://www.loom.com/share/976667350a174255a9cf83303c859344


wingkwong (changes requested):
L112 - 116 is not necessary. You can just do in this way.

```vue
<a-dropdown v-if="user.invite_token" :trigger="['click']" class="flex" placement="bottomRight" overlay-class-name="nc-dropdown-user-mgmt">
```

gitstart:
> L112 - 116 is not necessary. You can just do in this way.
> 
> ```
> <a-dropdown v-if="user.invite_token" :trigger="['click']" class="flex" placement="bottomRight" overlay-class-name="nc-dropdown-user-mgmt">
> ```

Hi @wingkwong, 
The changes has been applied

wingkwong (approved):
Wait for the test result.

Issue #4615, opened 2022-12-06
🐛 Bug: Foreign Key name too long


folk-connect:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Foreign key name too long error. This happens while created LTAR column. Actually this is due to long name of table. The table name is less than max length but the foreign key column generated is too long
<img width="950" alt="Screenshot 2022-12-06 164128" src="https://user-images.githubusercontent.com/113366228/205896549-48b86ba1-e115-40d1-a315-4f2ad7a4390f.png">

### Desired Behavior

As a user, we don't have any control over this. This should be addressed by the application itself and name should be truncated or user should be allowed to edit that name

### Project Details

Node: **v14.19.0**
Arch: **x64**
Platform: **linux**
Docker: **false**
Database: **mysql2**
ProjectOnRootDB: **true**
RootDB: **mysql2**
PackageVersion: **0.100.2**

PR #5030, opened 2023-02-04
fix: Populate unique fk constraint name 

pranavxc:
## Change Summary

Populate a unique foreign key constraint name rather than relying on knex generated name.

re #4615

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/lib/meta/api/columnApis.ts
+++ b/packages/nocodb/src/lib/meta/api/columnApis.ts
@@ -51,7 +51,7 @@ export enum Altered {
 }
 
 // generate unique foreign key constraint name for foreign key
-const generateFkConstraintName = (parent: TableType, child: TableType) => {
+const generateFkName = (parent: TableType, child: TableType) => {
   // generate a unique constraint name by taking first 10 chars of parent and child table name (by replacing all non word chars with _)
   // and appending a random string of 15 chars maximum length.
   // In database constraint name can be upto 64 chars and here we are generating a name of maximum 40 chars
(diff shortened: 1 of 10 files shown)


packages/nocodb/src/lib/meta/api/columnApis.ts
```diff
@@ -316,6 +326,7 @@ export async function columnAdd(
                 onUpdate: 'NO ACTION',
                 type: 'real',
                 parentColumn: parent.primaryKey.column_name,
+                foreignKeyName: generateFkConstraintName(parent, child), //generateFkConstraintName()
```
wingkwong, line 329: `//generateFkConstraintName()` can be removed

packages/nocodb/src/lib/meta/api/columnApis.ts
```diff
@@ -50,6 +50,16 @@ export enum Altered {
...
+// generate unique foreign key constraint name for foreign key
+const generateFkConstraintName = (parent: TableType, child: TableType) => {
+  const constraintName = `fk_${parent.table_name
+    .replace(/\W+/g, '_')
+    .slice(0, 10)}_${child.table_name
+    .replace(/\W+/g, '_')
+    .slice(0, 10)}_${randomID(15)}`;
+  return constraintName;
+};
+
```
wingkwong, line 62: I'd suggest to add some comments to explain why we slice the table name as i think it's not quite intuitive at the first sight. For example, we may mention sth like in order to prevent FK key name length from exceeding the max length during creation.

wingkwong:
@pranavxc Seems it's failing `columnLinkToAnotherRecord.spec.ts`.

wingkwong (approved):

PR #5033, opened 2023-02-04
Allow uploads for public base url users with editor role

djmaze:
## Change Summary

Allows uploads for public base url users with editors. This was not working anymore in the latest versions (and I am not sure it was ever implemented correctly).

This fix is probably not 100% correct, but I need a working docker image for my quickfix, but don't I am not able to create the image manully successfully, so I created this PR.

## Change type

## Test/ Verification

* Share a project via the public base url with editor role.
* Try to upload a file on an attachment field when using the base url.

### DIFF
--- a/packages/nocodb/src/lib/meta/api/attachmentApis.ts
+++ b/packages/nocodb/src/lib/meta/api/attachmentApis.ts
@@ -23,20 +23,26 @@ const isUploadAllowed = async (req: Request, _res: Response, next: any) => {
   try {
     // check user is super admin or creator
     if (
-      req['user'].roles?.includes(OrgUserRoles.SUPER_ADMIN) ||
-      req['user'].roles?.includes(OrgUserRoles.CREATOR) ||
-      (req['user'].isPublicBase && req['user'].roles?.includes(ProjectRoles.EDITOR)) ||
-      // if viewer then check at-least one project have editor or higher role
-      // todo: cache
-      !!(await Noco.ncMeta
-        .knex(MetaTable.PROJECT_USERS)
-        .where(function () {
-          this.where('roles', ProjectRoles.OWNER);
-          this.orWhere('roles', ProjectRoles.CREATOR);
-          this.orWhere('roles', ProjectRoles.EDITOR);
-        })
-        .andWhere('fk_user_id', req['user'].id)
-        .first())
+      req['user'].id &&
+      (req['user'].roles?.includes(OrgUserRoles.SUPER_ADMIN) ||
+        req['user'].roles?.includes(OrgUserRoles.CREATOR) ||
+        // if viewer then check at-least one project have editor or higher role
+        // todo: cache
+        !!(await Noco.ncMeta
+          .knex(MetaTable.PROJECT_USERS)
+          .where(function () {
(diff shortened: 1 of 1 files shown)


o1lab:
@djmaze : thanks for the PR. Please sign the CLA above.

wingkwong (changes requested):
If I just use the API without passing auth token, I'd get `Bad Request` rather than `Unauthorized`. I'd suggest to check if `id` and `isPublicBase` in the first if-statement first.

packages/nocodb/src/lib/meta/api/attachmentApis.ts
```diff
@@ -16,26 +16,33 @@ import Local from '../../v1-legacy/plugins/adapters/storage/Local';
 import { NC_ATTACHMENT_FIELD_SIZE } from '../../constants';
 
 const isUploadAllowed = async (req: Request, _res: Response, next: any) => {
-  if (!req['user']?.id) {
+  if (!req['user']) {
```
wingkwong, line 19: The test case is still failing. I believe you can simply do in this way.

```ts
  if (
    !req['user']?.id ||
    (req['user']?.isPublicBase &&
      !req['user'].roles?.includes(ProjectRoles.EDITOR))
  ) {
    NcError.unauthorized('Unauthorized');
  }

```

packages/nocodb/src/lib/meta/api/attachmentApis.ts
```diff
@@ -16,26 +16,33 @@ import Local from '../../v1-legacy/plugins/adapters/storage/Local';
...
+            this.orWhere('roles', ProjectRoles.EDITOR);
+          })
+          .andWhere('fk_user_id', req['user'].id)
+          .first()))
+    )
+      return next();
+    // check public base url with editor role
+    else if (
+      req['user'].isPublicBase &&
+      req['user'].roles?.includes(ProjectRoles.EDITOR)
```
wingkwong, line 45: these can be removed

packages/nocodb/src/lib/meta/api/attachmentApis.ts
djmaze: Not quite.. Then it fails if *either* the id is missing or the public url & role. But the success case is that *both* the id is missing as well as the public url & role is are there.

djmaze:
It would really help if someone could tell me how to run *single tests* in the development environment. (I couldn't easily find out, but am not a mocha / JS expert). Running all tests takes *ages* on my machine, so I switched to just pushing stuff here first.

packages/nocodb/src/lib/meta/api/attachmentApis.ts
wingkwong: Lemme push something to test.

wingkwong:
> It would really help if someone could tell me how to run _single tests_ in the development environment. (I couldn't easily find out, but am not a mocha / JS expert). Running all tests takes _ages_ on my machine, so I switched to just pushing stuff here first.

You can simply add `.only` to the test case(s) that you want to run.

ref: https://docs.nocodb.com/engineering/unit-testing#walk-through-of-writing-a-unit-test

image

djmaze:
@wingkwong Thanks!

PR #5042, opened 2023-02-06
refactor(nocodb): move cdn assets to public directory

wingkwong:
## Change Summary

Some pages like swagger will be broken if users host the app in a private network.

- replace hardcoded base64 string with favicon.ico
- add and replace by following assets
  - swagger-ui-bundle.4.5.2.min.js
  - axios.0.19.2.min.js
  - vue.2.6.14.min.js  
  - vuetify.2.x.min.js
  - redoc.standalone.min.js
  - vue.global.js
  - swagger-ui-bundle.4.5.2.min.css
  - fonts.roboto.css
  - materialdesignicons.5.x.min.css
  - vuetify.2.x.min.css

## Change type


packages/nocodb/src/lib/meta/api/swagger/redocHtml.ts
```diff
@@ -5,7 +5,7 @@ export default `<!DOCTYPE html>
     <!-- needed for adaptive design -->
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link href="https://fonts.googleapis.com/css?family=Montserrat:300,400,700|Roboto:300,400,700" rel="stylesheet">
+    <link href="/css/fonts.montserrat.css" rel="stylesheet">
```
pranavxc, line 8: what if nocodb is serving under a subpath (for eg:  using some proxy )? Since providing `/` in the beginning it will look under the root url.

pranavxc (commented):
The `/public` folder won't be part of the production build, have to add a step to copy the static files.

packages/nocodb/src/lib/meta/api/swagger/swaggerApis.ts
```diff
@@ -38,18 +38,24 @@ async function swaggerJson(req, res) {
   res.json(swagger);
 }
 
+function swaggerHtml(req, res) {
+  res.send(getSwaggerHtml(req.ncSiteUrl));
```
pranavxc, line 42: you can use `ncPublicUrl: process.env.NC_PUBLIC_URL || ''` here as well

pranavxc (approved):

pranavxc (approved):

PR #5043, opened 2023-02-06
fix(nocodb): prevent from hashing a null salt

wingkwong:
## Change Summary

- If A invites B, and B sign in without setting the password (using invite token), then it would show the reported error due to null salt. 
- This PR is to check if `user.salt` is null and remind them to sign up with the invite token to set the password first.

## Change type

### DIFF
--- a/packages/nocodb/src/lib/meta/api/userApi/initStrategies.ts
+++ b/packages/nocodb/src/lib/meta/api/userApi/initStrategies.ts
@@ -207,6 +207,13 @@ export function initStrategies(router): void {
           if (!user) {
             return done({ msg: `Email ${email} is not registered!` });
           }
+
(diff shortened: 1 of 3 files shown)


packages/nocodb/src/lib/meta/api/userApi/initStrategies.ts
```diff
@@ -207,6 +207,13 @@ export function initStrategies(router): void {
           if (!user) {
             return done({ msg: `Email ${email} is not registered!` });
           }
+
+          if (!user.salt) {
+            return done({
+              msg: `Please sign up with the invite token first!`,
```
pranavxc, line 213: In message, you can include `or use forgot password`

Issue #4992, opened 2023-02-01
🐛 Bug: Performance degradation when upgrading to 0.104.2


davidpfahler:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

I just upgraded our production environment from 0.101.0-beta.0 to 0.104.2. Switching from one table to another is noticeably slower which I suspect is due to the new row height feature requiring a lot more work on the client.

### Desired Behavior

Opening a table should display the data as fast as possible.

### Project Details

Node: **v16.17.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.104.2**

o1lab:
Thanks for the reporting.

@mertmit - please have a look

pavelkalin:
We also see this problem in our development

distresscode:
My instance is running slow too after the recent updates. I'm using Noco on Cloudron.

Node: **v18.12.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.104.3**

mertmit:
Thank you for reporting this issue.
If possible can you try PR build and feedback if this is resolved.
```docker run -d -p 8888:8080 nocodb/nocodb-timely:0.104.3-pr-5015-20230203-0852```

davidpfahler:
@mertmit Unfortunately, I cannot notice an improvement in performance. It actually appear to me that there is further drastic performance degradation going from `0.104.2` to `0.104.3`, but that is just a feeling. Data not not only seems to need long to render on the client side, but it actually take very long for any data whatsoever to arrive.

Just looking at the network tab, it seems like for some boards nocodb is making dozen even hundreds of requests when the board contains a file column and **is downloading all the files attached to the rows every time**. Boards without a file column seem ok.

davidpfahler:
@mertmit @o1lab You need to see this, this is pretty broken on `0.104.3`. This is the network tab when I open our invoices table, which has an attachments column that has one PDF per row:

https://user-images.githubusercontent.com/908242/217005100-11ae789d-cc40-4448-9122-5e46eb09bed2.mov

davidpfahler:
Yes, both this build and latest

o1lab:
We are looking into it & will keep it posted.

mertmit:
[...]

PR #5046, opened 2023-02-06
fix: drop url availability check with fetch

mertmit:
## Change Summary

Re #4992

wkw:
- remove the logic enriching `url` by fetching attachment one by one
- remove displaying image using `background-image`. now it's using NuxtImg.
- by default, it uses `${appInfo.value.ncSiteUrl}/${item.path}` with fallback source `${item.url}`. If `${item.url}` also fails, then a static 404 not found image will be used to display (to avoid infinite loop)

## Change type


packages/nc-gui/composables/useAttachment.ts
```diff
@@ -0,0 +1,35 @@
+import fileNotFoundImgSrc from '~/assets/img/file-not-found.png'
+
```
pranavxc, line 2: To avoid breaking the asset path in the production build we can do it at the template level by creating a new component for rendering images.

Something like the following : 


In the main component use it like this: 

```vue

<LazyAttachmentImage :srcs="srcs" />

```
The `srcs` is an array of URLs populated using `item.url` and `item.path`

AttachmentImage component:

```vue
<script setup lang="ts">
  const props = defineProps<{srcs: string[]}>

  const index = ref(0)

  const onError () => index.value++
</script>
<template>
   <img v-if="index < props.srcs.length" :src="props.srcs[index]" @onerror="onError"/>
   <img v-else src="~/assets/img/file-not-found.png" />
</template>
```

packages/nc-gui/components/cell/attachment/Image.vue
```diff
@@ -0,0 +1,25 @@
...
+
+const onError = () => index.value++
+</script>
+
+<template>
+  <LazyNuxtImg
+    v-if="index < props.src.length"
+    class="m-auto"
+    :src="props.src[index]"
+    :onerror="onError"
```
pranavxc, line 19: You can replace it with `@error="onError"`

wingkwong: Thanks. Updated.

packages/nc-gui/components/cell/attachment/Image.vue
```diff
@@ -0,0 +1,25 @@
+<script setup lang="ts">
+interface Props {
+  src: string[]
```
pranavxc, line 3: You can name it `srcs` since it's going to be array of string

PR #5060, opened 2023-02-08
feat: sticky primary column

### LINKED ISSUE #3583: Make the primary column sticky
- [ ] Each table to have one primary field.
- [ ] It will be sticky when horizontal scroll is performed.
- [ ] User shouldn't be able to hide it in fields.
- [ ] On delete of primary field, we should automatically select as primary field.
- [ ] #5007

mertmit:
## Change Summary

fixes #3583
- First column in grid is sticky
- Set primary value column as first column (disable hide / reorder options for it)
- Upgrader: Make primary value column the first column for each grid-view

## Change type

WIP: tests + test fixes


wingkwong (commented):
1.`title8` is PV. However, I could add a column `title9` before that. In fields, `title9` is considered as a PV while it is actually not.

image

image

2. Unexpected new PV 

- `Title` is the PV
- Delete `Title`
- `CreatedAt` suddenly shows up and becomes the new PV while `title5` should be the one. 

image

image

dstala:
- [ ] ID field (system field) listed in filter & sort menu when `show system fields` is disabled
  
Screenshot 2023-02-09 at 10 33 20 AM

- [ ] Disable `insert before` for sticky column 
  
Screenshot 2023-02-09 at 10 53 26 AM

- [ ] Disable `hide field` for sticky column; action can't be undone
  
Screenshot 2023-02-09 at 10 55 15 AM

- [ ] Disable `delete field` for sticky column; simplifies the way we handle primary value

dstala (changes requested):
https://github.com/nocodb/nocodb/pull/5060#issuecomment-1423659411

dstala (approved):

wingkwong (commented):
PV should be on the top under `fields` after setting a new one.

image

wingkwong (approved):

packages/nocodb/src/lib/version-upgrader/ncStickyColumnUpgrader.ts
```diff
@@ -0,0 +1,72 @@
...
+          null,
+          MetaTable.GRID_VIEW_COLUMNS,
+          { show: true },
+          primary_value_column.id,
+        );
+      }
+
+      if (primary_value_column.order ===  view_min_order && view_orders.filter((o) => o === view_min_order).length === 1) {
+        // if primary_value_column is in first order do nothing
+        return;
```
pranavxc, line 51: `return` or `continue` ?

mertmit: it should be continue as we need to apply this upgrader to each view, fixed it, thanks

pranavxc (approved):

Issue #5062, opened 2023-02-08
[MS SQL] 🐛 Bug: Kanban board view not rendering


andrzejhoc:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

As a backend database I'm using MS Sql and create custom column to use on kanban board as grouping field. The created column requirement is to be of text or ntext type. But when kanban board is rendering there is error comparing this column type to varchar. When I change directly on database column type from text to varchar than everything works ok. 

### Desired Behavior

I think there should be option to chose additionaly varchar and nvarchar column type. Another great option is to allow define as a kanban group field dictionary table

### Project Details

Node: v16.16.0
Arch: x64
Platform: win32
Docker: false
RootDB: sqlite3
PackageVersion: 0.104.3

PR #5067, opened 2023-02-09
fix: use nvarchar as default single/multi select datatype in mssql

pranavxc:
## Change Summary

Use nvarchar/varchar as the default single select datatype since the text field doesn't support grouping in MSSQL.

re #5062

## Change type

## Test/ Verification

Provide summary of changes.

## TODO 

- [ ] Handle `dtxp`

### DIFF
--- a/packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
+++ b/packages/nocodb-sdk/src/lib/sqlUi/MssqlUi.ts
@@ -971,9 +971,11 @@ export class MssqlUi {
         break;
       case 'MultiSelect':
         colProp.dt = 'nvarchar';
+        colProp.dtxp = 4000;
         break;
       case 'SingleSelect':
         colProp.dt = 'nvarchar';
+        colProp.dtxp = 4000;
         break;
       case 'Collaborator':
         colProp.dt = 'varchar';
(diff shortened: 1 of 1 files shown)


mertmit:
Once we brainstormed with @wingkwong on handling dtxp, there was an idea I don't exactly recall why we dropped but noting it down for reference:
Each time we add/drop/modify an option we can calculate maximum possible length for that column and modify the dtxp according to this.
- For Single Select we can make it length of longest option
- For Multi Select we can calculate sum of option lengths + option count (-1 for commas)
This way we can optimize data sizes for columns.

Concerns:
- MySQL should be handled separately as we use enum/set as default for it
- Keeping data as text might not be optimal way to go, even if we set size for each column filter and sort will be costly

### Alternative to text/varchar for xcdb
if we drop order (just like MySQL enum/set) we can use keeping meta to our advantage by using bitwise operations and keeping a number for each record. This way we can filter / sort using bit filters. If we implement this in a good way we can even filter by partial string. (first filter options with set filter then find corresponding number to run the filter and query it)
Downside of this data won't be human friendly (shouldn't be a problem for xcdb) and we should handle our responses (each api returning a select column should first replaced by corresponding options)

PR #5114, opened 2023-02-16
Mvp mobile layout and code scanner

spaudanjo:
## Change Summary

See the screen recording video for an overview of added features. 
Everything is feature toggle hidden so far. 

https://user-images.githubusercontent.com/249210/225698315-b85a6d6a-c693-4be5-b006-15039d07c635.mov

Introduces a:
* Mobile Layout which can activated via the main menus new entry `Toggle Mobile Mode`
  * this hides several buttons and links which are considered as not relevant when interacting with a non-desktop device (e.g. API snippets) and reduces the buttons to only the icons (no more text)

* a QR- and Barcode-scanner button which can be used to find a single row of a table via scanning a code (after selecting by which column to search). Atm, it explicitly only forks for unique matches (if there are more than one rows matching the scanned code and selected column, the user is informed that this is not supported atm; that's because the system wouldn't know for which of the matched rows the expanded view should be opened
  * in relation to to the QR scanner, it also 
    * sets explicit preferences for the QR code generator; during the development of the scanner integration, I realised that QR codes especially for very short encoded strings are not reliably recognised by scanners
    * enables the `ID` column type to be used for QR codes

* allow form editors to enable the scanner also for specific form fields

## Change type

## Test/ Verification

Provide summary of changes.


dstala (changes requested):
few observations
https://www.loom.com/share/c2c44e403a3443ce9530b016f725e738

spaudanjo:
@dstala as discussed, I changed the PR since the state management related flaws were not so easy to spot for now. The Views sidebar stays on the right also in mobile mode. Overall this would be a first step to make the app usable on small screens, but there is more room for next steps after this got merged. 

Please let me know whether you agree both with the code as well as the UX.

packages/nc-gui/pages/[projectType]/[projectId]/index/index.vue
```diff
@@ -26,6 +28,16 @@ const { isOpen, toggle } = useSidebar('nc-left-sidebar')
...
+
+const hideSidebarOnClickOrTouchIfMobileMode = () => {
+  if (isMobileMode.value && isOpen.value) {
+    toggle(false)
+  }
+}
+
+onMounted(() => {
+  mainArea.value?.addEventListener('click', hideSidebarOnClickOrTouchIfMobileMode)
+})
```
Comment (pranavxc) [inline packages/nc-gui/pages/[projectType]/[projectId]/index/index.vue:40]: Any reason to not use `@click` at the div to listen click event.

```
<div ref="mainArea" class="w-full min-h-[300px] flex-auto" @click="hideSidebarOnClickOrTouchIfMobileMode">
```

dstala (approved):

packages/nc-gui/pages/[projectType]/[projectId]/index/index.vue
Comment (spaudanjo) [reply packages/nc-gui/pages/[projectType]/[projectId]/index/index.vue]: Very good question 😂 
I changed it as suggested

wingkwong (commented):
1. With toggle on, clicking these buttons (or anywhere) would close the left-hand side navigation. Is it by design?
image

2. May include more info here. I couldn't get what this feature does purely looking at the UI. 
image
3. badge is cropped
image
4. Can you provide a screenshot for below statement? 
> also, the Views sidebar is moved from the right to the left side and covers the Table sidebar to save mobile screen space

packages/nc-gui/components/smartsheet/toolbar/QrScannerButton.vue
```diff
@@ -0,0 +1,162 @@
+<script setup lang="ts">
+import type { SelectProps } from 'ant-design-vue'
+import { ref } from 'vue'
+import { StreamBarcodeReader } from 'vue-barcode-reader'
+import type { ColumnType } from 'nocodb-sdk'
+import { UITypes } from 'nocodb-sdk'
+import { NOCO } from '#imports'
+import QrCodeScan from '~icons/mdi/qrcode-scan'
```
wingkwong, line 8: remove this line. auto-imported in `components.d.ts` already.

packages/nc-gui/components/smartsheet/toolbar/QrScannerButton.vue
```diff
@@ -0,0 +1,162 @@
...
+    console.error(error)
+  }
+}
+</script>
+
+<template>
+  <div>
+    <a-button class="nc-btn-find-row-by-scan nc-toolbar-btn" @click="showCodeScannerOverlay = true">
+      <div class="flex items-center gap-1">
+        <QrCodeScan />
```
wingkwong, line 126: Use `<MdiQrcodeScan />`

packages/nc-gui/composables/useViews.ts
```diff
@@ -2,7 +2,7 @@ import type { TableType, ViewType } from 'nocodb-sdk'
 import type { MaybeRef } from '@vueuse/core'
 import { ref, unref, useNuxtApp, watch } from '#imports'
 
-export function useViews(meta: MaybeRef<TableType | undefined>) {
+function _useViews(meta: MaybeRef<TableType | undefined>) {
```
wingkwong, line 5: why there is such change?

packages/nocodb/src/lib/migrations/v2/nc_026_add_enable_scanner_in_form_columns_meta_table.ts
```diff
@@ -0,0 +1,16 @@
+import { Knex } from 'knex';
```
wingkwong, line 1: `import type { Knex } from 'knex';`

packages/nc-gui/composables/useSharedFormViewStore.ts
```diff
@@ -37,7 +37,7 @@ const [useProvideSharedFormStore, useSharedFormStore] = useInjectionState((share
   const sharedView = ref<ViewType>()
   const sharedFormView = ref<FormType>()
   const meta = ref<TableType>()
-  const columns = ref<(ColumnType & { required?: boolean; show?: boolean; label?: string })[]>()
+  const columns = ref<(ColumnType & { required?: boolean; show?: boolean; label?: string; enable_scanner?: boolean })[]>()
```
wingkwong, line 40: use `BoolType` instead of `boolean`

packages/nc-gui/components/smartsheet/toolbar/QrScannerButton.vue
```diff
@@ -0,0 +1,162 @@
...
+import QrCodeScan from '~icons/mdi/qrcode-scan'
+
+const meta = inject(MetaInj, ref())
+
+const { t } = useI18n()
+const route = useRoute()
+const router = useRouter()
+
+const { $api } = useNuxtApp()
+const { project } = useProject()
```
wingkwong, line 17: since we've introduced pinia, this line should be

```vue
const { project } = storeToRefs(useProject())
```

you can import `storeToRefs`  from `'#imports`

packages/nc-gui/components/smartsheet/toolbar/QrScannerButton.vue
```diff
@@ -0,0 +1,162 @@
...
+import type { ColumnType } from 'nocodb-sdk'
+import { UITypes } from 'nocodb-sdk'
+import { NOCO } from '#imports'
+import QrCodeScan from '~icons/mdi/qrcode-scan'
+
+const meta = inject(MetaInj, ref())
+
+const { t } = useI18n()
+const route = useRoute()
+const router = useRouter()
```
wingkwong, line 14: prefer line break between each statement

spaudanjo:
@wingkwong 

> With toggle on, clicking these buttons (or anywhere) would close the left-hand side navigation. Is it by design?

Yes, this is by design having mobile devices in mind (since extra touches/clicks are more 'expensive'/annoying there and one cannot really work anyway with the main section if the sidebar is open - plus there is often also the right sidebar).

packages/nc-gui/composables/useSharedFormViewStore.ts
spaudanjo: Done (also changed the other types to the new wrapper types - that's ok?

packages/nc-gui/composables/useViews.ts
spaudanjo: removed - was a left-over from a former approach and discussion where there plan was to move the right side bar also to the left when in mobile mode and afai remember we had to share the same useView composable/state across siblings in the component tree then.

packages/nc-gui/composables/useSharedFormViewStore.ts
spaudanjo: they are indeed

```
          "label": {
            "$ref": "#/components/schemas/StringOrNull",
            "description": "Form Column Label"
          },
          "required": {
            "$ref": "#/components/schemas/Bool",
            "description": "Is this form column required in submission?"
          },
          "show": {
            "$ref": "#/components/schemas/Bool",
            "description": "Is this column shown in Form?"
          },
          "enable_scanner": {
            "$ref": "#/components/schemas/Bool",
            "description": "Indicates whether the 'Fill by scan' button is visible for this column or not.",
            "example": true
          },
```

wingkwong (approved):

pranavxc (approved):

PR #5134, opened 2023-02-17
fix: ts types for sql clients

mertmit:
## Change Summary

- Correct types instead of any
- fix broken clientType calls

## Change type


wingkwong (approved):
Verified `sqlClientType` in my another PR.

PR #5137, opened 2023-02-17
feat: implement NC_DISABLE_AUDIT

### LINKED ISSUE #4643: 🔦 Feature: Make stable bound to the audit log
### Please confirm if feature request does NOT exists already ?

### Describe the usecase for the feature

Right now it seems the audit log that noco maintains in it's sqlite db isn't bounded.  I'm running an automated application that uses an API token to make frequent updates.  Although I'm not adding much data, Noco is adding growing about 10MB/hour.  I'd like to be able to opt out of this audit log or set a bound to stop it from growing.

Shown is graph of disk used for Noco and disk used for the postgres database it's connected to.  Already noco has ratio'd postgres :)

### Suggested Solution

- Allow configuration to opt out of this audit log or set a bound to stop it from growing.

### Additional Context

Screen Shot 2022-12-08 at 8 36 34 PM

wingkwong:
## Change Summary

closes #4643

- introduce `NC_DISABLE_AUDIT` (by default: `false`)
- Show the message `Audit logs are currently disabled by administrators.` in Audit tab if it is disabled.
- Prevent from inserting audit logs if the see

## Change type

## Test/ Verification

image


packages/nocodb/src/lib/models/Audit.ts
```diff
@@ -44,6 +44,9 @@ export default class Audit implements AuditType {
       forceAwait: process.env['TEST'] === 'true',
     }
   ) {
+    if (process.env['NC_DISABLE_AUDIT'] ?? false) {
+      return;
+    }
```
pranavxc, line 49: Since here you are just checking env value is trruthy do the same in appInfo API as well.

`process.env.NC_DISABLE_AUDIT === 'true' ? false : true,` => `!process.env.NC_DISABLE_AUDIT,`

wingkwong: `process.env.NC_DISABLE_AUDIT` is a string. `!"false" -> false` and `!"true" -> false`.

dstala (changes requested):
Scenario
- Execute with `NC_DISABLE_AUDIT=true npm run watch:run`
- And then with `npm run watch:run`

I was expecting logging to be enabled back when NC_DISABLE_AUDIT wasn't defined. But that didn't happen. Was this by design? I would prefer considering not defined as equivalent to setting that variable to `false`

dstala:
> Weird. I couldn't reproduce it. By default, audit logs are enabled if `NC_DISABLE_AUDIT` is not defined. so it should be same as `NC_DISABLE_AUDIT=false`.
> 
> image

Hard refresh was required. Will it be possible to refresh it automatically on re-opening of settings/audit tab?

wingkwong:
That would be stored in the state. I think you gotta refresh to trigger util api to get the updated value.

dstala (approved):

Issue #5122, opened 2023-02-16
🐛 Bug: sticky column field can be dragged & realigned in sql views


dstala:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

in sql views, display value field can be realigned by drag-drop

### Desired Behavior

similar to table views

### Project Details

Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
RootDB: **pg**
PackageVersion: **0.104.3**

3dprogramin:
I can confirm from my end too, but version is **0.105.0** on my end.

Node: v16.18.1
Arch: amd64
Platform: linux
Docker: false
RootDB: pg
PackageVersion: 0.105.0

mertmit:
Confirmed, this is because we don't assign display value for sql views, I will try to cover this case with a PR

PR #5141, opened 2023-02-17
fix: missing display value

mertmit:
## Change Summary

Re #5122 
- assign first column as display value if no pk
- assign pk as display value if no other columns
- handle missing display value in ncStickyUpgrader
- drop possible duplicated display values when display value updated

## Change type

### DIFF
--- a/packages/nocodb/src/lib/version-upgrader/ncStickyColumnUpgrader.ts
+++ b/packages/nocodb/src/lib/version-upgrader/ncStickyColumnUpgrader.ts
@@ -38,6 +38,50 @@ export default async function ({ ncMeta }: NcUpgraderCtx) {
       view_columns_meta.push(col_meta);
     }
 
+    // if no display value column is set
+    if (!view_columns_meta.some((column) => column.pv)) {
+      const pkIndex = view_columns_meta.findIndex((column) => column.pk)
+  
+      // if PK is at the end of table
+      if (pkIndex === view_columns_meta.length - 1) {
+        if (pkIndex > 0) {
+          await ncMeta.metaUpdate(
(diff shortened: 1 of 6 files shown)


packages/nocodb/src/lib/version-upgrader/ncStickyColumnUpgrader.ts
```diff
@@ -38,6 +38,50 @@ export default async function ({ ncMeta }: NcUpgraderCtx) {
...
+            null,
+            MetaTable.COLUMNS,
+            { pv: true },
+            view_columns_meta[pkIndex - 1].id
+          );
+        } else if (view_columns_meta.length > 0) {
+          await ncMeta.metaUpdate(
+            null,
+            null,
+            MetaTable.COLUMNS,
```
pranavxc, line 59: If I'm not wrong it needs to be in a new upgrader file otherwise, it won't run on the user's machine who had already started using the latest version.

wingkwong: agree

mertmit: I proposed changing upgrader version to latest instead of creating a new upgrade file to keep code clean as this upgrader doesn't have any side effects, waiting for pranav to confirm if it is appropriate.

pranavxc: @mertmit : that sounds good, we can do it in that way :+1:

pranavxc (approved):

Issue #5147, opened 2023-02-17
Migration failed :  0100002 to 0104004


benintende:
Hi everyone,

after updating nocodb today I get the following error message in Terminal, and I can't open nocodb anymore:

(node:1732) NOTE: The AWS SDK for JavaScript (v2) will be put into maintenance mode in 2023.

Please migrate your code to use AWS SDK for JavaScript (v3).
For more information, check the migration guide at https://a.co/7PzMCcy
(Use `nocodb --trace-warnings ...` to show where the warning was created)
┌ Migration from 0100002 to 0104004 failed ────────────────────────────────────┐│                                                                              ││   Error                                                                      ││   -----                                                                      ││   Error: connect EHOSTDOWN 192.168.1.68:3306 - Local (192.168.1.7:50512)     ││   at internalConnect (node:net:953:16)                                       ││   at defaultTriggerAsyncIdScope (node:internal/async_hooks:465:18)           ││   at node:net:1044:9                                                         ││   at processTicksAndRejections (node:internal/process/task_queues:78:11)     ││                                                                              ││                                                                              ││   Please raise an issue in our github by using following link :              ││   https://github.com/nocodb/nocodb/issues/new?labels=Type%3A%20Bug&templat   ││   e=bug_report.md                                                            ││                                                                              ││   Or contact us in our Discord community by following link :                 ││   https://discord.gg/5RgZmkW ( message @o1lab, @pranavxc or @wingkwong )     ││                                                                              │└──────────────────────────────────────────────────────────────────────────────┘
Error: connect EHOSTDOWN 192.168.1.68:3306 - Local (192.168.1.7:50512)
    at internalConnect (node:net:953:16)
    at defaultTriggerAsyncIdScope (node:internal/async_hooks:465:18)
    at node:net:1044:9
    at processTicksAndRejections (node:internal/process/task_queues:78:11) {
  errno: -64,
  code: 'EHOSTDOWN',
  syscall: 'connect',
  address: '192.168.1.68',
  port: 3306,
  fatal: true
}

Any ideas/suggestions?

o1lab:
This has nothing to do with M1.

Which build were you trying - looks like schema migrations have failed. Provide more details on environment variables

benintende:
Hello, thanks for your quick answer! Not sure which build it was, I updated to the latest version available today via curl http://get.nocodb.com/macos-x64 -o nocodb -L && chmod +x nocodb && ./nocodb

benintende:
[...]

PR #5154, opened 2023-02-18
feat: upgrade timeout msg

wingkwong:
## Change Summary

- revise upgrade error message for timeout case

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/lib/version-upgrader/ncAttachmentUpgrader_0104002.ts
+++ b/packages/nocodb/src/lib/version-upgrader/ncAttachmentUpgrader_0104002.ts
@@ -156,11 +156,11 @@ export default async function ({ ncMeta }: NcUpgraderCtx) {
         }
         await Promise.all(updateRecords);
       } catch (e) {
-        // throw the custom timeout error message if applicable
-        throwTimeoutError(e, timeoutErrorInfo);
-
         // ignore the error related to deleted project
         if (!isProjectDeleted) {
+          // throw the custom timeout error message if applicable
+          throwTimeoutError(e, timeoutErrorInfo);
+          // throw general error
           throw e;
         }
       }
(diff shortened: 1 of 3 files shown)


packages/nocodb/src/lib/version-upgrader/ncAttachmentUpgrader_0104002.ts
```diff
@@ -149,6 +156,9 @@ export default async function ({ ncMeta }: NcUpgraderCtx) {
         }
         await Promise.all(updateRecords);
       } catch (e) {
+        // throw the custom timeout error message if applicable
+        throwTimeoutError(e, timeoutErrorInfo);
+
```
pranavxc, line 161: You can move this to the if condition, for deleted projects we don't have to throw the error

wingkwong: The if-condition is inside `throwTimeoutError`.  Since this function could be used in different upgraders, I try to remove the duplicate code as much as possible. If the error code is not `ETIMEDOUT` (e.g. deleted project case), basically it won't throw anything in `throwTimeoutError`.

packages/nocodb/src/lib/version-upgrader/ncUpgradeErrors.ts
```diff
@@ -0,0 +1,19 @@
+export function throwTimeoutError(e, timeoutErrorInfo) {
+  if (e.code === 'ETIMEDOUT') {
```
pranavxc, line 2: For different database client, the error code may be something else and it can be `EHOSTDOWN` as well(https://github.com/nocodb/nocodb/issues/5147)

pranavxc (approved):

Issue #5103, opened 2023-02-15
🐛 Bug: Wrong baseId is used in QuickImport, when a secondary database is only active. So the Table is created in the hidden DB.


semmjon:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Setup secondary catabase connection
![image](https://user-images.githubusercontent.com/42977205/218985622-98fc3fc8-d57f-4430-9616-014c59108592.png)

2. Deactivate base connection
![image](https://user-images.githubusercontent.com/42977205/218985808-fac9c778-64b3-47ca-a109-3953e3989145.png)

3. Import a Table from csv, excel, json with the Drag / Load Dialog
![image](https://user-images.githubusercontent.com/42977205/218985866-5c945f27-5153-429b-9427-b1d07f5aa7f6.png)
![image](https://user-images.githubusercontent.com/42977205/218985927-db5dd272-1252-42e1-9367-6f86e087ed8d.png)

### Desired Behavior

The Table will be created in the enabled Database not the hidden once.

### Project Details

```
Docker used: false
NocoDB Version: develop
NC_DB: NC_DB="mysql2://<ip>:3306?u=<user>&p=<password>&d=nocodb"
```

### Attachments

Possible Fix:

1. 
   File: https://github.com/nocodb/nocodb/blob/develop/packages/nc-gui/composables/useProject.ts
   Change:
   ![image](https://user-images.githubusercontent.com/42977205/218988734-771d6b74-4533-4373-a52a-bef198e259fe.png)
2. File: https://github.com/nocodb/nocodb/blob/develop/packages/nc-gui/pages/[projectType]/[projectId]/index/index/index.vue
    Change:
    ![image](https://user-images.githubusercontent.com/42977205/218989535-22ddda09-38c7-4975-b941-53af8f4e9cb6.png)

> :info: **Hurray!**: For me it will fix also this issue (that was still an issue): https://github.com/nocodb/nocodb/pull/4633

dstala:
@semmjon can you create a PR?

semmjon:
I'm partly working on other features or still have custom adjustments. I'll adjust my fork repository again and create a branch to do PR's so I don't accidentally merge anything wrong.

PR #5159, opened 2023-02-19
fix: wrong baseid in QuickImport

semmjon:
## Change Summary

PR for:
- closes https://github.com/nocodb/nocodb/issues/5103

## Change type

- [ x ] fix: (bug fix for the user, not a fix to a build script)

## Test/ Verification

1. The useProject.ts file in the nc-gui package:
A filter has been added to the bases computed property to only include those that have the enabled property set to true.

2. The index.vue file in the nc-gui package:
The baseId property has been added to the openQuickImportDialog function, and it is set to the id of the first base in the bases array that has the enabled property set to true.
In summary, these changes filter out disabled bases in the bases computed property, and pass the id of the first enabled base to the openQuickImportDialog function as the baseId property.

You can test it like in the issue description

### DIFF
--- a/packages/nc-gui/composables/useProject.ts
+++ b/packages/nc-gui/composables/useProject.ts
@@ -35,7 +35,7 @@ export const useProject = createSharedComposable(() => {
 
   const project = ref<ProjectType>({})
 
-  const bases = computed<BaseType[]>(() => project.value?.bases || [])
+  const bases = computed<BaseType[]>(() => project.value?.bases?.filter((base: BaseType) => base.enabled) || [])
 
   const tables = ref<TableType[]>([])
 
--- a/packages/nc-gui/composables/useProject.ts
+++ b/packages/nc-gui/composables/useProject.ts
@@ -35,7 +35,7 @@ export const useProject = createSharedComposable(() => {
 
   const project = ref<ProjectType>({})
 
-  const bases = computed<BaseType[]>(() => project.value?.bases?.filter((base: BaseType) => base.enabled) || [])
+  const bases = computed<BaseType[]>(() => project.value?.bases || [])
 
   const tables = ref<TableType[]>([])
 
--- a/packages/nc-gui/pages/[projectType]/[projectId]/index/index/index.vue
+++ b/packages/nc-gui/pages/[projectType]/[projectId]/index/index/index.vue
@@ -12,7 +12,6 @@ import {
   useUIPermission,
   watch,
 } from '#imports'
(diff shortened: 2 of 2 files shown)


packages/nc-gui/composables/useProject.ts
```diff
@@ -35,7 +35,7 @@ export const useProject = createSharedComposable(() => {
 
   const project = ref<ProjectType>({})
 
-  const bases = computed<BaseType[]>(() => project.value?.bases || [])
+  const bases = computed<BaseType[]>(() => project.value?.bases?.filter((base: BaseType) => base.enabled) || [])
```
wingkwong, line 38: By adding this change, it will break the share form views. You can reproduce the issue by creating a form view and then share it. The shared form view will be not shown and you can notice the errors from console.

The reason is that we haven't passed `enabled` in `useSharedFormViewStore`, hence `bases` would be empty (with your new change) causing undefined `sqlUis`.

https://github.com/nocodb/nocodb/blob/ea24fa91211f2bb1c6512504f5289b93b51833ee/packages/nc-gui/composables/useSharedFormViewStore.ts#L90-L98

wingkwong (changes requested):

packages/nc-gui/composables/useProject.ts
semmjon: I figured that by requiring at least the main database to be enabled (if no others exist/ enabled) that wouldn't happen.

packages/nc-gui/pages/[projectType]/[projectId]/index/index/index.vue
```diff
@@ -98,6 +98,7 @@ function openQuickImportDialog(type: QuickImportTypes, file: File) {
     'modelValue': isOpen,
     'importType': type,
     'onUpdate:modelValue': closeDialog,
+    'baseId': bases.value?.filter((base: any) => base.enabled)[0].id,
```
Comment (wingkwong) [inline packages/nc-gui/pages/[projectType]/[projectId]/index/index/index.vue:101]: don't use `any` when you know the type. Use `BaseType` instead.

wingkwong (changes requested):
Same change should be applied here as well.

https://github.com/nocodb/nocodb/blob/ef72406cb64ef90ff55a84402cca8e2fb6233960/packages/nc-gui/pages/%5BprojectType%5D/%5BprojectId%5D/index/index/index.vue#L129-L133

This is triggered by clicking `Add New Table`.

image

wingkwong (changes requested):
Still not cater 

image

wingkwong (approved):

Issue #5105, opened 2023-02-15
🔦 Feature: Rename Table CamelCase (Alias) only!


sgeist-ionos:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

A relatively small feature that can also be implemented quickly.
I want to be able to rename a table alias even if it's just CamalCase changes.

### Suggested Solution

Here is a git diff:

```javascript
diff --git a/packages/nocodb/src/lib/meta/api/tableApis.ts b/packages/nocodb/src/lib/meta/api/tableApis.ts
index 67019bc5c..69b5b7bbe 100644
--- a/packages/nocodb/src/lib/meta/api/tableApis.ts
+++ b/packages/nocodb/src/lib/meta/api/tableApis.ts
@@ -266,15 +266,11 @@ export async function tableUpdate(req: Request<any, any>, res) {
     );
   }
 
-  if (
-    !(await Model.checkTitleAvailable({
-      table_name: req.body.table_name,
-      project_id: project.id,
-      base_id: base.id,
-    }))
-  ) {
-    NcError.badRequest('Duplicate table name');
-  }
+  const schould_rename = await Model.checkTitleAvailable({
+    table_name: req.body.table_name,
+    project_id: project.id,
+    base_id: base.id,
+  });
 
   if (!req.body.title) {
     req.body.title = getTableNameAlias(
@@ -317,11 +313,13 @@ export async function tableUpdate(req: Request<any, any>, res) {
     req.body.table_name
   );
 
-  await sqlMgr.sqlOpPlus(base, 'tableRename', {
-    ...req.body,
-    tn: req.body.table_name,
-    tn_old: model.table_name,
-  });
+  if (schould_rename) {
+    await sqlMgr.sqlOpPlus(base, 'tableRename', {
+      ...req.body,
+      tn: req.body.table_name,
+      tn_old: model.table_name,
+    });
+  }
 
   Tele.emit('evt', { evt_type: 'table:updated' });
 
diff --git a/packages/nocodb/src/lib/models/Model.ts b/packages/nocodb/src/lib/models/Model.ts
index 3708b2e38..07da90980 100644
--- a/packages/nocodb/src/lib/models/Model.ts
+++ b/packages/nocodb/src/lib/models/Model.ts
@@ -752,16 +752,20 @@ export default class Model implements TableType {
     }: { title; project_id; base_id; exclude_id? },
[...]

PR #5161, opened 2023-02-19
feature: rename table camelcase (alias) only

GentleGhostCoder:
## Change Summary

PR for:
- https://github.com/nocodb/nocodb/issues/5105

## Change type

- [ x ] feat: (new feature for the user, not a new feature for build script)

## Test/ Verification

This commit adds a new feature (feature 5105). There are changes in two files, tableApis.ts and Model.ts.

The changes in the code allow the user to change the table alias without changing the table name, which was previously not possible. This is accomplished by checking if the new table alias is available and only renaming if it is. 

In the table unit test I changed the expaction of an status 400.
You should just expect the status 200 in the test.
Or would it make sense to create an extra response message (for changed alias only)?


packages/nocodb/src/lib/meta/api/tableApis.ts
```diff
@@ -277,15 +277,11 @@ export async function tableUpdate(req: Request<any, any>, res) {
...
-      base_id: base.id,
-    }))
-  ) {
-    NcError.badRequest('Duplicate table name');
-  }
+  const schould_rename = await Model.checkTitleAvailable({
+    table_name: req.body.table_name,
+    project_id: project.id,
+    base_id: base.id,
+  });
```
pranavxc, line 284: What if there are tables named `a` and `b` then the user tried to rename `a` as `b`, `should_rename` will be false but it won't return an error. The expected behavior is a bad request error with the `duplicate table name. 


Here you have to get the current table metadata and compare the name.

semmjon: yes you are right thank you
I will adjust that.

semmjon:
It works now. Maybe there is a nicer solution. You are welcome to make suggestions.

packages/nocodb/src/lib/models/Model.ts
```diff
@@ -750,19 +750,29 @@ export default class Model implements TableType {
...
       null,
       exclude_id && { id: { neq: exclude_id } }
-    ));
+    );
+
+    if (!meta) return true;
+    if (meta?.table_name || undefined == table_name) {
+      if (meta?.project_id == project_id) {
+        return false;
+      }
```
pranavxc, line 759: Is this if condition really required?  `project_id` will always be the same since we are querying with project_id

sgeist-ionos:
You are absolutely right. It must have been a mistake :)

wingkwong (changes requested):
@sgeist-ionos the changes are fine - can you move them to `nocodb`? `nocodb-legacy` will be removed soon.

wingkwong (changes requested):
please exclude all changes in `nocodb-legacy`. 

image

sgeist-ionos:
I noticed that the prefix behaves strangely when creating tables. When I create a table for the first time, the table has the following schema `nc_<id>___<table_name>` (3-underscore) and after I rename it once then only `nc_<id>__<table_name>` (2-underscore).

That's why I also added the line with the `.replace('___','__')` .

PR #5164, opened 2023-02-19
add feature upload xlsx in existing table

semmjon:
## Change Summary

PR for:
- closes https://github.com/nocodb/nocodb/issues/5021
- closes https://github.com/nocodb/nocodb/issues/4271

 I added the feature that enables users to upload Excel file templates and import data into an existing table. I accomplished this by modifying the QuickImport.vue file and adding some code to handle the uploading of the new file types. I also updated the ViewActions.vue file to display the new upload options in the project menu, and created a new QuickImportDialogType to differentiate between CSV, Excel, and JSON import options. Users will now be able to easily import data from Excel and JSON files into an existing table.

## Change type

- [x] feat: (new feature for the user, not a new feature for build script)
- [ ] fix: (bug fix for the user, not a fix to a build script)
- [ ] docs: (changes to the documentation)
- [ ] style: (formatting, missing semi colons, etc; no production code change)
- [ ] refactor: (refactoring production code, eg. renaming a variable)
- [ ] test: (adding missing tests, refactoring tests; no production code change)
- [ ] chore: (updating grunt tasks etc; no production code change)

## Test/ Verification

The first file, packages/nc-gui/components/dlg/QuickImport.vue, was changed to add support for uploading XLSX and JSON files into an existing table. Specifically, the following changes were made:

importState was changed from being an object created using reactive to a defaultImportState object. The importState variable is now created by making a reactive copy of defaultImportState. This is a best practice when using the reactive API.
A new defaultImportState object was created that defines default values for the fileList, url, jsonEditor, jsonSchema, delimiter, firstRowAsHeaders, and shouldImportData properties.
The handleImport function was modified to set templateEditorModal to false and to reset the importState to its default values after the import is complete.
The second file, packages/nc-gui/components/smartsheet/toolbar/ViewActions.vue, was modified to add support for uploading XLSX and JSON files. Specifically, the following changes were made:

A new QuickImportDialogType type was defined that specifies the different types of import dialogs that can be displayed.
A new quickImportDialogTypes array was defined that contains the different import dialog types.
A new quickImportDialogs object was defined that maps import dialog types to ref variables that control whether the dialog is displayed or not.
The a-menu-item template was modified to loop over each type of import dialog and display a menu item for each.
A new LazyDlgQuickImport component was added to the file. This component takes a v-model prop that controls whether the component is displayed or not. The component is used to display the import dialog when the corresponding menu item is clicked.

[...]

PR #5165, opened 2023-02-19
refactor bulk operations

semmjon:
## Change Summary

The changes made to the bulkDataAliasApis.ts file include the addition of a new function executeBulkOperation() and modifications to the existing functions bulkDataInsert(), bulkDataUpdate(), bulkDataUpdateAll(), bulkDataDelete(), and bulkDataDeleteAll().

The executeBulkOperation() function takes in parameters such as the request, response, the operation to perform (insert, update, or delete), and options to pass to the operation. It first gets the model, view, and base using the getModelAndBase() function, which retrieves the model and view from the request and then the base from the model. The baseModel is then obtained using the Model.getBaseModelSQL() function and the operation is executed on the baseModel object with the given options. If the operation throws an error, the error message is returned in the response with a 500 status code.

Each of the bulk data operations, bulkDataInsert(), bulkDataUpdate(), bulkDataUpdateAll(), bulkDataDelete(), and bulkDataDeleteAll(), has been modified to make use of the executeBulkOperation() function. For example, bulkDataInsert() now calls executeBulkOperation() with the parameters req, res, 'bulkInsert', and an array containing the request body and { cookie: req }.

The main change introduced by these modifications is the addition of error messages when an error occurs during a bulk operation. Previously, the API would return a generic error message without specifying the cause of the error. The new changes now pass the error that is returning from the database to the response.

## Change type

## Test/ Verification


wingkwong (commented):
Thanks for the PR. I agree moving the duplicate logic to `executeBulkOperation` is a good way to make it more clean. However, for the error messages, I got a doubt.

> Previously, the API would return a generic error message without specifying the cause of the error. 

Can you provide an example? In existing version, I can see the clear error message, e.g. 

```json
{"msg":"insert into `nc_pci____Sheet-3` (`title5`) values (DEFAULT), ('P') - Incorrect integer value: 'P' for column 'title5' at row 2"}
```

With your new changes, it shows below result while the info other than `message` (which is same as `msg` above) is not useful for debugging at all.

```
{"stack":"Error: insert into `nc_pci____Sheet-3` (`title5`) values (DEFAULT), ('P') - Incorrect integer value: 'P' for column 'title5' at row 2\n    at Packet.asError (/Users/wingkwong/Documents/GitHub/nocodb/packages/nocodb/node_modules/mysql2/lib/packets/packet.js:728:17)\n    at Query.execute (/Users/wingkwong/Documents/GitHub/nocodb/packages/nocodb/node_modules/mysql2/lib/commands/command.js:29:26)\n    at Connection.handlePacket (/Users/wingkwong/Documents/GitHub/nocodb/packages/nocodb/node_modules/mysql2/lib/connection.js:456:32)\n    at PacketParser.onPacket (/Users/wingkwong/Documents/GitHub/nocodb/packages/nocodb/node_modules/mysql2/lib/connection.js:85:12)\n    at PacketParser.executeStart (/Users/wingkwong/Documents/GitHub/nocodb/packages/nocodb/node_modules/mysql2/lib/packet_parser.js:75:16)\n    at Socket.<anonymous> (/Users/wingkwong/Documents/GitHub/nocodb/packages/nocodb/node_modules/mysql2/lib/connection.js:92:25)\n    at Socket.emit (node:events:520:28)\n    at Socket.emit (node:domain:475:12)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)","message":"insert into `nc_pci____Sheet-3` (`title5`) values (DEFAULT), ('P') - Incorrect integer value: 'P' for column 'title5' at row 2","code":"ER_TRUNCATED_WRONG_VALUE_FOR_FIELD","errno":1366,"sqlState":"HY000","sqlMessage":"Incorrect integer value: 'P' for column 'title5' at row 2","sql":"insert into `nc_pci____Sheet-3` (`title5`) values (DEFAULT), ('P')"}
```

semmjon:
Hi, i use it together with this function in the frontend: "extractSdkResponseErrorMsg" (you can see here the usage in the changes of https://github.com/nocodb/nocodb/pull/5164).
It was primarily intended to simplify debugging for the end user. In the future, I would like to extend the extractSdkResponseErrorMsg function and throw more user-friendly (specific) errors to convey to the user what to change.

wingkwong:
If `extractSdkResponseErrorMsg` needs to be there to make this PR work, then you have to add it in this PR as well. Besides, I still don't see how this PR simplifies debugging for the end users. From the above example, it shows more unnecessary info. Please share some examples that you tested to show it does improve the error messages.

semmjon:
The changes are in the backend and the extractSdkResponseErrorMsg should be used in the frontend (if needed) to are in the frontend. 
Before the changes, I got only no response from the backend...
You can debug this at the line 447 in the Editor.vue.
```
await $api.dbTableRow.bulkCreate('noco', projectName, tableId!, batchData).then((res) => {
                  console.log(res)
                }).catch((e: any) => {
                  console.log(e)
                })
```
If the bulkUpload is working I get:
```
[
    <some meaningless number>
]
```
If the bulkUpload fails (also not really helpful for the API / GUI - User):
```
Error: Request failed with status code 400
    at createError (createError.js:16:15)
    at settle (settle.js:17:12)
    at XMLHttpRequest.onloadend (xhr.js:54:7)
```

I know that we can also see the error message in the server-logs, but the users cant see it.
If you think the tracelog is too much, I can just put the DB message in the response.

semmjon:
You were right. The extra catch is not necessary. I removed this one. It can be obtained in the frontend through the `error.response`. Thanks!

packages/nocodb/src/lib/meta/api/dataApis/bulkDataAliasApis.ts
```diff
@@ -6,70 +6,61 @@ import ncMetaAclMw from '../../helpers/ncMetaAclMw';
 import { getViewAndModelFromRequestByAliasOrId } from './helpers';
 import apiMetrics from '../../helpers/apiMetrics';
 
-async function bulkDataInsert(req: Request, res: Response) {
+async function getModelAndBase(req: Request) {
```
wingkwong, line 9: Revise the function name as it also return View as well, not just model and base.

packages/nocodb/src/lib/meta/api/dataApis/bulkDataAliasApis.ts
```diff
@@ -6,70 +6,61 @@ import ncMetaAclMw from '../../helpers/ncMetaAclMw';
...
   const { model, view } = await getViewAndModelFromRequestByAliasOrId(req);
 
   const base = await Base.get(model.base_id);
+  return { model, view, base };
+}
 
+async function executeBulkOperation(
+  req: Request,
+  res: Response,
+  operation: string,
```
wingkwong, line 19: we need to ensure the operation like this.

```ts
operation: 'bulkInsert' | 'bulkUpdate' | 'bulkUpdateAll' | 'bulkDelete' | 'bulkUpdateAll'
```

wingkwong (changes requested):

wingkwong (commented):
Still not catering https://github.com/nocodb/nocodb/pull/5165#discussion_r1113790845

packages/nocodb/src/lib/meta/api/dataApis/bulkDataAliasApis.ts
```diff
@@ -1,75 +1,77 @@
...
+    | 'bulkUpdateAll'
+    | 'bulkDelete'
+    | 'bulkDeleteAll',
+  options:
+    | Parameters<typeof BaseModelSqlv2.prototype.bulkInsert>
+    | Parameters<typeof BaseModelSqlv2.prototype.bulkUpdate>
+    | Parameters<typeof BaseModelSqlv2.prototype.bulkDelete>
+    | Parameters<typeof BaseModelSqlv2.prototype.bulkUpdateAll>
+    | Parameters<typeof BaseModelSqlv2.prototype.bulkDeleteAll>
+) {
```
wingkwong, line 32: since you are revising options types as well, I think it could be further simplified as 

```ts
async function executeBulkOperation<
  operation extends
    | 'bulkInsert'
    | 'bulkUpdate'
    | 'bulkUpdateAll'
    | 'bulkDelete'
    | 'bulkDeleteAll'
>(
  req: Request,
  res: Response,
  operation,
  options: Parameters<typeof BaseModelSqlv2.prototype[operation]>
) {
```

wingkwong (changes requested):

packages/nocodb/src/lib/meta/api/dataApis/bulkDataAliasApis.ts
semmjon: Yes, I had already thought of that and tested it that way. Somehow I decided against it. Probably because I thought you would have to pass the template variable (redundantly) when calling the function. But it seems to be working (as it currently is). Good that you exchange ideas in the community :)

wingkwong (approved):

Issue #3500, opened 2022-09-05
🐛 Bug: import excel throw error


tomorrowcheng:
**Please enter the following details**

```
Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
Database: **mysql2**
ProjectOnRootDB: **false**
RootDB: **mysql2**
PackageVersion: **0.90.5**
```

**Steps To Reproduce**

1. import excel like this:

[测试表格.xlsx](https://github.com/nocodb/nocodb/files/9490181/default.xlsx)

**Expected Behavior**

import success

**Screenshots**

import failed

tomorrowcheng:
> any particular reason to use `0.90.5` ? Suggest to try the latest version.

used the latest version; Sorry, I gave the wrong excel；You can try this:
[测试表格.xlsx](https://github.com/nocodb/nocodb/files/9490327/default.xlsx)

wingkwong:
preliminary findings:

- dtxp looks fine: `FR\\BI','FR','FR\\FT`
- the data type in db is incorrect: `enum('FRBI','FR','FRFT')` while `enum('FR\\BI','FR','FR\\FT')` is expected

fulldecent:
I am experiencing this in the latest version 0.207

![Screenshot 2024-05-13 at 11 18 09](https://github.com/nocodb/nocodb/assets/382183/bac67df0-9c84-4bcd-a686-7629a88de853)

dstala:
Import defaults to single line text now. Please reopen this issue if you face it still in latest release

PR #5166, opened 2023-02-19
refactor xlsx import (ExcelTemplateAdapter.ts)

GentleGhostCoder:
## Change Summary

Additionally, I have implemented a feature that was not previously documented:
This feature allows for a more dynamic way of importing Excel files. This includes the ability to handle files that have spaced lines before the table or an autoFilter active and to import the Excel header correctly. In the future, we can add the configuration "dynamicHeaders" to the Editor.vue Component (Checkbox).

This changes made to the ExcelTemplateAdapter class in the ExcelTemplateAdapter.ts file, which has been refactored to improve the code's readability, structure, and maintainability.

The refactor changes made involve the following:

-An interface ParserConfig has been defined that contains the configuration options for the Excel template parser.
-The config property of the ExcelTemplateAdapter class has been updated to be of type ParserConfig.
- New utility methods isDecimalVal, isCurrencyVal, isPercentageVal, isMultiLineTextVal, isEmailVal, isUrlVal, isCheckboxVal, isNumberVal, isAllDate were added to determine the value types.
- New utility methods for formatting cell values like defaultFormater, defaultRawFormater, dateFormatter, dateTimeFormatter, checkBoxFormatter, currencyFormatter, percentFormatter were added.
-A function has been added to check whether all the values in a given array are of the Date type.
-A regular expression has been added to match special characters in strings.
-The ExcelTemplateAdapter class constructor has been updated to accept a parserConfig object as its second parameter, with default values for its properties.
-The fixImportedDate method is moved from the parse method (seperated) and the used constants added to the class

This Refactoring also (maybe) fix this issue. However import the example file works fine.
https://github.com/nocodb/nocodb/issues/3500

## Change type

## Test/ Verification

Provide summary of changes.


wingkwong (changes requested):
1. Tested several excel files - none of them worked
2. Those utils functions can be placed in `packages/nc-gui/utils/parsers/parserHelpers.ts`
3. This PR doesn't fix issue 3500. The issue is still reproducible. Actually it is related to MySQL. I think you're testing in other db engines.

semmjon:
No i use also mysql. Can you provide some example Excel Files?

wingkwong:
You may test with the files inside [pr5166-test.zip](https://github.com/nocodb/nocodb/files/10783837/pr5166-test.zip). Just some random files I've used before, it may not cover all the cases.

For issue 3500, what is your version of mysql?

semmjon:
Ok I got the error. 
My Excel was a little different, which is why I always loaded it as singlelinetext and not as single-selection. 
As soon as you load it as a single-selection, an enum is created and backslashes are not allowed or are not saved as such (without them). The solution is to use a single selection either the respective id (number) of the respective enum when uploading/inserting or just the same value without the backslash.

wingkwong:
For issue 3500, please see the findings [here](https://github.com/nocodb/nocodb/issues/3500#issuecomment-1246290953). And I think we can have another PR to handle that. 

Therefore, for this PR, let's focus on the refactoring instead.

wingkwong (changes requested):
These files can be successfully imported in existing version, but not in this PR.

- [pr5166-test.zip](https://github.com/nocodb/nocodb/files/10783837/pr5166-test.zip)
- [2COVID-19.xlsx.zip](https://github.com/nocodb/nocodb/files/10820833/2COVID-19.xlsx.zip)

semmjon:
There is a single bug while reading pr5166-test-4.xlsx. I'll fix it as soon as it fits.

semmjon:
OK, when i try to import `pr5166-test-4.xlsx`. I get the error:
```
ER_TOO_BIG_ROWSIZE
...
Row size too large (> 8126). Changing some columns to TEXT or BLOB may help. In current row format, BLOB prefix of 0 bytes is stored inline.
```

I think this has nothing to do with the import.
In the previous version, some columns were (incorrectly) recognized as checkboxes, which is why the error may not have occurred.
However, there are many vachar(255) data types (e.g. even with empty columns).
You could work around the error by changing a different data type instead of vachar(255) (e.g. TEXT / BLOB) or by setting the following option.
```
SET SESSION innodb_strict_mode=OFF;
```

wingkwong:
I could import that in your branch without seeing any error. Will have a review tomorrow.

wingkwong (changes requested):
I'd suggest you to start a docker container of the current version and cross-check the results.

1. Not working for all files when `Use First Row as Header` is unchecked. If it is unchecked, a default column name `field1`, `field2`, `field3`, ... will be used.
2. Couldn't import [tests-example.xls](https://github.com/nocodb/nocodb/files/10846594/tests-example.xls) in this PR branch
3. Imported Data (pr5166-test-2.xlsx) is wrong. Beside, the formatting is also wrong. (Left: PR branch, Right: Current version)
image
4. We shouldn't change the user data (Left: PR branch, Right: Current version)
image
5. In pr5166-test-4.xlsx, `new_cases` becomes empty for those with value `0` 
image
and some data are missing.
image

semmjon:
My apologies for the unsatisfactory testing. The more I would like to thank you for the following tests :)
Is there a way to run the unit tests locally? Does this also work with Docker containers?

semmjon:
I noticed other errors.
I think I'll leave that in my branch for the time being and test it further.
maybe I'm looking at how I can test things locally via tests.

wingkwong:
@semmjon 

> Is there a way to run the unit tests locally?

See [here](https://docs.nocodb.com/engineering/unit-testing#run-tests)

Converting back to draft at this moment. Please click `Ready for review` when it is ready.

dstala:
@GentleGhostCoder Closing this as its on very old code base. If you find any issue with CSV / Excel importer, create an issue & we can track further from that point.

PR #5181, opened 2023-02-21
Do not check mime type during attachment download

djmaze:
## Change Summary

Do not check mime type during attachment download. Also, just do a HEAD request instead of downloading the whole file.

The reasons for doing this are that 

a) the mime type list is not complete. Thus the download fails for some file types (e.g. `audio/flac`).
b) the mime type returned by S3-based CDNs (DigitalOcean Spaces in this case) can differ from the one stored by NocoDB. 

To be honest, I do not understand why a mime type check was added here at all.

## Change type

## Test/ Verification

1. Create a table with an attachment field.
2. Create a record with an attachment.
3. Try to download the attachment.


wingkwong (approved):
The original idea was the download links could be broken during the migration. Hence, adding a mimetype verification here to ensure the type is same. However, given the reasons you mentioned and the invalid data should be shown in the modal already, I think it'd better to remove it.

Let's wait for the tests. If they are good then we can merge it.

PR #5189, opened 2023-02-22
fix: Disable button to avoid duplicate api call when creating column, project and table

### LINKED ISSUE #5158: 🐛 Bug: clicking save button multiple times on "new column" or "edit column" will fire multiple queries

### Steps to reproduce ?

1. Use a table with millions of rows that will take a few seconds to load a page of 25 elements.
2. Edit a column or save a new column.
3. Click the "save" button multiple times.  Each time will fire a query.

If the queries are expensive and enough are fired, it can cause the UI to timeout.

### Desired Behavior

The button should be disabled to prevent multiple duplicate queries.

### Project Details

Node: **v16.17.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **sqlite3**
PackageVersion: **0.105.2**

### Attachments

Screenshot 2023-02-19 at 11 49 19 AM

pranavxc:
## Change Summary

close #5158

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/components/dlg/TableCreate.vue
+++ b/packages/nc-gui/components/dlg/TableCreate.vue
@@ -78,14 +78,18 @@ const systemColumnsCheckboxInfo = SYSTEM_COLUMNS.map((c, index) => ({
   disabled: index === 0,
 }))
 
+const creating = ref(false)
+
 const _createTable = async () => {
   try {
+    creating.value = true
     await validate()
   } catch (e: any) {
     e.errorFields.map((f: Record<string, any>) => message.error(f.errors.join(',')))
     if (e.errorFields.length) return
   }
   await createTable()
+  creating.value = false
 }
 
 onMounted(() => {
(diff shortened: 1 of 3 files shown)


packages/nc-gui/components/dlg/TableCreate.vue
```diff
@@ -78,14 +78,18 @@ const systemColumnsCheckboxInfo = SYSTEM_COLUMNS.map((c, index) => ({
...
 const _createTable = async () => {
   try {
+    creating.value = true
     await validate()
   } catch (e: any) {
     e.errorFields.map((f: Record<string, any>) => message.error(f.errors.join(',')))
     if (e.errorFields.length) return
   }
   await createTable()
+  creating.value = false
```
wingkwong, line 92: This line should be put inside `finally` block so that it will set to `false` even there is an error. Otherwise, users  cannot resubmit like below.

image

packages/nc-gui/pages/index/index/create.vue
```diff
@@ -48,6 +52,7 @@ const createProject = async () => {
   } catch (e: any) {
     message.error(await extractSdkResponseErrorMsg(e))
   }
+  creating.value = false
```
wingkwong, line 55: even though this line is working fine, I'd prefer to put it in `finally` block as well since it starts in `try` block.

wingkwong (changes requested):

dstala (approved):

wingkwong (approved):

PR #5200, opened 2023-02-23
fix bug when creating single- / multi-select (enum / set) with `\` inside

semmjon:
…side the name

## Change Summary

Related to: https://github.com/nocodb/nocodb/issues/3500

The commit message indicates that it fixes a bug that occurs when creating an enum/set and selecting single/multiple options that contain the **`\`** character in their name.

The first file, **Editor.vue**, has changes made to the remapColNames function starting at line 275. The changes include adding an else if block to handle cases where the **col.uid** is either **UITypes.SingleSelect** or **UITypes.MultiSelect**. In such cases, the data[key] string is modified by replacing all instances of the **`\`** character with an empty string (like the databases do this).

The second file, useViewData.ts, has changed to the useViewData function starting at line 279. The changes include modifying the value of [property] based on the dt property of the column in the metaValue object. If the dt is either **enum** or **set**, then the replaceAll function replaces all instances of the **`\`** character with an empty string before assigning the value to [property].

If you were to write a pull-request comment, you could say something like:

## Change type

## Test/ Verification

This is a relatively simple fix and may not be supported for all databases. I tested it with MYSQL, where I also encountered the error. It would have to be checked whether other databases behave in the same way. If this is not the case, you have to make the bug fix database specific.

The bug issue describes the error when importing an Excel file. However, this is independent of the fix, since the bug also arises when manually creating multi/single-selection columns. When importing, it is enough (similar to header normalization) not to allow it in the first place and to make a conversion.

### DIFF
--- a/packages/nc-gui/components/template/Editor.vue
+++ b/packages/nc-gui/components/template/Editor.vue
@@ -275,6 +275,8 @@ function remapColNames(batchData: any[], columns: ColumnType[]) {
       } else if (col.uidt === UITypes.DateTime && d) {
         const dateTimeFormat = getDateTimeFormat(data[key])
         d = dayjs(data[key], dateTimeFormat).format('YYYY-MM-DD HH:mm')
+      } else if (col.uid === UITypes.SingleSelect || col.uid === UITypes.MultiSelect) {
+        d = data[key].replace('\\', '')
       }
       return {
         ...aggObj,
--- a/packages/nc-gui/composables/useViewData.ts
+++ b/packages/nc-gui/composables/useViewData.ts
@@ -279,7 +279,11 @@ export function useViewData(
         encodeURIComponent(id),
         {
           // if value is undefined treat it as null
-          [property]: toUpdate.row[property] ?? null,
+          [property]:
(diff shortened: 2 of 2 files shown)


wingkwong (changes requested):
1. Column Type should be detected as SingleSelect
image
2. Even I change to SingleSelect, I'd get the same error. 

Tested in MYSQL 8.0.31.

dstala:
> 1. Column Type should be detected as SingleSelect
>    image
> 2. Even I change to SingleSelect, I'd get the same error.
> 
> Tested in MYSQL 8.0.31.

Had tried import yesterday, did work fine (was selected as single select). Had used docker mysql

semmjon:
Pick up again. It's not a fix for import, but for creating/updating enums/sets (single-select/multi-select). 
The import was resolved in the following PR:
https://github.com/nocodb/nocodb/pull/5166

wingkwong (changes requested):
This won't work. Even though you can create the option, you won't be able to select it. 
image
Besides, the changes won't cater options like `a\b` or `a\\\b`. MySQL would escape it since `\b` is backspace character.

wingkwong:
Closing - inactivity.

PR #5317, opened 2023-03-17
enhancement: webhooks

wingkwong:
## Change Summary

- add enable option for webhook
- add cloning webhook feature (by default, it will set the copy hook to inactive)
- add i18n in webhook
- add modal to confirm deletion
- fix some types related to hook / filter

## Change type

## Test/ Verification

image

image


dstala (changes requested):
- [x] webhook were triggered even after they were disabled. tried using webhook URL from webhook.site. after insert trigger. 

- [x] discussed with pranav. he mentioned, we can remove `auth` tab as its not being used. currently if i store some value, it vanishes on reload. let us hide it
  
Screenshot 2023-03-18 at 12 07 33 PM



Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
RootDB: **sqlite3**
PackageVersion: **0.105.3**

dstala (approved):

PR #5364, opened 2023-03-25
fix (dbConfig): ssl connection with file path is not handled correctly

kamalmahmudi:
## Change Summary

Let's say I have postgres server that only accepts ssl connection. But the cert on the server is self-signed. For this, I need to provide ca cert to nocodb. This is my command on packages/nocodb:

`NC_DB="pg://ssldbhost:5400?u=theuser&p=thepass&d=nocodb&connection.ssl.caFilePath=/config/ca.crt" npm run run`

I will get `SELF_SIGNED_CERT_IN_CHAIN`. This is because the `promisify` returns a `Promise` and the returned promise is converted by `toString()` first before it's resolved. And `dbConfig.connection.ssl.ca` now has `[object Promise]` as value.

## Change type

## Test/ Verification

Run the command from the `Change Summary` again. It should gives us `App started successfully.`.

- `dbConfig.connection.ssl.caFilePath` is an extra parameter for pg connection, and somehow here it must be removed or it won't work.
- We can also use `fs.readFileSync`, but I don't use it, to keep the original approach.


wingkwong (commented):
It looks good. Can you also apply the same changes on `/nocodb/packages/nocodb/src/lib/db/sql-client/lib/SqlClientFactory.ts`?

wingkwong (approved):

Issue #719, opened 2021-10-31
Doesn't create relations from Postgresql DB


nzscout:
**Please enter the following details**
Node: **v17.0.1**
Arch: **x64**
Platform: **linux**
Docker: **false**
Database: **pg**
ProjectOnRootDB: **false**
RootDB: **sqlite3**
PackageVersion: **0.80.0**

**Steps To Reproduce**
1. Creat a new PostgreSQL database
2. Import northwind database from https://github.com/pthom/northwind_psql
3. Create a new project in Nocodb using the existing database

**Expected behavior**
Tables and relations are created

**Actual behavior**
Tables are created but Relations are empty
Trying to reload them doesn't help

Relations are created if SQL Server database is used

o1lab:
Just tried northwind database and I able to see the relations (see below).  Are you not able to see a specific relation or all relations ?

<img width="2274" alt="Screen Shot 2021-10-31 at 10 15 20" src="https://user-images.githubusercontent.com/5435402/139577887-8f5d8c0c-9e08-4613-bfc8-2aecb8bb65be.png">

incoventures:
I have no relations at all. I tried to create a new project twice with the same result. The foreign keys are present in the database. My PostgreSQL is 13.4

o1lab:
I'm on PG 9.6.x

Connect to your pg via PGAdmin/tableplus and run below query to see what is the result like - this gives relation list for the database/schema.

```sql
SELECT DISTINCT tc.table_schema as ts,
                                                      tc.constraint_name as cstn,
                                                      tc.table_name as tn,
                                                      kcu.column_name as cn,
                                                      ccu.table_schema AS foreign_table_schema,
                                                      ccu.table_name   AS rtn,
                                                      ccu.column_name  AS rcn,
                                                      pc.confupdtype   as ur,
                                                      pc.confdeltype   as dr
                                               FROM information_schema.table_constraints AS tc
                                                      JOIN information_schema.key_column_usage AS kcu
                                                           ON tc.constraint_name = kcu.constraint_name
                                                             AND tc.table_schema = kcu.table_schema
                                                      JOIN information_schema.constraint_column_usage AS ccu
                                                           ON ccu.constraint_name = tc.constraint_name
                                                             AND ccu.table_schema = tc.table_schema
                                                      join (select conname, confupdtype, confdeltype
                                                            from pg_catalog.pg_constraint) pc
                                                           on pc.conname = tc.constraint_name
[...]

PR #5397, opened 2023-03-31
Fix/pg rels as non table owner

data-envoy:
## Change Summary

Updated 2 quereis to allow for table relations to be returned in a case where the pg user is not the table owner.

The changes have been thouroughly commented and this will need review please.

It should fix https://github.com/nocodb/nocodb/issues/719

## Change type

## Test/ Verification

The changes work when using the web interface + PG15.

I have not yet run tests. I hope they will be run as an action upon pull request.

To see this change 

1. Create a couple of tables with a foreign key relationship.
2. Create a new user with access to these tables but is not owner.
  ```
DO $$ BEGIN
  IF NOT EXISTS (SELECT 1 FROM pg_user WHERE usename = 'noco_data') THEN
    CREATE USER noco_data WITH PASSWORD 'password'; 
  END IF;
END $$;

GRANT USAGE  ON SCHEMA public TO noco_data;
GRANT SELECT ON all tables IN SCHEMA public TO noco_data;

GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE table_1 TO noco_data;
GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE table_2 TO noco_data;
  ```
3. Add the database as a datasource in the web ui.
4. Click ERD you will see table relations. (these were not being shown earlier)

### DIFF
--- a/packages/nocodb/src/lib/db/sql-client/lib/pg/PgClient.ts
+++ b/packages/nocodb/src/lib/db/sql-client/lib/pg/PgClient.ts
@@ -997,6 +997,8 @@ class PGClient extends KnexClient {
            JOIN pg_namespace sch ON sch.oid = tbl.relnamespace
            JOIN pg_attribute col ON (col.attrelid = tbl.oid AND col.attnum = u.attnum)
         where tbl.relname='${args.tn}'
+        -- Possible bug. column "constraint_name" & "constraint_type" does not exist
+        -- GROUP BY conname, col.attnum, contype, "schema", "table", definition
         GROUP BY constraint_name, col.attnum, constraint_type, "schema", "table", definition
         ORDER BY "schema", "table"; `);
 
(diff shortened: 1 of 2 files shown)


data-envoy:
I've removed sql comments that were causing a unit test to fail.

pranavxc (approved):

packages/nocodb/src/lib/db/sql-client/lib/pg/PgClient.ts
```diff
@@ -1053,28 +1055,56 @@ class PGClient extends KnexClient {
...
+        //     JOIN information_schema.key_column_usage AS kcu
+        //       ON tc.constraint_name = kcu.constraint_name
+        //       AND tc.table_schema = kcu.table_schema
+        //     JOIN information_schema.constraint_column_usage AS ccu
+        //       ON ccu.constraint_name = tc.constraint_name
+        //       AND ccu.table_schema = tc.table_schema
+        //     join (select conname,confupdtype,confdeltype from pg_catalog.pg_constraint) pc
+        //     on pc.conname = tc.constraint_name
+        // WHERE tc.constraint_type = 'FOREIGN KEY' AND tc.table_schema=? and tc.table_name='${args.tn}'
+        // order by tc.table_name;`,
```
pranavxc, line 1087: this old query can be removed

data-envoy:
I've removed the excess comments as i the review feedback.

Regarding a potential bug that I [found here](https://github.com/nocodb/nocodb/blob/ba2c951f70b1769eba13eafed51593471161f405/packages/nocodb/src/lib/db/sql-client/lib/pg/PgClient.ts#L1000).

Do you think the code should be updated in this PR, even though it's unrelated to this issue? I don't know it it's causing an issue, or if that code path is ever being run.

packages/nocodb/src/lib/db/sql-client/lib/pg/pg.queries.ts
```diff
@@ -150,11 +150,6 @@ AND t.table_name=?;`,
 
   relationList: {
```
data-envoy: One think I'm not sure about, what is the difference between pg queries here and in the PgClient.ts file?
Is this legacy code?

pranavxc: We are no longer using it, so you can ignore it.

salim-b:
@pranavxc May I ask what's the status of this PR? Would be really cool if #719 could be fixed finally 🙃

sebdodt:
Thanks for all the work on this! I’m also facing the foreign-key visibility issue when using a non-owner Postgres role. It’s great to see that this PR fixes it by querying `pg_catalog`. I’d really love to see this merged—right now I’m resorting to awkward workarounds (assigning ownership + event triggers) just so NocoDB can display relationships. Let me know if there’s anything I can do to help move this forward!

pranavxc:
@data-envoy: Created a new PR with your changes and resolved conflicts(#10365), Thanks for the fix. Closing this pr and soon we will merge after testing couple of scenarios.

PR #5430, opened 2023-04-06
enhancement: cache on list

wingkwong:
## Change Summary

- Currently empty list will not be cached (e.g. sort / filters) - which hit the db on every page load. This PR is to resolve this problem by adding a `NONE` since we cannot have empty set in redis.

TODO:

- sync to nest branch once it is merged

## Change type

## Test/ Verification

Provide summary of changes.


starbirdtech383:
looks good to me

packages/nocodb/src/lib/cache/RedisCacheMgr.ts
```diff
@@ -125,9 +131,22 @@ export default class RedisCacheMgr extends CacheMgr {
...
+        isNoneList,
+      });
+    }
+
+    return {
+      list: await Promise.all(
+        arr.map(async (k) => await this.get(k, CacheGetType.TYPE_OBJECT))
+      ),
+      isNoneList,
+    };
```
pranavxc, line 149: Instead of relying on the first value in the array `'NONE'`, what about checking the subkeys array, If it's defined it will return an array otherwise it will be undefined/null. Can we use the following method?

```
 const arr = (await this.get(key, CacheGetType.TYPE_ARRAY));
    log(`RedisCacheMgr::getList: getting list with key ${key}`);

    const isNoneList = !!arr;
```

The current solution is fine and the only drawback is we can't save `NONE` as the first value in the array which is never going to happen in our current scenario.

packages/nocodb/src/lib/models/GridViewColumn.ts
```diff
@@ -99,13 +103,10 @@ export default class GridViewColumn implements GridColumnType {
...
-      (
-        await NocoCache.getList(CacheScope.GRID_VIEW_COLUMN, [
-          column.fk_view_id,
-        ])
-      )?.length
-    )
+    const { list } = await NocoCache.getList(CacheScope.GRID_VIEW_COLUMN, [
+      column.fk_view_id,
+    ]);
+    if (list.length)
```
pranavxc, line 109: `isNoneList` check not required?

packages/nocodb/src/lib/models/MapViewColumn.ts
```diff
@@ -67,10 +67,10 @@ export default class MapViewColumn {
...
 
     // if cache is not present skip pushing it into the list to avoid unexpected behaviour
-    if (
-      (await NocoCache.getList(CacheScope.MAP_VIEW_COLUMN, [column.fk_view_id]))
-        ?.length
-    )
+    const { list } = await NocoCache.getList(CacheScope.MAP_VIEW_COLUMN, [
+      column.fk_view_id,
+    ]);
+    if (list?.length)
```
pranavxc, line 73: isNoneList check not required?

packages/nocodb/src/lib/cache/RedisCacheMgr.ts
wingkwong: > the only drawback is we can't save NONE as the first value in the array which is never going to happen in our current scenario.

It's in `setList`. We save `NONE` when the input list is empty. 

Will test your approach.

packages/nocodb/src/lib/models/MapViewColumn.ts
wingkwong: Same as above.

packages/nocodb/src/lib/cache/RedisCacheMgr.ts
wingkwong: Just tested a bit. In this way, we never trigger metaList2. Let's keep it as it is first.

pranavxc (approved):

PR #5455, opened 2023-04-12
feat(nc-gui): add check all on action column editor import modal

khisby:
## Change Summary

add checkbox for check all.   
this makes it easy to check all the checkboxes. then if there are checkboxes that you want to exclude, just select them one by one

## Change type

## Test/ Verification

Provide summary of changes.

image

### DIFF
--- a/packages/nc-gui/components/template/Editor.vue
+++ b/packages/nc-gui/components/template/Editor.vue
@@ -87,6 +87,8 @@ const isImporting = ref(false)
 
 const importingTips = ref<Record<string, string>>({})
 
+const checkAllRecord = ref(false)
+
 const uiTypeOptions = ref<Option[]>(
   (Object.keys(UITypes) as (keyof typeof UITypes)[])
     .filter(
@@ -615,6 +617,13 @@ function handleEditableTnChange(idx: number) {
 function isSelectDisabled(uidt: string, disableSelect = false) {
   return (uidt === UITypes.SingleSelect || uidt === UITypes.MultiSelect) && disableSelect
 }
+
+function handleCheckAllRecord(event, table_name){
+  const isChecked = event.target.checked;
+  for (const record of srcDestMapping.value[table_name]) {
+    record.enabled = isChecked;
+  }
+}
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


wingkwong (changes requested):
Thanks for the PR. Here's some comments.

1. Buttons are not aligned
image
2. If I import two files and uncheck all columns for both files at the beginning, checking all columns on the first file would also affect the second file. See below screenshots.

First File: Check all columns as expected
image

Second File: Unexpected check due to the above action
image

After the changes, please run `npm run lint` on `nc-gui`.

khisby:
> Thanks for the PR. Here's some comments.
> 
> 1. Buttons are not aligned
>    image
> 2. If I import two files and uncheck all columns for both files at the beginning, checking all columns on the first file would also affect the second file. See below screenshots.
> 
> First File: Check all columns as expected image
> 
> Second File: Unexpected check due to the above action image
> 
> After the changes, please run `npm run lint` on `nc-gui`.

Hi @wingkwong 
i've already fix all point.
I have tested and work well.
then, i've run `npm run lint`

packages/nc-gui/components/template/Editor.vue
```diff
@@ -615,6 +617,13 @@ function handleEditableTnChange(idx: number) {
 function isSelectDisabled(uidt: string, disableSelect = false) {
   return (uidt === UITypes.SingleSelect || uidt === UITypes.MultiSelect) && disableSelect
 }
+
+function handleCheckAllRecord(event, table_name) {
```
wingkwong, line 621: parameter implicitly has an 'any' type - set it when you know the type

```ts
import type { CheckboxChangeEvent } from 'ant-design-vue/es/checkbox/interface'
```

```ts
function handleCheckAllRecord(event: CheckboxChangeEvent, table_name: string) {
```

wingkwong (changes requested):
One more minor change.

packages/nc-gui/components/template/Editor.vue
khisby: thankyou for your suggest.
I have changed the code

wingkwong (approved):

wingkwong:
Thanks. Let's wait for the test.

PR #5497, opened 2023-04-17
fix: bridge table identification

mertmit:
## Change Summary

Inventory table on sakila db was mapped as mm incorrectly.
- Check if bridge table has exactly 2 primary keys
- SQLite pragma table_info returns pk as incremental so use >0 instead of === 1 to support composite pks

## Change type


pranavxc (approved):

mertmit:
this is included in #5608

PR #5505, opened 2023-04-19
refactor: timezone

wingkwong:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/plugins/a.dayjs.ts
+++ b/packages/nc-gui/plugins/a.dayjs.ts
@@ -4,6 +4,7 @@ import customParseFormat from 'dayjs/plugin/customParseFormat.js'
 import duration from 'dayjs/plugin/duration.js'
 import utc from 'dayjs/plugin/utc.js'
 import weekday from 'dayjs/plugin/weekday.js'
+import timezone from 'dayjs/plugin/timezone.js'
 import { defineNuxtPlugin } from '#imports'
 
 export default defineNuxtPlugin(() => {
@@ -12,4 +13,5 @@ export default defineNuxtPlugin(() => {
   extend(customParseFormat)
   extend(duration)
   extend(weekday)
+  extend(timezone)
 })
--- a/packages/nocodb/src/lib/meta/NcMetaIOImpl.ts
+++ b/packages/nocodb/src/lib/meta/NcMetaIOImpl.ts
@@ -1,5 +1,6 @@
 import CryptoJS from 'crypto-js';
 import { customAlphabet } from 'nanoid';
+import dayjs from 'dayjs';
 import { XKnex } from '../db/sql-data-mapper';
 import XcMigrationSource from '../migrations/XcMigrationSource';
 import NcConnectionMgr from '../utils/common/NcConnectionMgr';
@@ -94,6 +95,20 @@ export default class NcMetaIOImpl extends NcMetaIO {
     return (this.trx || this.connection) as any;
   }
 
+  private isMySQL(): boolean {
+    return (
+      this.config?.meta?.db?.client === 'mysql' ||
+      this.config?.meta?.db?.client === 'mysql2'
+    );
+  }
+
+  private now(): any {
+    if (this.isMySQL()) {
+      return dayjs().utc().format('YYYY-MM-DD HH:mm:ss');
+    }
+    return dayjs().utc().toISOString();
+  }
+
   public updateKnex(connectionConfig): void {
     this.connection = XKnex(connectionConfig);
   }
@@ -107,6 +122,10 @@ export default class NcMetaIOImpl extends NcMetaIO {
       migrationSource: new XcMigrationSourcev2(),
       tableName: 'xc_knex_migrationsv2',
     });
+    if (this.isMySQL()) {
+      // set timezone
+      await this.connection.raw(`SET time_zone = '+00:00'`);
+    }
     return true;
   }
 
@@ -238,8 +257,8 @@ export default class NcMetaIOImpl extends NcMetaIO {
     return this.knexConnection(target).insert({
       db_alias: dbAlias,
       project_id,
-      created_at: this.knexConnection?.fn?.now(),
-      updated_at: this.knexConnection?.fn?.now(),
+      created_at: this.now(),
+      updated_at: this.now(),
       ...data,
     });
   }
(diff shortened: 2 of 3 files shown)


packages/nc-gui/components/virtual-cell/Formula.vue
```diff
@@ -10,12 +11,27 @@ const cellValue = inject(CellValueInj)
...
 const { showEditNonEditableFieldWarning, showClearNonEditableFieldWarning, activateShowEditNonEditableFieldWarning } =
   useShowNotEditableWarning()
+
+const renderResult = (result?: string | null) => {
+  if (!result) {
+    return ''
+  }
+  // convert all date time values to local time
+  // the input is always YYYY-MM-DD hh:mm:ss+xx:yy
+  return result.replace(/\b(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\+\d{2}:\d{2})\b/g, (d) => {
```
pranavxc, line 29: `+` part could be `-` as well, so you can use `[+-]`

packages/nc-gui/components/virtual-cell/components/ItemChip.vue
```diff
@@ -60,7 +61,7 @@ export default {
     :class="{ active }"
     @click="openExpandedForm"
   >
-    <span class="name">{{ value }}</span>
+    <span class="name">{{ renderValue(value) }}</span>
```
pranavxc, line 64: We can use cell component itself to render it, the same way Lookup is rendered

packages/nc-gui/components/virtual-cell/components/ListChildItems.vue
```diff
@@ -148,7 +149,7 @@ const onClick = (row: Row) => {
           >
             <div class="flex items-center">
               <div class="flex-1 overflow-hidden min-w-0">
-                {{ row[relatedTableDisplayValueProp] }}
+                {{ renderValue(row[relatedTableDisplayValueProp]) }}
```
pranavxc, line 152: check previous comment

packages/nc-gui/components/virtual-cell/components/ListItems.vue
```diff
@@ -229,7 +230,7 @@ watch(vModel, (nextVal) => {
             :class="{ 'nc-selected-row': selectedRowIndex === i }"
             @click="linkRow(refRow)"
           >
-            {{ refRow[relatedTableDisplayValueProp] }}
+            {{ renderValue(refRow[relatedTableDisplayValueProp]) }}
```
pranavxc, line 233: check previous comment

packages/nc-gui/components/virtual-cell/components/utils.ts
```diff
@@ -0,0 +1,10 @@
+import dayjs from 'dayjs'
+
+export const renderValue = (value: string | number | boolean) => {
+  if (!value || typeof value !== 'string') return value
+  return value.replace(/\b(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\+\d{2}:\d{2})\b/g, (d) => {
```
pranavxc, line 5: + part could be - as well, so you can use [+-] (same as previous comment)

packages/nc-gui/composables/useMultiSelect/index.ts
```diff
@@ -106,6 +115,41 @@ export function useMultiSelect(
           if (typeof textToCopy === 'object') {
             textToCopy = JSON.stringify(textToCopy)
           }
+
+          if (columnObj.uidt === UITypes.Formula) {
+            textToCopy = textToCopy.replace(/\b(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\+\d{2}:\d{2})\b/g, (d: string) => {
```
pranavxc, line 120: same as the previous comment.

And you can keep the regex in some modules and reuse it.

packages/nc-gui/composables/useMultiSelect/index.ts
```diff
@@ -106,6 +115,41 @@ export function useMultiSelect(
           if (typeof textToCopy === 'object') {
```
pranavxc, line 115: Above this, you can add an additional if condition to check content is a datejs instance/object and return. Just ignore my comment if it doesn't make much difference.

packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -1600,6 +1612,45 @@ class BaseModelSqlv2 {
...
+                [
+                  `${sanitize(alias || this.model.table_name)}.${
+                    column.column_name
+                  }`,
+                ],
+              );
+            break;
+          } else if (this.isPg) {
+            // if there is no timezone info, convert it to UTC
+            if (column.dt !== 'timestamp with time zone') {
```
pranavxc, line 1637: You can include `timestamptz` as well which is an alias for `timestamp with time zone`.

packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -3192,6 +3292,86 @@ class BaseModelSqlv2 {
...
+  private _convertDateFormat(
+    dateTimeColumns: Record<string, any>[],
+    d: Record<string, any>,
+    isXcdbBase: BoolType,
+  ) {
+    if (d) {
+      for (const col of dateTimeColumns) {
+        if (d[col.title]) {
+          let keepLocalTime = true;
+
```
pranavxc, line 3305: You can avoid deeply nested if/for

In the above scenario, it could be

```js
if(!d)
   return d;
for (const col of dateTimeColumns) {
  if (!d[col.title]) 
     continue;
 let keepLocalTime = true;
```

This way code will be more readable

packages/nocodb/src/db/formulav2/formulaQueryBuilderv2.ts
```diff
@@ -537,6 +537,33 @@ async function _formulaQueryBuilder(
...
+            return {
+              builder: knex.raw(
+                `CONVERT_TZ(??, @@GLOBAL.time_zone, '+00:00')`,
+                [col.column_name],
+              ),
+            };
+          };
+        } else if (
+          knex.clientType() === 'pg' &&
+          col.dt !== 'timestamp with time zone'
```
pranavxc, line 552: Include `timestamptz`

packages/nocodb/src/meta/meta.service.ts
```diff
@@ -1030,6 +1035,20 @@ export class MetaService {
...
+  private isMySQL(): boolean {
+    return (
+      this.connection.clientType() === 'mysql' ||
+      this.connection.clientType() === 'mysql2'
+    );
+  }
+
+  private now(): any {
+    if (this.isMySQL()) {
+      return dayjs().format('YYYY-MM-DD HH:mm:ss');
```
pranavxc, line 1047: Does it expect utc ? like - `dayjs(val).utc().format('YYYY-MM-DD HH:mm:ss')`

packages/nc-gui/components/virtual-cell/components/ListChildItems.vue
wingkwong: as spoken, this will be handled in another PR since it may test with other column types

PR #5608, opened 2023-05-07
feat: duplicate project and table

mertmit:
## Change Summary

- new background jobs queue
  - at import job migrated to new queue
- duplicate jobs for project and table
  - schema, data, views, filters, sorts and webhooks included

## Change type

## Additional information

This uses local nc-plugin so actions are failing.


dstala (commented):
- [ ] Duplicate table - after creating a copy, open new table by default. currently displayed table remains same, doesn't change.
- [x] Duplicate table - Order of views not retained
- [x] Duplicate table - Error observed
```
║   ---- !! JOB FAILED !! ----                                                                                                                               ║
║   name: duplicate-model                                                                                                                                    ║
║   id:1                                                                                                                                                     ║
║   error:TypeError (col.colOptions.formula.match(...) is not iterable (cannot read property null))                                                          ║
║                                                                                                                                                            ║
║   stack: TypeError: col.colOptions.formula.match(...) is not iterable (cannot read property null)                                                          ║
║       at ImportService.importModels (/Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/modules/jobs/export-import/import.service.ts:741:23)               ║
║       at DuplicateProcessor.duplicateModel (/Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/modules/jobs/export-import/duplicate.processor.ts:178:19)   ║
║       at QueueService.jobWrapper (/Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/modules/jobs/fallback-queue.service.ts:75:7)                          ║
║       at run (/Users/rajuudava/NocoDB/nocodb/packages/nocodb/node_modules/p-queue/dist/index.js:163:29) 
```
Use attached sqlite file & try to duplicate Sheet-1, with all configuration parameters enabled (default). 

[noco copy.db.zip](https://github.com/nocodb/nocodb/files/11417381/noco.copy.db.zip)

wingkwong (commented):
need more comments for the plugin changes

packages/nocodb/src/schema/swagger.json
```diff
@@ -18255,6 +18509,11 @@
             "maxLength": 128,
             "minLength": 1,
             "type": "string"
+          },
+          "status": {
```
wingkwong, line 18513: enum would be better (assuming status is a known set)

packages/nocodb/src/run/docker.ts
```diff
@@ -17,6 +17,13 @@ server.set('view engine', 'ejs');
 
 process.env[`DEBUG`] = 'xc*';
 
+// (async () => {
+//   await nocobuild(server);
+//   const httpServer = server.listen(process.env.PORT || 8080, async () => {
+//     console.log('Server started');
+//   });
+// })().catch((e) => console.log(e));
+
```
wingkwong, line 26: can be deleted

packages/nc-gui/pages/index/index/index.vue
```diff
@@ -74,6 +82,40 @@ const deleteProject = (project: ProjectType) => {
   })
 }
 
+const duplicateProject = (project: ProjectType) => {
```
wingkwong, line 85: for `status`, prefer enum rather than hardcoding the value.

packages/nc-gui/components/smartsheet/Grid.vue
```diff
@@ -756,7 +756,7 @@ const closeAddColumnDropdown = () => {
 const confirmDeleteRow = (row: number) => {
   Modal.confirm({
     title: `Do you want to delete this row?`,
-    wrapClassName: 'nc-modal-row-delete',
+    wrapClassName: 'nc-modal-attachment-delete',
```
wingkwong, line 759: incorrect merge? It should be `nc-modal-row-delete`.

packages/nc-gui/pages/index/index/index.vue
```diff
@@ -39,9 +39,17 @@ const filterQuery = ref('')
...
+
+const pageChange = (p: number) => {
+  activePage.value = p
+}
+
 const loadProjects = async () => {
+  const lastPage = activePage.value
   const response = await api.project.list({})
   projects.value = response.list
+  activePage.value = lastPage
```
wingkwong, line 52: would `activePage.value` be changed after project list? Besides, I think it would be better to name `lastActivePage` or `prevActivePage` as `lastPage` sounds like the last page of the table.

packages/nc-gui/components/dashboard/TreeView.vue
```diff
@@ -389,6 +389,34 @@ const setIcon = async (icon: string, table: TableType) => {
...
+
+const duplicateTable = async (table: TableType) => {
+  if (!table || !table.id || !table.project_id) return
+
+  const isOpen = ref(true)
+
+  const { close } = useDialog(resolveComponent('DlgTableDuplicate'), {
+    'modelValue': isOpen,
+    'table': table,
+    'onOk': async (jobData: { name: string; id: string }) => {
```
wingkwong, line 401: missing try catch error handling and $e - see duplicateProject for reference.

wingkwong (commented):
- some changes are done in `nocodb-legacy`. please double check.

packages/nc-gui/components/dashboard/TreeView.vue
mertmit: I moved error handling part to component (wrapping api call) as it should be on both

packages/nocodb/src/services/bulk-data-alias.service.ts
```diff
@@ -44,12 +44,22 @@ export class BulkDataAliasService {
     param: PathParams & {
       body: any;
       cookie: any;
+      chunkSize?: number;
+      foreign_key_checks?: boolean;
+      raw?: boolean;
```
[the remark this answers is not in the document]
mertmit: raw: this option is to skip alias - column name mapping, before/after hooks, data validations. this is meant for internal usage (duplicate, export/import)
foreign_key_checks: if this option is true we skip foreign key checks on that transaction, allowing us to insert belongsTo records in place as foreign keys without waiting linked record to be created which allow us to avoid one extra iteration of update calls over those records.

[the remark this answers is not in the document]
wingkwong: please include these infos in code :D

wingkwong:
> fileCreateByStream: this is to write file directly from a readable stream, in export we are streaming all data as CSV, with streaming we avoid buffering all data to memory.

> fileReadByStream: when we want to import exported data we again stream it to avoid buffering all data to memory.

> getDirectoryList: this is also used for import, to get available models within storage.

Got it. Please add these info to `packages/nc-plugin/src/lib/IStorageAdapterV2.ts` and mention they are only available in Local atm.

dstala (approved):

packages/nc-gui/pages/index/index/index.vue
```diff
@@ -122,7 +161,7 @@ const getProjectPrimary = (project: ProjectType) => {
 
 const customRow = (record: ProjectType) => ({
   onClick: async () => {
-    await navigateTo(`/nc/${record.id}`)
+    if (record.status !== 'job') await navigateTo(`/nc/${record.id}`)
```
wingkwong, line 164: enum

packages/nc-gui/pages/index/index/index.vue
```diff
@@ -260,7 +304,7 @@ const copyProjectMeta = async () => {
 
       <a-table-column key="id" :title="$t('labels.actions')" data-index="id">
         <template #default="{ text, record }">
-          <div class="flex items-center gap-2">
+          <div v-if="record.status !== 'job'" class="flex items-center gap-2">
```
wingkwong, line 307: enum

packages/nc-gui/plugins/jobs.ts
```diff
@@ -0,0 +1,100 @@
...
+  const { appInfo } = $(useGlobal())
+
+  let socket: Socket | null = null
+  let messageIndex = 0
+
+  const init = async (token: string) => {
+    try {
+      if (socket) socket.disconnect()
+
+      const url = new URL(appInfo.ncSiteUrl, window.location.href.split(/[?#]/)[0])
```
wingkwong, line 15: may give an example

wingkwong (commented):
not sure why Currency.vue change would be included in this PR. Something wrong with the merge?

wingkwong (commented):
**Duplicating a table twice would break the backend. Restart is required.** 

  - Steps to reproduce:
    - Create a table `TableA`
    - Duplicate `TableA`. `TableA copy` is created as expected
    - Duplicate `TableA` again. See the error. No action can be made in UI until restarting backend.
    - Expected behaviour: `TableA copy copy` or `TableA copy_1` would be created without any failure.
    
image
image


Maybe socket is not working in my local. Can someone confirm it?

pranavxc:
> **Duplicating a table twice would break the backend. Restart is required.**
> 
> * Steps to reproduce:
>   
>   * Create a table `TableA`
>   * Duplicate `TableA`. `TableA copy` is created as expected
>   * Duplicate `TableA` again. See the error. No action can be made in UI until restarting backend.
>   * Expected behaviour: `TableA copy copy` or `TableA copy_1` would be created without any failure.
> 
> image image
> 
> Maybe socket is not working in my local. Can someone confirm it?

Not able to reproduce, it works without any issue in my case.


https://github.com/nocodb/nocodb/assets/61551451/9be95a41-9b51-4e6c-8bc5-694df7d8fee9

PR #5609, opened 2023-05-07
fix: drop mm table on column delete

mertmit:
## Change Summary

We were only removing it from meta (as we only call Model.delete over it) which causing unnecessary mm tables to exist in database.

## Change type


dstala (changes requested):
A small UI fix required. see if it can go along with this PR
- On deleting m2m column, the junction table doesn't disappear from right project tree. It needs an explicit reload. Below screenshot is after removing m2m column. 

Screenshot 2023-05-08 at 8 18 24 AM

pranavxc (approved):

PR #5663, opened 2023-05-18
feat(lang): add column local dropdown text

amajakai14:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.
showing select option DataType in Local language instead of default value by default

before. 
before_localize
after. 
after_localize
  
### DIFF
--- a/packages/nc-gui/components/smartsheet/column/EditOrAdd.vue
+++ b/packages/nc-gui/components/smartsheet/column/EditOrAdd.vue
@@ -178,7 +178,13 @@ useEventListener('keydown', (e: KeyboardEvent) => {
             <a-select-option v-for="opt of uiTypesOptions" :key="opt.name" :value="opt.name" v-bind="validateInfos.uidt">
               <div class="flex gap-1 items-center">
                 <component :is="opt.icon" class="text-grey" />
-                {{ $t(`datatype.${opt.name}`) !== `datatype.${opt.name}` ? $t(`datatype.${opt.name}`) : opt.name }}
+                {{
+                  $i18n.locale === 'en'
+                    ? opt.name
+                    : $t(`datatype.${opt.name}`) !== `datatype.${opt.name}`
+                    ? $t(`datatype.${opt.name}`)
+                    : opt.name
+                }}
               </div>
             </a-select-option>
           </a-select>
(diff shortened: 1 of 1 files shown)


wingkwong (changes requested):
prefer the following way for better readability.

```vue
{{ renderColumnType(opt.name) }}
```

```ts
const renderColumnType = (columnType: string) => {
  // when no translation is found, the value of t(`key`) would be `key`
  if (t(`datatype.${columnType}`) !== `datatype.${columnType}`) {
    // if the target translation is available, then use it
    return t(`datatype.${columnType}`)
  }
  // else we use the default one (i.e. English)
  return columnType
}
```

amajakai14:
> prefer the following way for better readability.
> 
> ```
> {{ renderColumnType(opt.name) }}
> ```
> 
> ```ts
> const renderColumnType = (columnType: string) => {
>   // when no translation is found, the value of t(`key`) would be `key`
>   if (t(`datatype.${columnType}`) !== `datatype.${columnType}`) {
>     // if the target translation is available, then use it
>     return t(`datatype.${columnType}`)
>   }
>   // else we use the default one (i.e. English)
>   return columnType
> }
> ```

thanks for suggestion!

wingkwong (approved):

packages/nc-gui/components/smartsheet/column/EditOrAdd.vue
```diff
@@ -79,6 +79,8 @@ const uiTypesOptions = computed<typeof uiTypes>(() => {
 })
 
 const renderColumnType = (columnType: string) => {
+  if (locale.value === 'en') return columnType
```
wingkwong, line 82: please revert this change. no need to add that.

wingkwong (changes requested):

packages/nc-gui/components/smartsheet/column/EditOrAdd.vue
```diff
@@ -78,6 +78,18 @@ const uiTypesOptions = computed<typeof uiTypes>(() => {
   ]
 })
 
+const renderColumnType = (columnType: string) => {
+  if (locale.value === 'en') return columnType
```
wingkwong, line 82: I think this line is not required. If it doesn't match, it will return english one by default.

wingkwong (changes requested):

dstala:
@amajakai14 closing this for now as its an year old. Please rebase & create a new PR if issue is still relevant.

Issue #5125, opened 2023-02-16
🐛 Bug: SwaggerUI not working with latest release


3dprogramin:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

SwaggerUI page isn't working anymore, with the release of version 0.105.0
It does work with previous version 0.104.3

Clone nocodb-seed, make sure it's nocodb latest version 0.105.0.
`npm install`
`npm start`

Log into the dashboard and click `Swagger: Rest APIs` link.
Page returned is blank and doesn't load anything.

Taking a look at the requests, looks like the swagger js and css files are missing and 404 is returned:
- `/css/swagger-ui-bundle.4.5.2.min.css`
- `/js/swagger-ui-bundle.4.5.2.min.js`

Once again, it does work just fine with the previous version.

### Desired Behavior

check attachments

### Project Details

NocoDB used as docker: false
NocoDB version: 0.105.0
Databased used: sqlite3
Project was created by: new project, but with previous version
Database on which spreadsheet is created: pg
OS: linux
Node.js version: tested both v16.19.0 and v16.18.1
Database version: postgres:latest docker

### Attachments

![404-swagger](https://user-images.githubusercontent.com/36532753/219476284-b404e8c6-c651-49cf-8e6f-406a8fb42c42.png)
![click_swagger](https://user-images.githubusercontent.com/36532753/219476287-f0d25082-f751-44b6-8fda-eaa96a7420d5.png)
![working_previous_version](https://user-images.githubusercontent.com/36532753/219476288-bf863893-294e-463d-a7f6-b6e492250084.png)

shrsv:
I am facing this exact issue in the latest docker v`0.107.1`. Tried v`0.107.0` - the issue seems to persist.

Shahin-rmz:
Hello, same here

Nciso:
same, :)

joncutrer:
I am seeing this error when running noco docker, checked out tag version 0.107.1

wingkwong:
@Shahin-rmz @Nciso @joncutrer Can you guys try on the following version?

```
docker run -d -p 9999:8080 nocodb/nocodb-timely:0.107.1-pr-5693-20230522-0338
```

Shahin-rmz:
> are you guys using Docker or Docker Compose?


I am using docker, postgres version.

lhr0909:
[...]

PR #5693, opened 2023-05-22
fix(nocodb): incorrect public path for docker

wingkwong:
## Change Summary

- in docker, `process.cwd()` is `/usr/src/app` while the public directory is `/usr/src/app/docker/public`
- closes: #5125

## Change type

## Test/ Verification

image


lhr0909:
@wingkwong Hi, I am still seeing this issue on my deployment on FlyIO, but locally it was fine. I think it is because FlyIO uses non-docker container runtime under the hood. Is there a way we can use https://github.com/sindresorhus/is-inside-container instead of is-docker for the checks? This lib is written by the same author and references is-docker so it is compatible with both docker and non-docker container environments. Thanks!

Issue #5701, opened 2023-05-22
🐛 Bug: Metadata not in sync and breaking most of the features


Sov3rain:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Spin up an instance of Noco with a separate PostgresQL database for metadata.
2. Create another PostgresQL database to save your real data.
3. Add your "data" database to any project.
4. Start populating your database (create multiple tables, relations, etc). For this example, we have two tables
  - Organization (with a relation to many Products)
  - Product (with a relation to one Organization) 
5. Go to the "Team and Settings" menu, then "Data sources", check out the sync button: you metadata db is out of sync. Click the "sync metadata" button, you'll see on each of your table, under the "sync state" the message "Table removed".
6. Click on "sync now". An error message appears stating "The referenced row does not exist".
7. Return to the main interface.
8. Everything is broken at this point. It is missing some tables or all of them. I can't add rows, can't change filters. 

### Desired Behavior

Well, I don't know where to start. At this point I can't use this product as a whole, because every time I add a separate data source, metadata DB goes out of sync and it breaks everything.

### Project Details

Node: **v16.19.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.107.2**

Deployed on Railway.app

### Attachments

![image](https://github.com/nocodb/nocodb/assets/10755450/36418bf6-c454-4b0f-ad0f-eef0e56f3694)
![image](https://github.com/nocodb/nocodb/assets/10755450/b0abce3a-35bb-41cc-ba57-617f6f208ffe)
![image](https://github.com/nocodb/nocodb/assets/10755450/cbd03bc1-daca-4d66-8718-d0b5f7c99d6e)

Sov3rain:
After some digging, it may be related to #4442, after a redeploy, I'm in sync with the metadata db but I'm out of sync on the data db, which gives me errors on relations:
![image](https://github.com/nocodb/nocodb/assets/10755450/2c8c3650-676d-4e79-abe7-0d570c83d048)

o1lab:
To summarise the problem metadata db is showing 'Out of sync' when any change is made to data db. This needs to be corrected. @mertmit @pranavxc

Sov3rain:
Well, in the meantime, do you advise not using the sync feature? Because right now, clicking the "sync now" button is breaking everything.

wingkwong:
[...]

Issue #5568, opened 2023-05-02
🐛 Bug: model list cache ignores base_id


mertmit:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Model.list function caches based on project_id,
so when we call list tables api for base when cache enabled it always results with all the available bases in project.

### Desired Behavior

Base list tables api should only return models belonging to that base.

Opening this to discuss how to approach that issue, as storing models based on project_id might be more efficient if so we might filter cached response before returning.

### Project Details

develop

wingkwong:
We can include `base_id` in most cases. However, how do we deal with `xcVisibilityMetaGet` and `swaggerJson` where `base_id` is null?

wingkwong:
> We can include `base_id` in most cases. However, how do we deal with `xcVisibilityMetaGet` and `swaggerJson` where `base_id` is null?

After checking the code, those parts are fine since we've removed null value in cache key.

PR #5707, opened 2023-05-23
fix(nocodb): model list cache

wingkwong:
## Change Summary

Previously we only allowed one base so `base_id` wasn't included in the cache key. Since currently multiple bases are possible, hence this PR is to include it.

- closes: #5701
- closes: #5568

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/models/Model.ts
+++ b/packages/nocodb/src/models/Model.ts
@@ -138,7 +138,7 @@ export default class Model implements TableType {
 
     await NocoCache.appendToList(
       CacheScope.MODEL,
-      [projectId],
+      [projectId, baseId],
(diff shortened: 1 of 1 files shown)


dstala (approved):

wingkwong:
> image

> I think we miss base_id scope on Model delete, this might cause unwanted behavior.

@mertmit I think it's not related. Those two lines are to delete cache mapping either with id or title (i.e. `aliasOrId`) - which is used in `getByAliasOrId` only. We include `base_id` in listing cache only. We already have `deepDel` with `CHILD_TO_PARENT`. Hence, it's covered.

PR #5711, opened 2023-05-23
feat: worker container

mertmit:
## Change Summary

This PR provides a new environment variable to create instance as worker. Also there are some refactoring included on nc-config.
- Worker
  - Worker is a NocoDB instance without port binding and controllers. It will solely listen and process the jobs on jobs queue
  - To create a worker instance user should provide `NC_WORKER_CONTAINER` environment value as `true` -all the other inputs will be interpreted as false-
  - In order to use workers primary instance and worker have to use Redis (`NC_REDIS_URL`)
- NcConfig changes are mostly refactoring in order to separate helpers & class and make function names more self explanatory

## Architecture

architecture

## Change type

## Additional Information

I will provide an example guide on how to create a job using new job implementation soon.


Zerbaib (approved):

Zerbaib (approved):

wingkwong (commented):
Is `NC_WORKER_CONTAINER` just a boolean type (`true` / `false`)? If not, please show one example.

mertmit:
> Is `NC_WORKER_CONTAINER` just a boolean type (`true` / `false`)? If not, please show one example.

I pushed a new commit related to this.
It was based on presence before but clearly it can cause ambiguity and not a good practice.
So with new commit `NC_WORKER_CONTAINER` have to set to `true` in order to work, all the other values will be interpreted as false.

wingkwong:
> > Is `NC_WORKER_CONTAINER` just a boolean type (`true` / `false`)? If not, please show one example.
> 
> I pushed a new commit related to this. It was based on presence before but clearly it can cause ambiguity and not a good practice. So with new commit `NC_WORKER_CONTAINER` have to set to `true` in order to work, all the other values will be interpreted as false.

Ya I wanted to point out the common pitfall if the env var is a boolean type - if it is not defined, it would consider as `true` so we have to check the value instead, which you already included in your latest commit.

MauriRJUNLP:
Can you show an example of how it would look if I integrate it with docker-compose?

emre1e:
Slm Mert,

i tried this out. But nothing works to get a worker node working.

The Worker Node starts as worker but not excecuting single Job.

```
nc-exp:
    depends_on:
      root_db:
        condition: service_healthy
      redis:
        condition: service_started
    image: nocodb/nocodb-timely:0.111.4-pr-6375-20230914-0909
    ports:
      - "8080:8080"
      - "4000:4000"
    restart: always
    environment:
      NC_DB: "pg://root_db:5432?u=postgres&p=password&d=root_db"
      NC_REDIS_URL: "redis://redis:6379"
      DB_QUERY_LIMIT_MAX: "10000"
      NC_EXPORT_MAX_TIMEOUT: "100000"
    networks:
      - default
  worker:
    depends_on:
      root_db:
        condition: service_healthy
      redis:
        condition: service_started
    image: nocodb/nocodb-timely:0.111.4-pr-6375-20230914-0909
    restart: unless-stopped
    environment:
      NC_DB: "pg://root_db:5432?u=postgres&p=password&d=root_db"
      NC_WORKER_CONTAINER: "true"
      NC_REDIS_URL: "redis://redis:6379"
    networks:
      - default
```

PR #5730, opened 2023-05-24
fix: drop dtxp max length from column request schema

mertmit:
## Change Summary

Column dtxp property will be options for select columns which can be longer than 255 (StringOrNull type has 255 maxLength). We have no limit on option count, so it can exceed that limit which results with ajv schema error (TableReqType > NormalColumnRequestType > dtxp).

Note: it is already like this in ColumnType

## Change type


starbirdtech383:
LGTM

starbirdtech383 (approved):

PR #5764, opened 2023-05-30
fix: set refresh-token on login and correction in refresh-token based api

pranavxc:
## Change Summary

- [x] Set refresh-token on the login
- [x] Correction in refresh-token based token generation api 

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/controllers/users/users.controller.ts
+++ b/packages/nocodb/src/controllers/users/users.controller.ts
@@ -1,5 +1,3 @@
-import { promisify } from 'util';
-import { AuditOperationSubTypes, AuditOperationTypes } from 'nocodb-sdk';
 import {
   Body,
   Controller,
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/controllers/users/users.controller.ts
```diff
@@ -118,8 +65,9 @@ export class UsersController {
   ])
   @UseGuards(AuthGuard('local'))
   @HttpCode(200)
-  async signin(@Request() req) {
-    return this.usersService.login(req.user);
+  async signin(@Request() req, @Response() res) {
+    await this.setRefreshToken({ req, res });
+    res.json(this.usersService.login(req.user));
```
mertmit, line 70: UsersService login function is async and returns a promise (but it is synchronous so not necessary) which we don't await here. This cause empty response to returned by this endpoint, breaking the login process.

So we should drop await from UsersService login function.

packages/nocodb/src/controllers/users/users.controller.ts
```diff
@@ -136,22 +84,19 @@ export class UsersController {
   @Post(`/auth/google/genTokenByCode`)
   @HttpCode(200)
   @UseGuards(AuthGuard('google'))
-  async googleSignin(@Request() req) {
-    return this.usersService.login(req.user);
+  async googleSignin(@Request() req, @Response() res) {
+    await this.setRefreshToken({ req, res });
+    res.json(this.usersService.login(req.user));
```
mertmit, line 89: Possible to cause similar issue, see other comment.

mertmit (changes requested):

packages/nocodb/src/controllers/users/users.controller.ts
pranavxc: That method is synchronous so I removed `async` from the method

packages/nocodb/src/services/users/users.service.ts
```diff
@@ -486,9 +486,9 @@ export class UsersService {
     return this.login(user);
   }
 
-  async login(user: any) {
+  login(user: any) {
```
wingkwong, line 489: can we use `UserType` here?

packages/nocodb/src/controllers/users/users.controller.ts
```diff
@@ -269,4 +214,27 @@ export class UsersController {
       return res.status(400).json({ msg: e.message });
     }
   }
+
+  async setRefreshToken({ res, req }) {
+    const userId = req.user?.id;
+
+    if (!userId) return;
+
+    const user: any = await User.get(userId);
```
wingkwong, line 223: `UserType`?

packages/nc-gui/composables/useApi/interceptors.ts
```diff
@@ -69,6 +69,7 @@ export function addAxiosInterceptors(api: Api<any>) {
           await state.signOut()
           // todo: handle new user
 
+          debugger
```
wingkwong, line 72: remove

wingkwong (approved):

Issue #5798, opened 2023-06-01
🐛 Bug:  eyeSlash icon is not shown in form


DIWAKARKASHYAP:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

click on form and look the current behaviour of form
![image](https://github.com/nocodb/nocodb/assets/94478026/11feb2f3-d0f8-43ff-8581-4bf7dd752912)

### Desired Behavior

![image](https://github.com/nocodb/nocodb/assets/94478026/24d4e9f5-2d6d-47f4-b94e-dd173c324a2e)

### Project Details

form

PR #5799, opened 2023-06-01
 eyeSlash icon is not shown in form #5798

DIWAKARKASHYAP:
## Change Summary

- closes: https://github.com/nocodb/nocodb/issues/5798

## Change type

## Test/ Verification

Provide summary of changes.


DIWAKARKASHYAP:
please check issue #5798

o1lab:
Thanks for contributing. Please sign CLA above

DIWAKARKASHYAP:
> Thanks for contributing. Please sign CLA above

done

wingkwong (changes requested):
The expected behaviour is that the icon will only show when you hover a form entry. See below screenshot - I'm hovering the first entry.

image

By default, we set `opacity-0` so that all remove icons will be invisible at the beginning. When we hover `.nc-editable` block, the icon will be visible. Previously we were using `MdiEyeOffOutline` so the scss would work. However, at some point we changed to `iconMap.eyeSlash` causing the reported issue.

Therefore, removing `opacity-0` is not a correct fix. Instead, you just need to update the scss part in the same file.

https://github.com/nocodb/nocodb/blob/7550bdde93fcb290aae4e96c6d325da346e34547/packages/nc-gui/components/smartsheet/Form.vue#L839-L843

Simply change to the following - then it should work

```scss
.nc-editable:hover {
  :deep(.nc-field-remove-icon) {
    @apply opacity-100;
  }
}
```

DIWAKARKASHYAP:
got it , 
thanks , i will make the changes

wingkwong (approved):

PR #5807, opened 2023-06-02
feat: Drag n drop LTAR and Lookup creation

pranavxc:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/components/dashboard/TreeView.vue
+++ b/packages/nc-gui/components/dashboard/TreeView.vue
@@ -192,6 +192,8 @@ const initSortable = (el: Element) => {
         JSON.stringify({
           id: dragEl.dataset.id,
           title: dragEl.dataset.title,
+          type: dragEl.dataset.type,
+          baseId: dragEl.dataset.baseId,
         }),
       )
     },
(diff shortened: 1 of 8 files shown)


dstala:
@pranavxc 
one observation- 
while trying to add LTAR/LookUps via this method, i unintentionally re-ordered table entries in left tree view. 
not a deal breaker, but if we can find a way to handle this UX - it would be great.

pranavxc:
> > > @pranavxc one observation- while trying to add LTAR/LookUps via this method, i unintentionally re-ordered table entries in left tree view. not a deal breaker, but if we can find a way to handle this UX - it would be great.
> > 
> > 
> > If the user is not dropping the table inside Treeview, we should avoid reordering. This way we can take care of the issue.
> 
> sounds good. we can try.

@dstala Done 


https://github.com/nocodb/nocodb/assets/61551451/237e8322-e020-4881-b8bc-a29aab4924ce

dstala (changes requested):
- [ ] Disable drag-drop of SQL view on table
Screenshot 2023-06-07 at 7 41 12 PM

- [ ]Identify drag-drop across projects & block them appropriately
Currently we see a modal similar to what we saw above in case of SQL view; and subsequently fails with `table does not exist`

mertmit (approved):

Issue #5803, opened 2023-06-02
UX : Delete row related


dstala:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

- `Delete row` pops up confirmation modal
- `Delete selected row` doesn't

### Desired Behavior

Since we have Undo-redo working for delete row, we can get rid of confirmation modal even for `Delete Row`

Additionally, `Delete Selected Row` should be in disabled state by default in context menu (grey text?). It should get enabled only when user selects some rows on the grid.

### Project Details

Node: **v16.19.1**
Arch: **arm64**
Platform: **linux**
Docker: **true**
RootDB: **sqlite3**
PackageVersion: **0.108.1**

dstala:
PW tests will fail. Assign to me once your changes are completed on this.

PR #5813, opened 2023-06-03
revise: cell context menu

mertmit:
## Change Summary

- Show only applicable options when multiple cells are selected
- Delete selected range of rows

## Change type


packages/nc-gui/composables/useViewData.ts
```diff
@@ -615,6 +616,82 @@ export function useViewData(
...
+          const id = meta?.value?.columns
+            ?.filter((c) => c.pk)
+            .map((c) => rowObj[c.title as string])
+            .join('___')
+
+          const successfulDeletion = await deleteRowById(id as string)
+          if (!successfulDeletion) {
+            continue
+          }
+          removedRowsData.push({ id, row: clone(formattedData.value[row]), rowIndex: row })
```
pranavxc, line 641: One doubt related to undo-redo(not related to the current PR) - on deleting a row are we keeping the entire data including the value of the hidden column or just the value visible in the view?

pranavxc (approved):

pranavxc:
re #5803

Issue #5804, opened 2023-06-02
🐛 Bug: (Swagger UI) Unable to render this definition


hialvaro:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Visit the Swagger API docs for a project from `APIs & Support` Help Center
2. See the error:

![](https://i.imgur.com/YGlrZlT.png)

### Desired Behavior

Have a functioning API Docs.

### Project Details

Node: **v16.19.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **sqlite3**
PackageVersion: **0.108.1**

pranavxc:
It expects an access token in your browser localstorage since `swagger.json` endpoint is not public now. 

To access the page do the following steps

1. Open your dashboard URL and login with your credential
2. Now you can access the Swagger UI enpoint

If still facing the issue check your browser network tab and share the `swagger.json` endpoint response details.

dstala:
https://github.com/nocodb/nocodb/issues/5804#issuecomment-1578750458

PR #5840, opened 2023-06-07
enhancement: After successful login redirect to previous page

pranavxc:
## Change Summary

- After successful login redirect back to the previous page
- If the token is not present while accessing Swagger UI redirects to sign in page and after signing redirects back to Swagger UI.

re #5804

## Change type

## Test/ Verification

Provide summary of changes.

## TODO 

- Refactor and remove duplicate code
- Handling hash-based and full URL

### DIFF
--- a/packages/nc-gui/composables/useApi/interceptors.ts
+++ b/packages/nc-gui/composables/useApi/interceptors.ts
@@ -1,12 +1,13 @@
 import type { Api } from 'nocodb-sdk'
 import { navigateTo, useGlobal, useRouter } from '#imports'
+import { Router } from 'vue-router'
 
 const DbNotFoundMsg = 'Database config not found'
 
 export function addAxiosInterceptors(api: Api<any>) {
   const state = useGlobal()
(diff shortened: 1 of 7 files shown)


dstala:
Few observations. 
- On first attempt to access swagger docs, we are redirecting to sign-in page. And there is no indication specifying that its additional sign-in required for swagger. It would feel, as if we were logged out by trying to access swagger docs
- After sign-in on the new page, it goes back to project page - doesn't open swagger docs automatically. swagger docs open after another access from the project menu

Ideal flow IMO would be
- on accessing swagger docs, open another tab - use a pop up modal that specifies password required for swagger access and to take only password as input (as mac does when we try to change some system configurations) & redirect to swagger page.

pranavxc:
> After sign-in on the new page, it goes back to project page - doesn't open swagger docs automatically. swagger docs open after another access from the project menu

Because `localhost:8080` serves an old version of UI, I tested it with the latest build(taken locally), and worked as expected.

PR #5896, opened 2023-06-14
feat: vertical fill using handle

mertmit:
## Change Summary

- fill handle on selected range corner
- allow filling records vertically using handle

## Change type


dstala (commented):
https://nocodb.getoutline.com/doc/fill-handle-kihFrhxh8Q

PR #5918, opened 2023-06-17
fix: various fixes for UI

mertmit:
## Change Summary

- MultiSelect order when cell is active
- Improve hiding context menu (avoid showing changed menu before closing)
- Revised mouse event handling
  - moved click part to mousedown/up (to avoid handling same part twice)
  - scroll to cell on mouseover while selecting
  - disable edit mode on changing active cell
  - avoid setting `isMouseDown` flag if edit enabled (some cells shrink after edit mode)
 
## Change type

### DIFF
--- a/packages/nc-gui/components/cell/MultiSelect.vue
+++ b/packages/nc-gui/components/cell/MultiSelect.vue
@@ -316,10 +316,10 @@ const handleClose = (e: MouseEvent) => {
 useEventListener(document, 'click', handleClose, true)
 
 const selectedOpts = computed(() => {
-  return options.value.reduce<(SelectOptionType & { index: number })[]>((selectedOptions, option) => {
-    const index = vModel.value.indexOf(option.value!)
-    if (index !== -1) {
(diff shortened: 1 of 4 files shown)


packages/nc-gui/components/cell/MultiSelect.vue
```diff
@@ -316,10 +316,10 @@ const handleClose = (e: MouseEvent) => {
...
 
 const selectedOpts = computed(() => {
-  return options.value.reduce<(SelectOptionType & { index: number })[]>((selectedOptions, option) => {
-    const index = vModel.value.indexOf(option.value!)
-    if (index !== -1) {
-      selectedOptions.push({ ...option, index })
+  return vModel.value.reduce<SelectOptionType[]>((selectedOptions, option) => {
+    const fnd = options.value.find((o) => o.value === option)
+    if (fnd) {
+      selectedOptions.push(fnd)
```
wingkwong, line 322: suggest to use a better name instead of `fnd`. maybe `selectedOption`

Issue #5045, opened 2023-02-06
🔦 Feature: bulk / batch update column for multiple rows


davidpfahler:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

As a user, I would like to be able to batch update one column of several selected rows at once. For example, if I had a `Todos` table with columns `title` and `status`, I would like to be able to select several rows and change their `status` from `to do` to `done`.

### Suggested Solution

I do not want to presume what the best UI/UX would be for this feature, but one idea could be to have an Excel-style "Extend series or fill"

o1lab:
We will prioritise this after our release next week.

max-santiago:
Are there any updates on this?

wingkwong:
Currently you can achieve this by using Bulk Update API with conditions. The below example would update `status` to `done` for those records with ID greater than 5.

```
curl -X 'PATCH' \
http://localhost:8080/api/v1/db/data/bulk/noco/foo/bar/all?where=(Id,gt,5) \
-H "Content-Type: application/json" \
-H 'accept: application/json' \
-H 'xc-auth: <REDACTED>' \
--data '{ "status": "done" }'
```

davidpfahler:
@wingkwong Thanks, but this is impractical (at least for me) to link several entries at once to another entry (in my case link time entries to an invoice).

davidpfahler:
It seem this was not included in release `0.106.0` correct?

wingkwong:
@davidpfahler so far I don't see there s any PR linking to this issue. The batch-update on UI is not available in 0.106.0.

davidpfahler:
@wingkwong I was just asking because @o1lab said it would be in the next release. I understand plans change, it's just that it would be a really useful feature.

dstala:
Bulk update from UI is available in NocoDB hosted solution. 

For a free technical preview, visit www.nocodb.com

aks-cadesign:
when will it be available on the self hosted solution? Unfortunately, without this feature, NocoDB is completely useless for a lot of use-cases.

PR #5928, opened 2023-06-19
feat: bulk update records using UI

mertmit:
## Change Summary

Re #5045
- This PR allows bulk update on both selected records or all records in a view.
  - Select records (or all records in case you want to update all in current view) then right click

## Change type

image
image
image
image


packages/nc-gui/components/dlg/BulkUpdate.vue
```diff
@@ -0,0 +1,444 @@
...
+        <div v-if="updateMode === BulkUpdateMode.ALL" class="flex items-center">
+          <component :is="iconMap.contentSaveExit" class="mr-1" />
+          Bulk Update All
+        </div>
+        <div v-if="updateMode === BulkUpdateMode.SELECTED" class="flex items-center">
+          <component :is="iconMap.contentSaveStay" class="mr-1" />
+          Bulk Update Selected
+        </div>
+      </a-dropdown-button>
+      -->
```
wingkwong, line 219: if these are not in use, suggest to remove.

packages/nc-gui/components/dlg/BulkUpdate.vue
```diff
@@ -0,0 +1,444 @@
...
+      <a-button v-if="props.selectedAllRecords" class="nc-bulk-update-save-btn" type="primary" @click="save">
+        <div class="flex items-center">
+          <component :is="iconMap.contentSaveExit" class="mr-1" />
+          Bulk Update All
```
wingkwong, line 223: i18n or add a TODO tag. apply to other texts.

dstala (changes requested):
- [x] [**Marked by Naveen as not supported**] Sakila ext DB PG table. Actor table. Select all - bulk update. Unable to `undo`. Where as Undo works for smaller subset of bulk update.
- [ ] Good to show all the columns in bulk update modal? I would prefer to see only visible fields (not hidden). Discuss.
- [x] UI background needs to be sorted out when we have many columns to update in bulk modal
Screenshot 2023-06-20 at 11 27 40 AM

- [x] Bulk update when a filter is enabled threw error (issue not noticed in secondary view)
Screenshot 2023-06-20 at 11 29 53 AM

dstala (approved):

dstala:
- [ ] Long text : let us display bigger block on loading it into bulk upload form instead of toggling to bigger block (text area) on click. This way- clicking on next fields after updating long text doesn't realign blocks
- [ ] Tab doesn't work when navigating between fields in upload form
- [ ] Checkbox, on first load is not aligned (somehow the square box appears at the bottom of circle). I am using chrome.
  
Screenshot 2023-06-23 at 5 07 39 PM

- [ ] Date picker (or time/year picker) : after opening it - escape key should only close picker. Subsequent escape should close Bulk update modal
- [ ] Let us swap close icon with hide icon (eye) in form columns. Hide icon should appear only on hovering over that particular column; otherwise it remains hidden (as in form view). Currently form view hide icon is tad bigger. You can keep reduce its size a bit.
- [ ] Consider removing border (my opinion- ignore if you do not agree) to make it look similar to expanded form.
Screenshot 2023-06-23 at 5 11 39 PM

PR #5954, opened 2023-06-24
fix: dynamic batch based on byte size

mertmit:
## Change Summary

This should fix heap out of memory issues on import

## Change type

### DIFF
--- a/packages/nocodb/src/modules/jobs/jobs/at-import/helpers/readAndProcessData.ts
+++ b/packages/nocodb/src/modules/jobs/jobs/at-import/helpers/readAndProcessData.ts
@@ -148,6 +148,7 @@ export async function importData({
                   tableName: table.title,
                   body: insertArray,
                   cookie: {},
+                  skip_hooks: true,
                 });
 
                 logBasic(
@@ -183,6 +184,7 @@ export async function importData({
             tableName: table.title,
             body: tempData,
             cookie: {},
+            skip_hooks: true,
           });
 
           logBasic(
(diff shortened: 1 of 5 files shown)


packages/nocodb/src/modules/jobs/jobs/at-import/helpers/readAndProcessData.ts
```diff
@@ -1,14 +1,15 @@
...
 import type { BulkDataAliasService } from '../../../../../services/bulk-data-alias.service';
 import type { TablesService } from '../../../../../services/tables.service';
 // @ts-ignore
 import type { AirtableBase } from 'airtable/lib/airtable_base';
 import type { TableType } from 'nocodb-sdk';
 
-const BULK_DATA_BATCH_SIZE = 500;
-const ASSOC_BULK_DATA_BATCH_SIZE = 1000;
+const BULK_DATA_BATCH_COUNT = 100; // check size for every 100 records
+const BULK_DATA_BATCH_SIZE = 102400; // in bytes
```
o1lab, line 12: 100 * 1024 

will be easy to read - 100 kb

packages/nocodb/src/modules/jobs/jobs/at-import/helpers/readAndProcessData.ts
```diff
@@ -300,26 +302,31 @@ export async function importLTARData({
...
-              await services.bulkDataService.bulkDataInsert({
-                projectName,
-                tableName: assocMeta.modelMeta.title,
-                body: insertArray,
-                cookie: {},
-              });
-
-              importedCount += insertArray.length;
-              insertArray = [];
+            if (assocTableData.length % BULK_DATA_BATCH_COUNT === 0) {
```
o1lab, line 305: Does line 305 to 329 repeating ? if so let make it a function (DRY)

dstala (approved):

PR #6073, opened 2023-07-22
Fixed #5523 : Labeling columns with special characters causes form to break

### LINKED ISSUE #5523: 🐛 Bug: Labeling columns with special characters causes form to break

### Steps to reproduce ?

1. Create a sheet
2. Label column "?@% ++ ()"
3. Create form
4. Submit form
5. Get error the "column on the sheet does not exist" or "sql query error"

### Desired Behavior

Columns should be able to have special characters.  Or if columns shouldn't for technical reasons it should throw an error preventing the user from naming them in this way.

### Project Details

Node: **v18.14.1**
Arch: **x64**
Platform: **win32**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.105.3**

### Attachments

_No response_

shashankvish0010:
## Change Summary

Added a check for special characters in the field name using a regular expression. If the field name contains special characters, a warning message will be shown to the user, and the column creation will be canceled.
fixed #5523

## Change type

## Test/ Verification

To test and verify the fix:

Create a sheet
Label column "?@% ++ ()"
Save the column
Warning Message will popup: Special characters are not allowed in the column name.

Screenshot (253)

### DIFF
--- a/packages/nc-gui/composables/useColumnCreateStore.ts
+++ b/packages/nc-gui/composables/useColumnCreateStore.ts
@@ -101,6 +101,14 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
               })
             },
           },
+          {
+            validator: (rule: any, value: any) => {
+              if (/[!@#$%^&*(),.?":{}|<>]/.test(value)) {
+                return Promise.reject(new Error('Special characters are not allowed in the column name.'))
+              }
+              return Promise.resolve()
+            },
+          },
           fieldLengthValidator(project.value?.bases?.[0].type || ClientType.MYSQL),
         ],
         uidt: [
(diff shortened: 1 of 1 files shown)


packages/nc-gui/composables/useColumnCreateStore.ts
```diff
@@ -215,6 +215,11 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
     const addOrUpdate = async (onSuccess: () => void, columnPosition?: Pick<ColumnReqType, 'column_order'>) => {
       try {
         if (!(await validate())) return
+        const specialCharsRegex = /[!@#$%^&*(),.?":{}|<>]/
+        if (specialCharsRegex.test(formState._rawValue.column_name)) {
+          message.warning('Special characters are not allowed in the column name.')
+          return
+        }
```
wingkwong, line 222: I'd prefer to put the logic in `validators` (see L78) so that the invalid case will return in L217. Also please think about the backend side, like I can still input special characters using API.

wingkwong (changes requested):

dstala:
Closing this thread as associated issue is closed now

Issue #5256, opened 2023-03-06
Enhancement : better column datatype mapping during import


dstala:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Current
![Screenshot 2023-03-06 at 11 49 44 AM](https://user-images.githubusercontent.com/86527202/223040507-5ca751d3-eaed-4cfd-a542-510144038c9e.png)

Expected
![Screenshot 2023-03-06 at 11 50 30 AM](https://user-images.githubusercontent.com/86527202/223040514-26f06dd6-609d-4cc8-9fc7-b4602eebcd9c.png)

### Desired Behavior

column name often can indicate datatype? can we improvise?

### Project Details

Node: **v16.14.0**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
RootDB: **pg**
PackageVersion: **0.105.3**

PR #6077, opened 2023-07-24
Enhancement #5256 : better column datatype mapping during import

shashankvish0010:
## Change Summary

I made improvements to the code in the CSVTemplateAdapter class to enhance the detection of data types and improve template generation.
The changes include:
- Added additional checks for detecting phone numbers as UITypes.PhoneNumber.
- Modified conditions for detecting email and URL types as UITypes.Email and UITypes.URL, respectively.
- Implemented a length-based decision to detect long texts as UITypes.LongText specific for postal addresses or specific fields having a length more than 20 is considered as LongText, as LongText are detected when the length > 255 as per parserHelpers.
- enhanced #5256

## Change type

## Test/ Verification
To test and verify the fix:

- Import a CSV file
- And select the Auto-Select Field Types
- Import the file, and you will see the correct field types. 

Screenshot (258)
Screenshot (259)

### DIFF
--- a/packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
+++ b/packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
@@ -83,7 +83,7 @@ export default class CSVTemplateAdapter {
   detectInitialUidt(v: string) {
     const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/
     const urlRegex = /^(ftp|http|https):\/\/[^ "]+$/
-    const phoneRegex = /\b(?:\+?\d{1,3}(?:[\s-]+\(\d+\))?)?[\s-]*(?:\d[\s-]*){6,14}\b/
+    const phoneRegex = /\+(9[976]\d|8[987530]\d|6[987]\d|5[90]\d|42\d|3[875]\d|2[98654321]\d|9[8543210]|8[6421]|6[6543210]|5[87654321]|4[987654310]|3[9643210]|2[70]|7|1)\d{1,14}$/
     if (phoneRegex.test(v)) return UITypes.PhoneNumber
     if (emailRegex.test(v)) return UITypes.Email
     if (urlRegex.test(v)) return UITypes.URL
--- a/packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
+++ b/packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
@@ -81,6 +81,17 @@ export default class CSVTemplateAdapter {
   }
 
   detectInitialUidt(v: string) {
+    const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/
+    const urlRegex = /^(ftp|http|https):\/\/[^ "]+$/
+    const phoneRegex = /\b(?:\+?\d{1,3}(?:[\s-]+\(\d+\))?)?[\s-]*(?:\d[\s-]*){6,14}\b/
+    if (phoneRegex.test(v)) return UITypes.PhoneNumber
+    if (emailRegex.test(v)) return UITypes.Email
+    if (urlRegex.test(v)) return UITypes.URL
+    if (!isNaN(Number(v)) && !isNaN(parseFloat(v))) return UITypes.Number
+    if (validateDateWithUnknownFormat(v)) return UITypes.DateTime
+    if (['true', 'True', 'false', 'False', '1', '0', 'T', 'F', 'Y', 'N'].includes(v)) return UITypes.Checkbox
+    if (v.length > 20) return UITypes.LongText
+    return UITypes.SingleLineText
     if (!isNaN(Number(v)) && !isNaN(parseFloat(v))) return UITypes.Number
     if (validateDateWithUnknownFormat(v)) return UITypes.DateTime
     if (['true', 'True', 'false', 'False', '1', '0', 'T', 'F', 'Y', 'N'].includes(v)) return UITypes.Checkbox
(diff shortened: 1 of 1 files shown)


packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
```diff
@@ -81,9 +81,16 @@ export default class CSVTemplateAdapter {
   }
 
   detectInitialUidt(v: string) {
+    const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/
+    const urlRegex = /^(ftp|http|https):\/\/[^ "]+$/
+    const phoneRegex = /\b(?:\+?\d{1,3}(?:[\s-]+\(\d+\))?)?[\s-]*(?:\d[\s-]*){6,14}\b/
```
pranavxc, line 86: This will treat any string containing a phone number as Phonenumber type.

For example:

```js
/\b(?:\+?\d{1,3}(?:[\s-]+\(\d+\))?)?[\s-]*(?:\d[\s-]*){6,14}\b/.test('abc +911234567890 abc')
```

pranavxc: Replace word boundary with start and end anchors

```js
/^(?:\+?\d{1,3}(?:[\s-]+\(\d+\))?)?[\s-]*(?:\d[\s-]*){6,14}$/
```

packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
```diff
@@ -81,9 +81,16 @@ export default class CSVTemplateAdapter {
   }
 
   detectInitialUidt(v: string) {
+    const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/
+    const urlRegex = /^(ftp|http|https):\/\/[^ "]+$/
+    const phoneRegex = /\+(9[976]\d|8[987530]\d|6[987]\d|5[90]\d|42\d|3[875]\d|2[98654321]\d|9[8543210]|8[6421]|6[6543210]|5[87654321]|4[987654310]|3[9643210]|2[70]|7|1)\d{1,14}$/
```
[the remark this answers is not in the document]
shashankvish0010: > highly doubt that

on what ?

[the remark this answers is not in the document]
wingkwong: phoneRegex. Tried several numbers on https://regex101.com/. None works.

packages/nc-gui/utils/parsers/CSVTemplateAdapter.ts
```diff
@@ -83,8 +83,9 @@ export default class CSVTemplateAdapter {
   detectInitialUidt(v: string) {
     const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/
     const urlRegex = /^(ftp|http|https):\/\/[^ "]+$/
-    const phoneRegex = /\+(9[976]\d|8[987530]\d|6[987]\d|5[90]\d|42\d|3[875]\d|2[98654321]\d|9[8543210]|8[6421]|6[6543210]|5[87654321]|4[987654310]|3[9643210]|2[70]|7|1)\d{1,14}$/
-    if (phoneRegex.test(v)) return UITypes.PhoneNumber
+    const phoneRegex = /^\+(?:[0-9]?){1,3}[-.\s]?(?:\(\d+\)|[0-9]+[-.\s]?)*[0-9]+$/
+    const cleanedV = v.replace(/^'+/, '')
```
wingkwong, line 87: why we need this line? can you provide some examples?

shashankvish0010: Sure, so when I used only the above phoneRegex, without the words /b/, I was not getting the desired result as phoneNumber from phoneRegex I found out that I was getting the data in v as '+918967453423 and as per regex it's not a valid phoneNumber as it starts with an extra ' in starting so I added this line to exactly remove the extra ' from v and check only the cleanedV in phoneNumber if its a phoneNumber it would be returned if not normal cases would be executed

wingkwong: Why `'` specifically?  

Besides, your regex doesn't work with following phone numbers.

```
(877) 359-6695
800-692-7753
0365987456
```

shashankvish0010: > Why `'` specifically?
> 
> Besides, your regex doesn't work with following phone numbers.
> 
> ```
> (877) 359-6695
> 800-692-7753
> 0365987456
> ```

as phone regex is treated as a string, but when I checked the data (v), I found out that an extra ' is there at the start of any phone number, which makes the regex pattern different from that of the phone number, and just by removing that extra ' at the time of checking the cases, it works fine.

wingkwong: Can you provide your test csv? Also have you tested the above patterns? I was using https://regex101.com/ to verify your regex.

shashankvish0010: > 
[test.csv](https://github.com/nocodb/nocodb/files/12305887/test.csv)

Sure, yes, I tested it on the website you shared to verify the regex and got a positive response.

wingkwong: Failing

```
1-(123) 456-7890
1 (123) 456 7890
+1-222-333-4444
1-(123)-456-7890
```

dstala:
Import mechanism is altered now. Default mapping to Single line text always

Issue #5765, opened 2023-05-30
🐛 Bug: form view do not extent text area


hodanli:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. create text field
2. create form view
3. create share link
4. enter long text to the form using shared view

## Expected behavior

text ares should expand.

### Project Details

deployed on digitalocean server using docker

### Attachments

Screenshots

![image](https://github.com/nocobase/nocobase/assets/6535865/3b2a8a4d-b691-4262-9883-856247021a67)

ShubhamPatilsd:
Will be reserving this for students at [CodeDay Labs](https://labs.codeday.org/) to be working on in our summer cohort.

Mercywanjohi:
> Will be reserving this for students at [CodeDay Labs](https://labs.codeday.org/) to be working on in our summer cohort.

Claiming this issue, will be working with @PyJim for the CodeDay Labs internship.

wingkwong:
fixed in #6121

PR #6089, opened 2023-07-28
Fixed: Form view do not extent text area

PyJim:
## Change Summary

Issue #5765 in Nocodb was resolved by addressing a problem with the 'nc-form-after-submit-msg' textarea. Previously, the textarea was not extending correctly, causing text overflow. To fix this, We added height: auto; and overflow: auto; to the nc-input.nc-cell-longtext class.. This solution ensures that users can now enter and view large messages within the textarea without any issues.

## Change type

## Test / Verification

To test and verify:

create text field
create form view
create share link
enter long text to the form using shared view

textareaTEST


pranavxc (changes requested):
Please exclude the `package-lock.json` if you are not adding/removing any dependency.

PyJim:
package-lock.json excluded.

wingkwong (changes requested):
@PyJim You are not excluding `package.json` but deleting, which is not expected. The expected file changed is `packages/nc-gui/pages/[projectType]/form/[viewId]/index.vue` only.

PyJim:
Okay

PyJim:
Please how do I exclude the package-lock.json file. Wanted to use the .gitignore file but it cannot be edited.

PR #6091, opened 2023-07-28
Fixed: form view do not extent text area

PyJim:
## Change Summary

An issue in Nocodb was addressed, specifically related to the 'nc-form-after-submit-msg' textarea. Previously, the textarea was not extending correctly, resulting in text overflow. To resolve this problem, we added overflow: auto; and height: auto; properties to the .nc-cell-longtext class. As a result, users can now comfortably enter and view large messages within the textarea. This fix was implemented in response to the reported issue with the number #5765.

## Change type

## Test/ Verification
1. create text field
2. create form view
3. create share link
4. enter long text to the form using shared view

textareaTEST

### DIFF
--- a/packages/nc-gui/components/smartsheet/Form.vue
+++ b/packages/nc-gui/components/smartsheet/Form.vue
@@ -845,7 +845,7 @@ watch(view, (nextView) => {
 }
 
 .nc-input {
-  @apply appearance-none w-full !bg-white rounded px-2 py-2 my-2 border-solid border-1 border-primary border-opacity-50;
+  @apply appearance-none w-full h-auto min-h-40px !bg-white rounded px-2 py-2 my-2 border-solid border-1 border-primary border-opacity-50;
 
   :deep(input) {
     @apply !px-1;
--- a/packages/nc-gui/pages/[projectType]/form/[viewId]/index.vue
+++ b/packages/nc-gui/pages/[projectType]/form/[viewId]/index.vue
@@ -122,6 +122,8 @@ p {
 
         &.nc-cell-longtext {
           @apply !p-0.5rem h-auto;
+          word-break: break-all;
+          white-space: pre-line;
         }
 
         textarea {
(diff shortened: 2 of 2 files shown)


wingkwong (commented):
The behaviour doesn't match with the normal form. Please double check. Also please use utility classes instead of CSS.

### Case 1: when the field is not focused

**Normal Form**

image

**Shared Form**

image

### Case 2: when the field is focused

**Normal Form**

image

**Shared Form**

image

PyJim:
Made a few changes

wingkwong (changes requested):
Your change breaks the basic form. Please verify both form first.

image

PyJim:
Alright

PyJim:
With basic form do you mean the unshared form?

wingkwong:
@PyJim Yes. Basic Form View. You can type something long and un-focus the field. Then you should see the following

image

PyJim:
Oh yeah I had not considered that. Just worked on that as well

PyJim:
Oh yeah I had not considered that. Just worked on that as well

wingkwong (changes requested):
Un-focused input on basic form view not aligned with shared form view - with same content. Please verify all the cases first.

image

PyJim:
Please can I get the content sample you used so I use that for testing?

PyJim:
Also, based on the content, which one is preferable?

wingkwong:
You can just randomly type something with multiple lines.

```
wqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwewqewq eq
wqeqwe
qwe
```

PyJim:
Should be okay now

wingkwong (changes requested):
Now they are both aligned. However, the expected behaviour is exactly opposite. 

### Expected behaviour

when the field is focused:
image

when the field is not focused:
image

### Current behaviour in this PR


when the field is focused:
image

when the field is not focused:
image

wingkwong:
Closing - as it has been fixed in #6121.

PR #6134, opened 2023-08-06
chore: Skip unnecessary computations based on the fields of dbRow.

willnewii:
## Change Summary

When there is a table with 40 fields and 1000 rows of data, it will require performing 40*1000 times (excluding system fields, of course). However, if it's based on a view or specified fields, can these unnecessary operations be skipped?

## Change type

- [x] chore: Skip unnecessary computations based on the fields of dbRow.

### DIFF
--- a/packages/nocodb/src/modules/datas/helpers.ts
+++ b/packages/nocodb/src/modules/datas/helpers.ts
@@ -273,6 +273,8 @@ export async function getDbRows(param: {
 
       for (const column of view.model.columns) {
         if (isSystemColumn(column) && !view.show_system_fields) continue;
+        // eslint-disable-next-line no-prototype-builtins
+        if (!dbRow.hasOwnProperty(column.title)) continue;
         dbRow[column.title] = await serializeCellValue({
           value: row[column.title],
           column,
--- a/packages/nocodb/src/modules/datas/helpers.ts
+++ b/packages/nocodb/src/modules/datas/helpers.ts
@@ -268,13 +268,15 @@ export async function getDbRows(param: {
       break;
     }
 
+    const needSerializeColumns = view.model.columns.filter((column) => {
+      // eslint-disable-next-line no-prototype-builtins
+      return rows[0].hasOwnProperty(column.title);
+    });
+
     for (const row of rows) {
       const dbRow = { ...row };
 
-      for (const column of view.model.columns) {
-        if (isSystemColumn(column) && !view.show_system_fields) continue;
-        // eslint-disable-next-line no-prototype-builtins
-        if (!dbRow.hasOwnProperty(column.title)) continue;
+      for (const column of needSerializeColumns) {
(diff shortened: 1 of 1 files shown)


wingkwong (commented):
Can you show an example and explain how it helps in details? I don't understand what you are trying solve here. Also in what scenario `!dbRow.hasOwnProperty(column.title)` would be `true`?

willnewii:
In the current scenario, with a table comprising 40 non-system fields and 1000 records, the serializeCellValue function will be invoked 40 * 1000 times. 
when utilizing the export/excel API and specifying the fields parameter, the dbRow object doesn't encompass all the fields. By utilizing '!dbRow.hasOwnProperty(column.title)', unnecessary fields that don't need to be returned can be ignored, thereby reducing the processing frequency.

wingkwong:
> In the current scenario, with a table comprising 40 non-system fields and 1000 records, the serializeCellValue function will be invoked 40 * 1000 times. when utilizing the export/excel API and specifying the `fields` parameter, the dbRow object doesn't encompass all the fields. By utilizing `!dbRow.hasOwnProperty(column.title)`, unnecessary fields that don't need to be returned can be ignored, thereby reducing the processing frequency.

With the existing code, by specifying `fields[]=Title` (I got ~ 5 non-system fields), I can see only `Title` is  being serialised. Can you provide a video showing how to reproduce in your case?

```
[
  { Title: 'wqeqweqweqw' },
  { Title: 'qwqwe' },
  { Title: 'qweqweqwe qweqweqwe qweqweqweqweqweqwe' },
  { Title: 'qweqweqwe qweqweqwe qweqweqweqweqweqwe' },
  { Title: 'qweqwe' },
  { Title: '3' },
  { Title: null }
]
````

**Comment by wingkwong** [comment]
I tried on a table with 599 records & 10 non-system fields. I got the count of 5990 with AND without your change. Filtering some fields also returns the expected number. 

1. Can you show your project info? 
2. With & Without your change, are those exported data same?

**Comment by willnewii** [comment]
Sorry, I misunderstood it before, adjusted it. The strategy is to first use rows[0] to determine which fields need to be formatted. This eliminates the need to traverse all the fields. Of course, this only makes sense if a field or view is specified.

File: packages/nocodb/src/modules/datas/helpers.ts [hunk 1]
```diff
@@ -268,11 +268,15 @@ export async function getDbRows(param: {
       break;
     }
 
+    const needSerializeColumns = view.model.columns.filter((column) => {
```
wingkwong, line 271: @willnewii can you provide a scenario in detail so that I could test the difference? and why we need to filter `view.model.columns` based on the first row? Supposedly if a field is hidden, then it won't be in `view.model.columns`. In this case, after filtering, `needSerializeColumns` should be same as `view.model.columns`.

PR #6140, opened 2023-08-07
fix: qrcode text size limit

### LINKED ISSUE #6128: 🐛 Bug: QR-Code limited number of characters

### Steps to reproduce ?

create "singlelineText" and fill with URL with around 85 chars
create "QR code" column and choose newly created singleLineText column

with my iPhone I'm not able to read the QR code.

results in "undefined"

for a text with less chars its working fine

### Desired Behavior

display properly the URL when capturing the QR-Code

### Project Details

Node: **v16.20.0**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.109.6**

### Attachments

_No response_

sreehari2003:
## Change Summary

- closes #6128 

## Change type

Provide summary of changes.

reduced the text limit on qrcode

### DIFF
--- a/packages/nc-gui/components/virtual-cell/QrCode.vue
+++ b/packages/nc-gui/components/virtual-cell/QrCode.vue
@@ -3,7 +3,7 @@ import { useQRCode } from '@vueuse/integrations/useQRCode'
 import type QRCode from 'qrcode'
 import { RowHeightInj, computed, inject, ref } from '#imports'
 
-const maxNumberOfAllowedCharsForQrValue = 2000
+const maxNumberOfAllowedCharsForQrValue = 62
 
 const cellValue = inject(CellValueInj)
 
@@ -55,11 +55,13 @@ const { showEditNonEditableFieldWarning, showClearNonEditableFieldWarning } = us
     @ok="handleModalOkClick"
   >
     <template #footer>
-      <div class="mr-4" data-testid="nc-qr-code-large-value-label">{{ qrValue }}</div>
+      <div class="mr-4 overflow-scroll p-2" data-testid="nc-qr-code-large-value-label">
+        {{ qrValue }}
+      </div>
     </template>
     <img v-if="showQrCode" :src="qrCodeLarge" alt="QR Code" />
   </a-modal>
-  <div v-if="tooManyCharsForQrCode" class="text-left text-wrap mt-2 text-[#e65100] text-xs">
+  <div v-if="tooManyCharsForQrCode" class="text-left text-wrap mt-2 text-[#e65100] text-[10px]">
     {{ $t('labels.qrCodeValueTooLong') }}
   </div>
   <img


packages/nc-gui/components/virtual-cell/QrCode.vue
```diff
@@ -3,7 +3,7 @@ import { useQRCode } from '@vueuse/integrations/useQRCode'
 import type QRCode from 'qrcode'
 import { RowHeightInj, computed, inject, ref } from '#imports'
 
-const maxNumberOfAllowedCharsForQrValue = 2000
+const maxNumberOfAllowedCharsForQrValue = 62
```
wingkwong, line 6: how to get this number?

sreehari2003: i tried entering few characters , when the character count reach 63 qr code is not generating

wingkwong (commented):
I don't think this is the complete fix for the issue. This PR only prevents the cell to not show too many chars for QR code while they are possible to store in QR code standard. A QR code can store up to 7089 digits or 4296 characters and the current approach is limiting down to 62 characters. 

Currently we're using version 4  in `qrCodeOptions` which is pretty small. That's the reason why we cannot store too many chars. There're options we can configure the version (See [here](https://github.com/soldair/node-qrcode#qr-code-capacity) - we can bump the version to allow more characters).

Therefore, I think the better approach would be dynamically setting `version` in `qrCodeOptions` based on the length of the data in the cell, i.e. `qrValue?.value.length`. The more characters the data has, the higher version it requires. I'll leave you to do the math.

wingkwong:
@sreehari2003 Just skimmed through the docs a bit, `options.version` actually is not required for our case. If it's not specified, then the more suitable value will be used. They already do the math for you :D.  Ref: (https://github.com/soldair/node-qrcode#qr-code-capacity)

image

sreehari2003:
> @sreehari2003 Just skimmed through the docs a bit, `options.version` actually is not required for our case. If it's not specified, then the more suitable value will be used. They already do the math for you :D. Ref: (https://github.com/soldair/node-qrcode#qr-code-capacity)
> 
> image

is this qr is scannable ? i tried scanning some large text encoded qr and phone is not able to scan it but small qr is working , I have updated the pr

wingkwong:
> is this qr is scannable ? i tried scanning some large text encoded qr and phone is not able to scan it but small qr is working , I have updated the pr

Yes. It'll link to search engine with the given text. By the way, as you may have noticed the CI/CD failed, since the version is dynamic now, please update the `tests/playwright/tests/db/columns/columnQrCode.spec.ts` accordingly.

sreehari2003:
> > is this qr is scannable ? i tried scanning some large text encoded qr and phone is not able to scan it but small qr is working , I have updated the pr
> 
> Yes. It'll link to search engine with the given text. By the way, as you may have noticed the CI/CD failed, since the version is dynamic now, please update the `tests/playwright/tests/db/columns/columnQrCode.spec.ts` accordingly.

i checked the code but I don't see anywhere in the code where version is depended on , this is my first time working withe test cases

wingkwong:
@sreehari2003 Previously the version was hardcoded to `version: 4`. Since the version in this PR is dynamic, the library determines the version. Therefore, the value will be different. You just need to update the expected value.

PR #6153, opened 2023-08-09
Fixed #6151 : File name extends the area on CSV import

### LINKED ISSUE #6151: 🐛 Bug: File name extends the area on CSV import.

### Steps to reproduce ?

Steps to reproduce:
1. Import a CSV file with a large filename.
2. You will see that the filename extends the area.

### Desired Behavior

The filename should fit inside the box.

### Project Details

Version 0.109.4

### Attachments

Screenshot (268)

shashankvish0010:
## Change Summary

Added overflow property to a class template collapse. fixes #6151 

## Change type

## Test/ Verification

1. Import a CSV file with a large filename.
2. You could see that a scroll effect comes at the bottom to see the full file name, and the name resides in the box.

Screenshot (270)

### DIFF
--- a/packages/nc-gui/components/template/Editor.vue
+++ b/packages/nc-gui/components/template/Editor.vue
@@ -929,7 +929,7 @@ function handleCheckAllRecord(event: CheckboxChangeEvent, tableName: string) {
 
 <style scoped lang="scss">
 .template-collapse {
-  @apply bg-white;
+  @apply bg-white overflow-auto;
 }
 
 .template-form {
--- a/packages/nc-gui/components/template/Editor.vue
+++ b/packages/nc-gui/components/template/Editor.vue
@@ -749,7 +749,7 @@ function handleCheckAllRecord(event: CheckboxChangeEvent, tableName: string) {
               </a-form-item>
               <span v-else class="font-weight-bold text-lg flex items-center gap-2" @click="setEditableTn(tableIdx, true)">
                 <component :is="iconMap.table" class="text-primary" />
-                {{ table.table_name }}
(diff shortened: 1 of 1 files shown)


wingkwong (changes requested):
I think this change only looks good when you don't scroll. Ideally if it's too long, it should show ellipsis `...` instead without breaking the layout. Users should see the full text in edit mode.

image

shashankvish0010:
> I think this change only looks good when you don't scroll. Ideally if it's too long, it should show ellipsis `...` instead without breaking the layout. Users should see the full text in edit mode.
> 
> image

with recent changes the filename would look like this:

packages/nc-gui/components/template/Editor.vue
```diff
@@ -749,7 +749,7 @@ function handleCheckAllRecord(event: CheckboxChangeEvent, tableName: string) {
               </a-form-item>
               <span v-else class="font-weight-bold text-lg flex items-center gap-2" @click="setEditableTn(tableIdx, true)">
                 <component :is="iconMap.table" class="text-primary" />
-                {{ table.table_name }}
+                {{ table.table_name.length > 52 ? table.table_name.slice(0,52)+".." : table.table_name }}
```
wingkwong, line 752: It should be based on the width instead of a hardcoded value. Think of different device sizes. You can use `truncate` class from windicss which gives

```css
.truncate {
  overflow: hidden;
  -o-text-overflow: ellipsis;
  text-overflow: ellipsis;
  white-space: nowrap;
}
```

ref: https://windicss.org/utilities/general/typography.html

dstala:
Not observed in v0.202.5

PR #6182, opened 2023-08-15
fix: language name correction for ko

honnip:
## Change Summary

The word `한국인` means the Korean people, so in this case, '한국어' is correct, which means the language of Korea.

## Change type

## Test/ Verification

https://en.wikipedia.org/wiki/Korean_language


wingkwong (approved):
감사합니다

PR #6254, opened 2023-08-28
fix(nc-gui): search implemented for gallery view

reenphygeorge:
## Change Summary

Implemented search for gallery view

### DIFF
--- a/packages/nc-gui/components/smartsheet/Gallery.vue
+++ b/packages/nc-gui/components/smartsheet/Gallery.vue
@@ -39,6 +39,7 @@ const reloadViewDataHook = inject(ReloadViewDataHookInj)
 const openNewRecordFormHook = inject(OpenNewRecordFormHookInj, createEventHook())
 
 const { isViewDataLoading } = storeToRefs(useViewsStore())
+const { isSqlView, xWhere } = useSmartsheetStoreOrThrow()
 
 const expandedFormDlg = ref(false)
 const expandedFormRow = ref<RowType>()
@@ -54,7 +55,7 @@ const {
   addEmptyRow,
   deleteRow,
   navigateToSiblingRow,
-} = useViewData(meta, view)
+} = useViewData(meta, view, xWhere)
 
 provide(IsFormInj, ref(false))
 provide(IsGalleryInj, ref(true))
@@ -84,8 +85,6 @@ const isRowEmpty = (record: any, col: any) => {
 
   return Array.isArray(val) && val.length === 0
 }
-
-const { isSqlView } = useSmartsheetStoreOrThrow()
 
 const { isUIAllowed } = useUIPermission()
 const hasEditPermission = computed(() => isUIAllowed('xcDatatableEditable'))


packages/nc-gui/composables/useViewData.ts
```diff
@@ -160,14 +160,15 @@ export function useViewData(
   }
 
   async function loadData(params: Parameters<Api<any>['dbViewRow']['list']>[4] = {}) {
+    console.log(params)
```
wingkwong, line 163: please remove console log.

dstala (changes requested):
- [ ] didn't work for non primary value field
- [ ] didn't work when connected to external db (tried with sakila PG)

wingkwong (changes requested):
Your solution got a few problems. Suggest you read till the end first.

1. By hardcoding `(Title,like,%${search.value.query}%)`, you only assume the column is `Title`. Columns with other names would fail. Also, you cannot use `like` for bigint type. You may check out `xWhere` logic from either `useKanbanViewStore.ts` / `useSmartsheetStore.ts` for reference.

image 

2. After filtering, when you go to another views and switch back, the data should be filtered. In your case, filter text is here but the data is not filtered. This is because you only take care of `reloadViewDataHook` but not the view switch, i.e. the `loadData()` in `watch()`. 

Therefore, I think you may do the following

- undo everything you made
- retrieve `xWhere` from `useSmartsheetStoreOrThrow` in `Gallery.vue`, then pass `xWhere` to `useViewData` as the third parameter.

wingkwong:
Moreover, I think the search logic got another problem. @dstala please help confirm. probably we may create a new issue for it.

Assuming there are two columns, `Title1` and `Title2`

- setting search as `Title2=123` in View A, the data is shown as expected.
- switch to View B
- switch back to View A
- search become `Title1=123`, the data is shown unexpectedly.

The expected behaviour should be still showing data with `Title2=123`.

packages/nc-gui/composables/useViewData.ts
```diff
@@ -35,6 +35,7 @@ export function useViewData(
   meta: Ref<TableType | undefined> | ComputedRef<TableType | undefined>,
   viewMeta: Ref<ViewType | undefined> | ComputedRef<(ViewType & { id: string }) | undefined>,
   where?: ComputedRef<string | undefined>,
+  xWhere?:  ComputedRef<string | undefined>,
```
wingkwong, line 38: Here `xWhere` is useless. you don't need to modify this file at all. In fact, you are passing `xWhere` in `useViewData(meta, view, xWhere)` as the third parameter, which is `where` here.

wingkwong (changes requested):

dstala (approved):

wingkwong (approved):

dstala:
> Moreover, I think the search logic got another problem. @dstala please help confirm. probably we may create a new issue for it.
> 
> Assuming there are two columns, `Title1` and `Title2`
> 
> * setting search as `Title2=123` in View A, the data is shown as expected.
> * switch to View B
> * switch back to View A
> * search become `Title1=123`, the data is shown unexpectedly.
> 
> The expected behaviour should be still showing data with `Title2=123`.

@wingkwong created issue. Also asked Shreyas to redesign search box to include column name info display

dstala (approved):

Issue #5919, opened 2023-06-17
🐛 Bug: Deleting a table/project with attachment columns doesn't delete them on filesystem


GiantCrocodile:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Create a table with an attachments column. Then upload a file and delete the table or project. Check the folder /nc/uploads/noco/test/Sheet-1/[...]: the uploaded file is still there.

### Desired Behavior

Delete the files on filesystem as well when the table or project get's deleted

### Project Details

Node: **v18.12.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.109.1**

amisdun:
@GiantCrocodile How can I reproduce this, it's not clear to me.

GiantCrocodile:
@amisdun I just tried this again with latest nocodb version and it still happens.

I created a new table with name "Sheet-1". Then I added a new column of type "Attachment". Afterwards I uploaded a .txt file to the already existing row. I can see that the file is in the filesystem:

![grafik](https://github.com/nocodb/nocodb/assets/6271057/6c5645f5-d455-45b9-8473-98eb808053e6)

When I delete the attachment column afterwards, the file is still preserved on filesystem. Same when I delete the whole table in nocodb.

I'm running nocoDB via Cloudron (https://www.cloudron.io/). Not sure if this is related. I think if the filesystem would have wrong permissions (e. g. read-only), the upload would fail so if upload works, deleting should work as well.

wingkwong:
I can confirm that there is missing logic to handle the files in file system during record deletion.

GiantCrocodile:
Would be good if this gets fixed as it could be a GDPR violation/issue for the person who runs the instance in case they are not aware of this and someone uploads personal data according to GDPR.

Vyerni:
Appears this issue, or at least a variant of it is still an issue in the latest version. 

If I upload an attachment to an entry, I can see the attachment in the mapped folder to usr/app/data. However, when I then remove the attachment from the entry in NocoDB, the attachment no longer appears available, but the file still exists within the usr/app/data mapped folder.

dstala:
to be tracked here
https://github.com/nocodb/nocodb/issues/7895

ghost:
same problem here

PR #6271, opened 2023-08-30
Bug: Deleting a table/project with attachment columns doesn't delete them on filesystem

gitstart:
## Change Summary
#closes: https://github.com/nocodb/nocodb/issues/5919

Create a table with an attachments column. Then upload a file and delete the table or project. Check the folder /nc/uploads/noco/test/Sheet-1/[...]: the uploaded file is still there.

## Change type

- [x] Fix: Ensure Deletion of Attachments in Table/Project with Attachment Columns on Filesystem
## Test/ Verification

Provide summary of changes.
https://www.loom.com/share/6e6ed36c51c04fb683d6283293262029

---

### DIFF
--- a/packages/nocodb/src/controllers/old-datas/old-datas.service.ts
+++ b/packages/nocodb/src/controllers/old-datas/old-datas.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { nocoExecute } from 'nc-help';
+import { AttachmentsService } from 'src/services/attachments.service';
 import type { OldPathParams } from '~/modules/datas/helpers';
 import getAst from '~/helpers/getAst';
 import { NcError } from '~/helpers/catchError';
@@ -8,6 +9,8 @@ import NcConnectionMgrv2 from '~/utils/common/NcConnectionMgrv2';
 
 @Injectable()
 export class OldDatasService {
+  constructor(private readonly attachmentService: AttachmentsService) {}
+
   async dataList(param: OldPathParams & { query: any }) {
     const { model, view } = await this.getViewAndModelFromRequest(param);
     const source = await Source.get(model.source_id);
@@ -121,7 +124,15 @@ export class OldDatasService {
       dbDriver: await NcConnectionMgrv2.get(source),
     });
 
-    return await baseModel.delByPk(param.rowId, null, param.cookie);
+    const { data, deletedXcFiles } = await baseModel.delByPk(
+      param.rowId,
+      null,
+      param.cookie,
+    );
+
+    await this.attachmentService.xcFilesDelete(deletedXcFiles);
+
+    return data;
   }
 
   async getViewAndModelFromRequest(req) {
--- a/packages/nocodb/src/db/BaseModelSqlv2.ts
+++ b/packages/nocodb/src/db/BaseModelSqlv2.ts
@@ -19,6 +19,7 @@ import Validator from 'validator';
 import { customAlphabet } from 'nanoid';
 import DOMPurify from 'isomorphic-dompurify';
 import { v4 as uuidv4 } from 'uuid';
+import type { XcFile } from '~/interface/IStorageAdapter';
 import type LookupColumn from '~/models/LookupColumn';
 import type { Knex } from 'knex';
 import type { XKnex } from '~/db/CustomKnex';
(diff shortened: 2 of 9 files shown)


wingkwong (changes requested):
Ideally we should handle in `attachments.service.ts`. If I've set up a storage, says S3, then your solution won't work. You may check out  `fileDelete()` under `packages/nocodb/src/plugins` and implement them.

gitstart:
> Ideally we should handle in `attachments.service.ts`. If I've set up a storage, says S3, then your solution won't work. You may check out `fileDelete()` under `packages/nocodb/src/plugins` and implement them.

Absolutely, your input is greatly appreciated. To effectively test the deletion on S3 following the feature implementation, we kindly request the S3 credentials for access or any other way we can test it out. Your recommendation to explore the fileDelete() function within packages/nocodb/src/plugins will be instrumental in our testing process.

gitstart:
> > Ideally we should handle in `attachments.service.ts`. If I've set up a storage, says S3, then your solution won't work. You may check out `fileDelete()` under `packages/nocodb/src/plugins` and implement them.
> 
> Absolutely, your input is greatly appreciated. To effectively test the deletion on S3 following the feature implementation, we kindly request the S3 credentials for access or any other way we can test it out. Your recommendation to explore the fileDelete() function within packages/nocodb/src/plugins will be instrumental in our testing process.

The main problem was related to deleting files from the filesystem, which is why we didn't focus on S3 deletion.

wingkwong:
> The main problem was related to deleting files from the filesystem, which is why we didn't focus on S3 deletion.

Yes I understand. What you said is Local storage plugin (which is used by default if you haven't set up others). That's why I said you should move the deletion implementation to `plugins/storage/Local.ts`, so that we could delete the files by using something like `await storageAdapter.fileDelete(...)`. By doing so, we can separate the deletion logic for different storage plugins. You may take a look at `fileCreate` for reference.

For other storage types, just keep it as it is at this moment. 

Also for a complete solution, even thought not mentioned in the issue, I think it should also cater the cases like deleting row including attachments & deleting attachment columns. Please feel free to ping me at Discord for discussion.

mayomi1:
> > The main problem was related to deleting files from the filesystem, which is why we didn't focus on S3 deletion.
> 
> Yes I understand. What you said is Local storage plugin (which is used by default if you haven't set up others). That's why I said you should move the deletion implementation to `plugins/storage/Local.ts`, so that we could delete the files by using something like `await storageAdapter.fileDelete(...)`. By doing so, we can separate the deletion logic for different storage plugins. You may take a look at `fileCreate` for reference.
> 
> For other storage types, just keep it as it is at this moment.
> 
> Also for a complete solution, even thought not mentioned in the issue, I think it should also cater the cases like deleting row including attachments & deleting attachment columns. Please feel free to ping me at Discord for discussion.

Hi @wingkwong , 

We have pushed the update.

gitstart:
> > The main problem was related to deleting files from the filesystem, which is why we didn't focus on S3 deletion.
> 
> Yes I understand. What you said is Local storage plugin (which is used by default if you haven't set up others). That's why I said you should move the deletion implementation to `plugins/storage/Local.ts`, so that we could delete the files by using something like `await storageAdapter.fileDelete(...)`. By doing so, we can separate the deletion logic for different storage plugins. You may take a look at `fileCreate` for reference.
> 
> For other storage types, just keep it as it is at this moment.
> 
> Also for a complete solution, even thought not mentioned in the issue, I think it should also cater the cases like deleting row including attachments & deleting attachment columns. Please feel free to ping me at Discord for discussion.

Hi @wingkwong ,

We have pushed the update.

mertmit:
Hi I truly appreciate effort you've put into this fix, but there are other aspects which need to be figured out.
We allow copy/paste for cells and duplicating tables/bases which reuses same attachments. So it is not promised that a file stays within uploaded table only.

Recently I've added a feature to make image links temporarily available ([NC_SECURE_ATTACHMENTS](https://docs.nocodb.com/getting-started/self-hosted/environment-variables)). Moving forward we are planning to make this default behavior and in this new approach we moved away from using base/tableid in path bc of that reason.

So I agree there should be a delete mechanism for unused attachments but this will definitely require some extra steps and we need to discuss internally on the approach.

I am sincerely sorry for taking this on hold but I will keep you updated after discussing with the team.

Vyerni:
Is there at least any function or something that allows the app to scan for attachments in file storage, but are not actually present anywhere in the DB? Considering the storage path and name becomes obfuscated it's not something that can be manually done either

PR #6292, opened 2023-09-01
fix(nc-gui): fix search reset on view toggle

reenphygeorge:
## Change Summary

* Fixed search reset issue on toggling between views
* Added a feature to reload data trigger on field change

## Change type


packages/nc-gui/components/smartsheet/toolbar/SearchData.vue
```diff
@@ -90,7 +94,12 @@ watchDebounced(
       @click="isDropdownOpen = !isDropdownOpen"
     >
       <GeneralIcon icon="search" class="ml-1 h-3.5 w-3.5 text-gray-500 group-hover:text-black" />
-
+      <NcBadge color="purple">
+      <p class="pt-1 text-center text-xs font-regular text-gray-400"
```
wingkwong, line 98: 1. Use `truncate` which gives `overflow: hidden; text-overflow: ellipsis; white-space: nowrap;` to truncate overflowing text with an ellipsis (…) if needed.
2. I think `font-regular` makes no impact and is unnecessary here.

packages/nc-gui/components/smartsheet/toolbar/SearchData.vue
```diff
@@ -55,15 +55,19 @@ function onPressEnter() {
 }
 
 const displayColumnLabel = computed(() => {
-  return columns.value?.find((column) => column.primaryValue)?.label
+  const columnLabel = search.value.field.length === 0?columns.value?.find(column => column.primaryValue)?.label:
+  columns.value?.find(column => column.value === search.value.field)?.label
+  return (columnLabel && columnLabel.length > 10)
+    ? columnLabel.slice(0, 10) + '...'
+    : columnLabel;
```
wingkwong, line 62: 1. You can simply use `truncate` class instead of writing your own logic. See the comment at L98.
2. This looks long, packed and hard to read. You may simplify it like

```ts
const displayColumnLabel = computed(() => {
  if (search.value.field) {
    // use search field label if specified
    return columns.value?.find((column) => column.value === search.value.field)?.label
  }
  // use primary value label by default
  return columns.value?.find((column) => column.primaryValue)?.label
})
```

packages/nc-gui/components/smartsheet/toolbar/SearchData.vue
```diff
@@ -117,14 +126,13 @@ watchDebounced(
       :style="{
         width: '12rem',
       }"
-      :placeholder="`${$t('general.search')} in ${displayColumnLabel}`"
+      :placeholder="`${$t('general.search')}`"
```
wingkwong, line 129: can be simplified a bit

```ts
:placeholder="$t('general.search')"
```

wingkwong (commented):
Not sure if this is by design but it looks weird to me.

image

packages/nc-gui/components/smartsheet/toolbar/SearchData.vue
reenphygeorge: @wingkwong  I've used truncate but not working. Is it me doing anything wrong or any other issue ?

wingkwong: I've tested locally and it worked fine. `truncate` is added on L98, not in this code.

reenphygeorge: It's done

dstala (commented):
Screenshot 2023-09-03 at 10 25 12 AM

Blue badge highlights & takes users focus even when they are not using search bar. Let it be subtle.
- [x] Remove badge & blue background
- [x] Remove `in` and quotes around column title.
- [ ] Column title, let it be in grey 
- [ ] Dropdown icon, can appear on hover. Until hover, the space can be used to display column title. 
- [ ] Overall, space used to display column title can be reduced hence. We can display first 5-6 characters of column name & `...`. On hover. bring dropdown in place of `...`

wingkwong (changes requested):
You don't need `serachbarHover` at all. You can simply use `hover:xxxxx`. For example, if you want `w-16` on hover only, you can use `hover:w-16`. This would have higher precedence on hover. Most of the time you don't need to put `!` to override some styles.

dstala:
Looks better. Few minor changes
- [ ] On hover, the left part of search modal displaying title flickers & resizes. See if we can make it remain constant
- [ ] Overall size of the search bar can be reduced. The right part for search text input can be reduced.
- [ ] Reduce font size in field list drop down to 0.75rem as in column context menu & align icon size accordingly.

wingkwong (approved):

Issue #6239, opened 2023-08-26
🔦 Feature:  Ability to hide/unhide the table/project


Kartheravi:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

Some tables in the project are static and those tables are used very occasionally, in that case when the project of 30+ tables it is occupying lot of real estate in the left nav. of the page when all the projects are appearing in the left nav. 

P.S: In the older version, i push those tables to the bottom of the project so out of sight, out of mind. 

### Suggested Solution

(In 0.111) It would be great if the user can hide/unhide the table.

dstala:
having `starred` projects should address this & https://github.com/nocodb/nocodb/issues/6238

Kartheravi:
Hi @dstala, 
In the previous UI, i used the drag and drop feature extensively. the new UI was awesome and looks great. I faced these challenges when i was working with baserow in the past,

dstala:
cloud hosted solution now has concept of workspace. 
both self hosted & cloud hosted solutions support drag drop for tables.
no immediate plans in our roadmap to support hiding tables additionally

PR #6324, opened 2023-09-08
NCDBOSS-73 - 🔦 Feature: Ability to hide/unhide the table/project

gitstart:
## Change Summary
#closes:  https://github.com/nocodb/nocodb/issues/6239

### Desired Behavior
user can hide/unhide the table.

## Change type

- [x] feature:It would be great if the user can hide/unhide the table.
## Provide summary of changes.
https://www.loom.com/share/21b5517c72744da09e963bb504a8a045

---

### DIFF
--- a/.github/workflows/dispatch-oss.yml
+++ b/.github/workflows/dispatch-oss.yml
@@ -1,0 +1,20 @@
+name: "Dispatch OSS"
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ develop ]
+
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          const result = await github.rest.repos.createDispatchEvent({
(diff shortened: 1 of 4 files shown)


.github/workflows/dispatch-oss.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+    - uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          const result = await github.rest.repos.createDispatchEvent({
+            owner: 'nocodb',
+            repo: 'nocohub',
+            event_type: 'OSS'
+          })
```
dstala, line 20: revert changes in this file

dstala (changes requested):
Resolve merge conflicts & merge develop;

o1lab:
Unfortunately, hiding _just_ at the frontend will not do - as the idea here is to store the hidden status in backend as well and it is hidden for other roles. And there are other trickier questions like should it be hidden from the linked table. So this is a complex feature. However thanks for the effort @gitstart. Let our core team look into this.

Issue #6328, opened 2023-09-08
🐛 Bug: Unauthorized Access


wingkwong:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Some users reported Unauthorized Access after upgrading to latest version `0.111.2`. 

**from user sukhan**

> Something's broken. All applications that access the API give the error "Forbidden - perhaps check your credentials?". The key is correct.

![image](https://github.com/nocodb/nocodb/assets/35857179/0f1c2fa6-4770-427e-b6f8-569a8582b7f3)

**from user RAPTOR_6**

> I can't even log in anymore Error: Invalid credentials. Before, it gave me permission errors in GUI and I thought I would just relogin, which turn out to be a bad idea. The thing is that I had a backup from 2 hours before and the 0.111.0 and 0.111.1 versions are not working with database restored from backup, as I can't login using my admin account. I can't reset the password either, as the SMTP is broken for my provider from the start.

> I try to login using the GUI and it is not working, eg. Invalid credentials, this is the output:

```
09/08/2023 9:46:43 AM [Nest] 11  - 09/08/2023, 7:46:43 AM   ERROR [i] Invalid credentials
09/08/2023 9:46:43 AM Error: Invalid credentials
09/08/2023 9:46:43 AM     at Function.badRequest (/usr/src/app/docker/main.js:1:681451)
09/08/2023 9:46:43 AM     at f.validate (/usr/src/app/docker/main.js:1:1377403)
```

**from user Blusky**

> Using the API with an API token that previously worked. I also tried creating a new API Key (in case of wether or not it's a api key migration issue), same error. Also using n8n.

```
[Nest] 11  - 09/08/2023, 7:43:13 AM   ERROR [i] Unauthorized access
Error: Unauthorized access
    at Function.forbidden (/usr/src/app/docker/main.js:1:681524)
    at _.intercept (/usr/src/app/docker/main.js:1:803076)
    at /usr/src/app/node_modules/@nestjs/core/interceptors/interceptors-consumer.js:23:36
    at InterceptorsConsumer.intercept (/usr/src/app/node_modules/@nestjs/core/interceptors/interceptors-consumer.js:25:24)
    at /usr/src/app/node_modules/@nestjs/core/router/router-execution-context.js:46:60
    at async /usr/src/app/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

sasch901:
[...]

PR #6329, opened 2023-09-08
fix: assign project_roles for apitoken

mertmit:
Re #6328

## Change type

### DIFF
--- a/packages/nocodb/src/middlewares/extract-ids/extract-ids.middleware.ts
+++ b/packages/nocodb/src/middlewares/extract-ids/extract-ids.middleware.ts
@@ -180,8 +180,8 @@ export class ExtractIdsMiddleware implements NestMiddleware, CanActivate {
 }
 
 function getUserRoleForScope(user: any, scope: string) {
-  if (scope === 'project' || scope === 'workspace') {
-    return user?.project_roles || user?.workspace_roles;
+  if (scope === 'project') {
+    return user?.project_roles;
   } else if (scope === 'org') {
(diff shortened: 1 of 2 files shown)


pranavxc (approved):

packages/nocodb/src/middlewares/extract-ids/extract-ids.middleware.ts
```diff
@@ -180,8 +180,8 @@ export class ExtractIdsMiddleware implements NestMiddleware, CanActivate {
 }
 
 function getUserRoleForScope(user: any, scope: string) {
-  if (scope === 'project' || scope === 'workspace') {
-    return user?.project_roles || user?.workspace_roles;
+  if (scope === 'project') {
```
wingkwong, line 183: 1. Why we don't need to handle workspace? 
2. What if the scope is neither `project` nor `org`? What is the default return here?

pranavxc: 1. We don't have workspace concept here
2. There is only `org` and `project` scope at the moment

wingkwong (approved):

packages/nocodb/src/middlewares/extract-ids/extract-ids.middleware.ts
mertmit: - ws is not present for os
- scope is default to 'project', for others we specify via @Acl(perm, { scope: 'org' })

PR #6333, opened 2023-09-08
fix: super access

mertmit:
## Change Summary

- Fixes a bug with init admin from env (where changes are not applied but error is easy to miss)
- Navigate guest to root page from project
- Assign owner for super user for each project

## Change type


pranavxc (approved):

packages/nocodb/src/helpers/initAdminFromEnv.ts
```diff
@@ -227,6 +231,56 @@ export default async function initAdminFromEnv(_ncMeta = Noco.ncMeta) {
...
+              count: 1,
+            });
+
+            await User.insert(
+              {
+                email,
+                salt,
+                password,
+                email_verification_token,
+                roles,
```
pranavxc, line 278: let's set the token version as well, otherwise first login will fail

pranavxc (approved):

wingkwong (approved):

PR #6376, opened 2023-09-14
Profile page

mustafapc19:


packages/nc-gui/components/account/Profile.vue
```diff
@@ -0,0 +1,116 @@
...
+watch(
+  () => form.value.title,
+  async () => {
+    try {
+      isErrored.value = !(await formValidator.value.validate())
+    } catch (e: any) {
+      isErrored.value = true
+    }
+  },
+)
```
pranavxc, line 65: Suggestion: This watch can be replaced with `finishFailed` or `validate` event if possible.

https://antdv.com/components/form/#events

PR #6508, opened 2023-09-27
install pnpm only first time or on version mismatch

starbirdtech383:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/.github/workflows/playwright-test-workflow.yml
+++ b/.github/workflows/playwright-test-workflow.yml
@@ -17,32 +17,30 @@ jobs:
     timeout-minutes: 100
     steps:
       - name: Checkout
-        uses: actions/checkout@v3       
+        uses: actions/checkout@v3
+      - name: Check node,pnpm Installation and set Path
+        shell: bash
+        working-directory: scripts/self-hosted-gh-runner
+        timeout-minutes: 1
+        run: |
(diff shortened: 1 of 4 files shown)


starbirdtech383:
its noticed that pnpm bootstrap tries to download nodejs from https://nodejs.org/download/release/v18.14.0/node-v18.14.0-linux-x64.tar.gz
image

and it often times out. Its likely these things might get rate limited by these servers
ideally we should not get into this as node and pnpm are already installed
Few things i can think of
as part of bootstrap, can we skip downloading nodejs ? do we have control ?
if its needed, then why is it not using cache?
https://github.com/nocodb/nocohub/actions/runs/6322616490/job/17168567382

other related delays in pnpm get store path as well 
image

Conclusion : 
Looks like a bug in pnpm, started the discussion with pnpm team here https://github.com/orgs/pnpm/discussions/7140 tagged the code in question in this discussion
Found a workaround for now and have a pr open.
Basically pnpm setup step deletes the nodejs directory within pnpm folder and then any pnpm install command starts downloading nodejs again.
Tried standalone mode in setup pnpm but it didn’t help.

fix/ Workaround : is to not setup pnpm if required version is already available on machine

PR #6513, opened 2023-09-28
fix: various issues

mertmit:


wingkwong (commented):
I'm pretty sure `tests/playwright/tests/utils/general.ts` was the old code and it will override the latest code that merged 11 hours ago. Need to double check other files as well. Maybe sync the latest once to your branch first.

PR #6572, opened 2023-10-04
feat: signed attachments

mertmit:
## Change Summary

Adds signed/temporary attachment support for both local and S3.

So we will return signedPath/signedUrl with api response which will be available only set period of time (Defaults to 2 hours can be changed via NC_ATTACHMENT_EXPIRE_SECONDS).
(Note: expiration time is rounded to next xy:z0 so within period of 10 mins we return same link)

This change is backwards compatible (supports old links) unless user specifies NC_SECURE_ATTACHMENTS=true in his environment.

## Change type

### DIFF
--- a/packages/nocodb/src/cache/RedisCacheMgr.ts
+++ b/packages/nocodb/src/cache/RedisCacheMgr.ts
@@ -98,7 +98,7 @@ export default class RedisCacheMgr extends CacheMgr {
   async setExpiring(key: string, value: any, seconds: number): Promise<any> {
     if (typeof value !== 'undefined' && value) {
       log(
-        `RedisCacheMgr::set: setting key ${key} with value ${value} for ${seconds} seconds`,
+        `RedisCacheMgr::setExpiring: setting key ${key} with value ${value} for ${seconds} seconds`,
       );
       if (typeof value === 'object') {
         if (Array.isArray(value) && value.length) {
--- a/packages/nocodb/src/cache/RedisCacheMgr.ts
+++ b/packages/nocodb/src/cache/RedisCacheMgr.ts
@@ -95,6 +95,30 @@ export default class RedisCacheMgr extends CacheMgr {
   }
 
   // @ts-ignore
+  async setExpiring(key: string, value: any, seconds: number): Promise<any> {
+    if (typeof value !== 'undefined' && value) {
+      log(
+        `RedisCacheMgr::set: setting key ${key} with value ${value} for ${seconds} seconds`,
+      );
+      if (typeof value === 'object') {
+        if (Array.isArray(value) && value.length) {
+          return this.client.sadd(key, value);
+        }
+        return this.client.set(
+          key,
+          JSON.stringify(value, this.getCircularReplacer()),
+          'EX',
+          seconds,
+        );
+      }
+      return this.client.set(key, value, 'EX', seconds);
+    } else {
+      log(`RedisCacheMgr::set: value is empty for ${key}. Skipping ...`);
+      return Promise.resolve(true);
+    }
+  }
+
+  // @ts-ignore
   async getAll(pattern: string): Promise<any> {
     return this.client.hgetall(pattern);
   }
--- a/packages/nocodb/src/cache/RedisMockCacheMgr.ts
+++ b/packages/nocodb/src/cache/RedisMockCacheMgr.ts
@@ -93,7 +93,7 @@ export default class RedisMockCacheMgr extends CacheMgr {
   async setExpiring(key: string, value: any, seconds: number): Promise<any> {
     if (typeof value !== 'undefined' && value) {
       log(
-        `RedisMockCacheMgr::set: setting key ${key} with value ${value} for ${seconds} seconds`,
+        `RedisMockCacheMgr::setExpiring: setting key ${key} with value ${value} for ${seconds} seconds`,
       );
 
       // TODO: better way to handle expiration in mock redis
--- a/packages/nocodb/src/cache/RedisMockCacheMgr.ts
+++ b/packages/nocodb/src/cache/RedisMockCacheMgr.ts
@@ -90,6 +90,34 @@ export default class RedisMockCacheMgr extends CacheMgr {
   }
 
   // @ts-ignore
+  async setExpiring(key: string, value: any, seconds: number): Promise<any> {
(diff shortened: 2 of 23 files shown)


packages/nocodb/src/cache/RedisCacheMgr.ts
```diff
@@ -94,6 +94,30 @@ export default class RedisCacheMgr extends CacheMgr {
...
+        return this.client.set(
+          key,
+          JSON.stringify(value, this.getCircularReplacer()),
+          'EX',
+          seconds,
+        );
+      }
+      return this.client.set(key, value, 'EX', seconds);
+    } else {
+      log(`RedisCacheMgr::set: value is empty for ${key}. Skipping ...`);
```
wingkwong, line 116: `RedisCacheMgr::setExpiring`

packages/nocodb/src/cache/RedisMockCacheMgr.ts
```diff
@@ -89,6 +89,34 @@ export default class RedisMockCacheMgr extends CacheMgr {
     }
   }
 
+  // @ts-ignore
+  async setExpiring(key: string, value: any, seconds: number): Promise<any> {
+    if (typeof value !== 'undefined' && value) {
+      log(
+        `RedisMockCacheMgr::set: setting key ${key} with value ${value} for ${seconds} seconds`,
```
wingkwong, line 96: `RedisMockCacheMgr::setExpiring`

packages/nocodb/src/services/attachments.service.ts
```diff
@@ -34,24 +35,49 @@ export class AttachmentsService {
...
+          // url will be constructed in `useAttachmentCell`
+          attachment.path = `download/${filePath.join('/')}/${fileName}`;
+
+          promises.push(
+            TemporaryUrl.getTemporaryUrl({
+              path: attachment.path.replace(/^download\//, ''),
+            }).then((r) => (attachment.signedPath = r)),
+          );
+        } else {
+          if (attachment.url.includes('.amazonaws.com/')) {
```
wingkwong, line 69: what if i use minio? in this case, the url may not include `.amazonaws.com/`

packages/nocodb/src/controllers/attachments-secure.controller.ts
```diff
@@ -0,0 +1,71 @@
...
+@Controller()
+export class AttachmentsSecureController {
+  constructor(private readonly attachmentsService: AttachmentsService) {}
+
+  @UseGuards(GlobalGuard)
+  @Post(['/api/v1/db/storage/upload', '/api/v1/storage/upload'])
+  @HttpCode(200)
+  @UseInterceptors(UploadAllowedInterceptor, AnyFilesInterceptor())
+  async upload(@UploadedFiles() files: Array<any>, @Request() req) {
+    const path = `${moment().format('YYYY/MM/DD')}/${hash(req.user.id)}`;
```
wingkwong, line 32: what would happen if I upload the same file multiple times / multiple files with the same name on the same day?

packages/nocodb/src/plugins/s3/S3.ts
```diff
@@ -104,6 +106,14 @@ export default class S3 implements IStorageAdapterV2 {
     });
   }
 
+  public async getSignedUrl(key, expires = 7200) {
```
wingkwong, line 109: better call it `expiresInSeconds`

wingkwong (commented):
1. I'm still trying to understand the flow. can you briefly describe it step by step? from my understanding, it is like when we read a file (says local), it checks the attachment object and generate some temp urls (adding to cache), and then return a url like `https://xxxxxx/dltemp/xxx/xxxx`. When we browse that link, it fetches from cache and return the original url / official presigned one (?)
2. Does `tempUrl` mean presigned url? If so, I'd prefer the latter name as the name indicates it has time limit in general while `temp` doesn't.
3. Can you show one `tempUrl` as an example 
4. please put `NC_SECURE_ATTACHMENTS` and `NC_ATTACHMENT_EXPIRE_SECONDS` in noco-doc.

packages/nocodb/src/controllers/attachments-secure.controller.ts
mertmit: we don't keep name it is a nanoid, so they will go into same folder but shouldn't clash

packages/nocodb/src/plugins/s3/S3.ts
mertmit: 1. Upload part is not changed other than naming, we used to get path from query params now we decide instead.

I am assuming we are running instance with NC_SECURE_ATTACHMENTS set to true.
Read/List apis over attachment columns will return signedPath/signedUrl beside path/url which will be available for set period of time and url's will be not accessible anymore.

The steps you are explaining is pretty much correct.
- We bind path prefixed by set expire time to a signed link (temporary link) within cache
- When user tries to fetch that link we get path using that link.
As api calls trigger signing of the links only valid users can generate those links.

2. True, I will rename them accordingly
3. `http://localhost:8080/dltemp/1696436400000/2023/10/04/hAAj58GvTkWBGYJbpmzlcz0YfLHc43fE/lnSUOhOMuUO2OZuOri.jpg`
In order to avoid brute force attacks I will add a random token there as well which is missing currently
4. Sure, thank you for pointing out

Let me know if you still have any doubts/recommendations

mertmit:
1. Upload part is not changed other than naming, we used to get path from query params now we decide instead.

I am assuming we are running instance with NC_SECURE_ATTACHMENTS set to true.
Read/List apis over attachment columns will return signedPath/signedUrl beside path/url which will be available for set period of time and url's will be not accessible anymore.

The steps you are explaining is pretty much correct.
- We bind path prefixed by set expire time to a signed link (temporary link) within cache
- When user tries to fetch that link we get path using that link.
As api calls trigger signing of the links only valid users can generate those links.

2. True, I will rename them accordingly
3. `http://localhost:8080/dltemp/1696436400000/2023/10/04/hAAj58GvTkWBGYJbpmzlcz0YfLHc43fE/lnSUOhOMuUO2OZuOri.jpg`
In order to avoid brute force attacks I will add a random token there as well which is missing currently
4. Sure, thank you for pointing out

Let me know if you still have any doubts/recommendations

wingkwong (approved):
add those env vars to doc then all good.

PR #6586, opened 2023-10-05
fix: long text to single text paste

mertmit:
## Change Summary

Small fix until we find a better way to handle.

## Change type


dstala (commented):
Long text to long text : we still induce double quotes
Long text `abc "abc"` to single line text is becoming `abc ""abc""`

We can live with this as is for time being IMO. & can revisit later as requried.

dstala (approved):

PR #6624, opened 2023-10-09
Refactor: Rename row/column to record/field

mustafapc19:


wingkwong (commented):
please document this breaking change.

Issue #6465, opened 2023-09-24
🐛 Bug: Airtable import of base with column using a long name fails 


tjoneslo:
### Please confirm if bug report does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Create an AirTable base with a column name longer than 60 characters like:

Acknowledge you have read the terms and services agreement linked at the top of the form

Attempt to import this base into Nocodb using the directions given. You get the following error  

```
Importing Tables...
:: [1/3] Program Proposals
Column name By_checking_this_box_you_acknowledge_that_Readercon_s_progra exceeds 59 characters
```
I will note that the column name supplied is 60 characters long, which implies it is being truncated, but just one character too many. 

### Desired Behavior

Import the AirTable base with without error. 

You can rename the column in AirTable but that won't rename the column in the underlying database import, so changing the name won't work around the column too long issue. 

### Project Details

Node: **v16.20.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.111.4**

PR #6639, opened 2023-10-10
fix: import various

### LINKED ISSUE #6611: 🐛 Bug: Import CSV from Airtable export with Checkbox column fails

### Steps to reproduce ?

Using an existing Airtable table make sure there is a checkbox column in the table, and at least one row has value checked
export it to CSV file.

If you review the values in the CSV file you will see the checkbox column in the rows where the checkbox is selected have the data values of "checked"

Import the same file into NocoDB. The import fails with the error: 

```
invalid input syntax for type boolean: "checked"
```

### Desired Behavior

Checkbox columns import correctly. 

There is a workaround, where if you edit the CSV file before importing it, and change the checkbox column values of "checked" to "TRUE", the import works correctly. But you need to remember to update the CSV file before importing it. 

### Project Details

NocoDB version 0.111.4 via docker loading using a Postgres 12 database

### Attachments

_No response_

mertmit:
## Change Summary

Closes #6611 #6465

## Change type

### DIFF
--- a/packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
+++ b/packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
@@ -314,7 +314,7 @@ export class AtImportProcessor {
     const nc_getSanitizedColumnName = (table, name) => {
       let col_name = nc_sanitizeName(name);
 
-      // truncate to 60 chars if character if exceeds above 50
+      // truncate to 50 chars if character if exceeds above 50
       col_name = col_name?.slice(0, 50);
 
       // for knex, replace . with _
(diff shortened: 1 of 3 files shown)


packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
```diff
@@ -314,11 +314,11 @@ export class AtImportProcessor {
     const nc_getSanitizedColumnName = (table, name) => {
       let col_name = nc_sanitizeName(name);
 
-      // truncate to 60 chars if character if exceeds above 60
-      col_name = col_name?.slice(0, 60);
+      // truncate to 60 chars if character if exceeds above 50
```
wingkwong, line 317: `truncate to 50 chars if character if exceeds above 50`?

packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
```diff
@@ -328,13 +328,31 @@ export class AtImportProcessor {
...
+      }
+      if (duplicateColumn) {
+        let iterator = 1;
+        while (
+          table.columns.find(
+            (x) =>
+              x.column_name?.toLowerCase() ===
+              `${col_name}_${++iterator}`?.toLowerCase(),
+          )
+        ) {}
```
wingkwong, line 349: increasing iterator here would be more easy to understand.

mertmit: I've revised them for better readability

PR #6640, opened 2023-10-10
feat: limit column title to 255 for all db types

mertmit:
## Change Summary

- Allow column title up to 255 for all database types, and truncate column_name on backend if required

## Change type

### DIFF
--- a/packages/nc-gui/utils/validation.ts
+++ b/packages/nc-gui/utils/validation.ts
@@ -98,13 +98,15 @@ export const fieldRequiredValidator = () => {
   }
 }
 
-export const fieldLengthValidator = (sqlClientType: string) => {
+export const fieldLengthValidator = (_sqlClientType: string) => {
   return {
     validator: (rule: any, value: any) => {
       const { t } = getI18n().global
 
-      // no limit for sqlite but set as 255
-      let fieldLengthLimit = 255
+      const fieldLengthLimit = 255
+
+      /*
+      Now we allow 255 for all databases, truncate will be handled by backend for column_name
 
       if (sqlClientType === 'mysql2' || sqlClientType === 'mysql') {
         fieldLengthLimit = 64
@@ -113,6 +115,7 @@ export const fieldLengthValidator = (sqlClientType: string) => {
       } else if (sqlClientType === 'mssql') {
         fieldLengthLimit = 128
       }
+      */
 
       return new Promise((resolve, reject) => {
         if (value?.length > fieldLengthLimit) {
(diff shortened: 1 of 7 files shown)


packages/nc-gui/utils/validation.ts
```diff
@@ -98,13 +98,15 @@ export const fieldRequiredValidator = () => {
...
   return {
     validator: (rule: any, value: any) => {
       const { t } = getI18n().global
 
-      // no limit for sqlite but set as 255
-      let fieldLengthLimit = 255
+      const fieldLengthLimit = 255
+
+      /*
+      Now we allow 255 for all databases, truncate will be handled by backend for column_name
```
wingkwong, line 109: i would suggest to put this comment above L106. then delete the below block. or you may include those numbers in the comment for reference.

```
if (sqlClientType === 'mysql2' || sqlClientType === 'mysql') {
  fieldLengthLimit = 64
} else if (sqlClientType === 'pg') {
  fieldLengthLimit = 59
} else if (sqlClientType === 'mssql') {
  fieldLengthLimit = 128
}
```

packages/nc-gui/utils/validation.ts
```diff
@@ -98,13 +98,15 @@ export const fieldRequiredValidator = () => {
   }
 }
 
-export const fieldLengthValidator = (sqlClientType: string) => {
+export const fieldLengthValidator = (_sqlClientType: string) => {
```
wingkwong, line 101: `_sqlClientType: string` can be removed. we no longer need that. double check the caller.

packages/nocodb/src/services/columns.service.ts
```diff
@@ -1043,19 +1061,36 @@ export class ColumnsService {
...
 
-      if (
-        (param.column.title || param.column.column_name).length > mxColumnLength
-      ) {
+      if (!isVirtualCol(param.column)) {
+        param.column.column_name = sanitizeColumnName(param.column.column_name);
+      }
+
+      if (param.column.column_name.length > mxColumnLength) {
+        let colName = param.column.column_name.slice(0, mxColumnLength - 5);
```
wingkwong, line 1069: may add a comment to say `5` is a buffer for the suffix.

packages/nocodb/src/helpers/columnHelpers.ts
```diff
@@ -262,3 +262,7 @@ export async function populateRollupForLTAR({
   );
   await GridViewColumn.update(viewCol.id, { show: false });
 }
+
+export const sanitizeColumnName = (name: string) => {
```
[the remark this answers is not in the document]
mertmit: I will try that and remove that logic if I can in next iteration

PR #6672, opened 2023-10-11
chore: exclude pw package.json

mertmit:
## Change Summary

- Exclude pw from script as it should always target local sdk

## Change type


wingkwong (changes requested):
if you exclude it and run `pnpm i`, pnpm won't be able to find `nocodb-sdk` in pr / nightly build case since the name `nocodb-sdk/package.json` is changed to `nocodb-sdk-xxxxxx` unless you can make sure each install doesn't include pw.

mertmit:
Thank you for pointing out,
I've modified to just replace package name for pw. This way we don't have to replace it everytime.
Please let me know if this is looking good.

wingkwong (approved):

PR #6685, opened 2023-10-13
Mobile expanded form

dstala:
- Added responsive for mobile in expanded form hidden fields

### DIFF
--- a/packages/nc-gui/components/smartsheet/expanded-form/index.vue
+++ b/packages/nc-gui/components/smartsheet/expanded-form/index.vue
@@ -546,7 +546,7 @@ export default {
         <template v-else>
           <div class="flex flex-row w-full">
             <NcButton
-              v-if="props.showNextPrevIcons"
+              v-if="props.showNextPrevIcons && !isFirstRow"
               v-e="['c:row-expand:prev']"
               type="secondary"
               class="nc-prev-arrow !w-10"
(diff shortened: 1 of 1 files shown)


packages/nc-gui/components/smartsheet/expanded-form/index.vue
```diff
@@ -811,7 +818,7 @@ export default {
 }
 
 .nc-expanded-cell-header {
-  @apply w-full text-gray-700 xs:mb-2;
+  @apply w-full !sm:text-gray-500 !xs:text-gray-600 xs:mb-2;
```
wingkwong, line 821: `!` is actually not required as long as one didn't add `!text-gray-500` on `<LazySmartsheetHeaderCell>` / `<LazySmartsheetHeaderVirtualCell>`. Due to mobile first, `text-gray-500` is not even required on both tag since `sm:` one would cover it already.

Currently it is like 
- default: text-gray-500
- xs: text-gray-600 overriding the default one
- sm: text-gray-500 again overriding xs and default one (viewpoint take higher precedence)

image

wingkwong (commented):
1. better write something or show some screenshots. Not sure which part is meant to be fixed.

packages/nc-gui/components/smartsheet/expanded-form/index.vue
mustafapc19: @wingkwong Have resolved your suggestions

PR #6728, opened 2023-10-18
feat: groupby airtable import

dstala:
## Change Summary
Group by configuration import

## Change type

## Test/ Verification
Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
+++ b/packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
@@ -2162,7 +2162,8 @@ export class AtImportProcessor {
           datatype === UITypes.SingleLineText ||
           datatype === UITypes.Formula ||
           datatype === UITypes.Checkbox ||
-          datatype === UITypes.Collaborator
+          datatype === UITypes.Collaborator ||
+          datatype === UITypes.Number
         ) {
           ncGroup.push({
             group_column_id: columnId,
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
```diff
@@ -2120,6 +2128,84 @@ export class AtImportProcessor {
...
+        if (
+          datatype === UITypes.Date ||
+          datatype === UITypes.DateTime ||
+          datatype === UITypes.Links ||
+          datatype === UITypes.MultiSelect ||
+          datatype === UITypes.SingleSelect ||
+          datatype === UITypes.SingleLineText ||
+          datatype === UITypes.Formula ||
+          datatype === UITypes.Checkbox ||
+          datatype === UITypes.Collaborator
```
mertmit, line 2165: missing UITypes.Number which we support through UI

packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
```diff
@@ -2120,6 +2128,84 @@ export class AtImportProcessor {
...
+            direction: group.order,
+          });
+        } else {
+          // skip group by over other data types
+          updateMigrationSkipLog(
+            await sMap.getNcNameFromAtId(viewId),
+            colSchema.title,
+            colSchema.uidt,
+            `group config skipped; group over ${datatype}  not supported`,
+          );
```
mertmit, line 2178: this is not directly related to this PR I just wanted to point out we keep those logs in memory while operating but never make use of them.
maybe we can log them with stats or maybe include in T.event

mertmit (approved):

packages/nocodb/src/modules/jobs/jobs/at-import/at-import.processor.ts
dstala: will check in few instances & see how better we can log. will take it up in a separate PR

PR #6731, opened 2023-10-18
fix: Avoid getting unnecessary system column if getHiddenColumn is defined

pranavxc:
## Change Summary

- Avoid getting unnecessary system column(HM relation to junction table) if getHiddenColumn is defined

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/helpers/getAst.ts
+++ b/packages/nocodb/src/helpers/getAst.ts
@@ -130,23 +130,25 @@ const getAst = async ({
       ).ast;
     }
     let isRequested;
+
     if (getHiddenColumn) {
       isRequested =
         !isSystemColumn(col) ||
         col.column_name === 'created_at' ||
         col.column_name === 'updated_at' ||
         col.pk;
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/helpers/getAst.ts
```diff
@@ -18,7 +18,7 @@ const getAst = async ({
     nested: { ...(query?.nested || {}) },
     fieldsSet: new Set(),
   },
-  getHiddenColumn = false,
+  getHiddenColumn = query?.['getHiddenColumn'],
```
wingkwong, line 21: `query?.['getHiddenColumn'] || false` maybe?

wingkwong (approved):

packages/nocodb/src/helpers/getAst.ts
pranavxc: `undefined` or `null` will be falsey value so without `|| false` also fine

PR #6737, opened 2023-10-19
Darkened table scrollbar

mustafapc19:


dstala:
@mustafapc19 
- [ ] check firefox
- [ ] confirm width of all 3 scroll bars with @WarWolf47 (vertical in sidebar, both vertical and horizontal in grid)

PR #6759, opened 2023-10-20
chore: logic for replacing version

mertmit:
## Change Summary

- Depending on `workspace: ^` we need to manually push that change to master after release, we can instead use package name to replace the version
(see rn after the release master have versions instead of workspace: ^)
- Add --ignore-workspace for pnpm install of nocodb-sdk as we need to publish before bumping other packages

## Change type

### DIFF
--- a/scripts/upgradeNocodbSdk.js
+++ b/scripts/upgradeNocodbSdk.js
@@ -22,7 +22,8 @@ const replacePackageVersion = (filePath) => {
     return new Promise((resolve, reject) => {
         return fs.readFile(filePath, 'utf8', function (err, data) {
             if (err) return reject(err)
-            var result = data.replace(/workspace:\^/g, nocodbSdkPackage.version);
+            const rgx = new RegExp(`"${nocodbSdkPackage.name}": "(.*)"`, "g");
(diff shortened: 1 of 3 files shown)


wingkwong (commented):
If using a release version, then all sdk changes done in locally won't reflect in local dev, fe & be will take the version of the release one instead of local one.

mertmit:
Yes we should definitely target local after releases (even we can improve our release action to revert it back when finished) but imo it shouldn't break the release actions if something goes wrong with that.
So if this doesn't cause an extra issue might be good to have.

scripts/upgradeNocodbSdk.js
```diff
@@ -22,7 +22,8 @@ const replacePackageVersion = (filePath) => {
     return new Promise((resolve, reject) => {
         return fs.readFile(filePath, 'utf8', function (err, data) {
             if (err) return reject(err)
-            var result = data.replace(/workspace:\^/g, nocodbSdkPackage.version);
+            const rgx = new RegExp(`"${nocodbSdkPackage.name}": "(.*)"`, "g");
```
pranavxc, line 25: No need for the capturing group here since we are not using it.


Following is enough
```javascript
 const rgx = new RegExp(`"${nocodbSdkPackage.name}": ".*"`, "g");
```

pranavxc (approved):

PR #6831, opened 2023-10-30
Add support for `URLENCODE` function

mitya57:
## Change Summary

This adds `URLENCODE` that percent-encodes special characters, so that column value can be safely substituted into URLs.

## Change type

## Test/ Verification

Tested on PostgreSQL. But it should work fine on other backends, too.

### DIFF
--- a/packages/noco-docs/docs/070.fields/040.field-types/060.formula/030.string-functions.md
+++ b/packages/noco-docs/docs/070.fields/040.field-types/060.formula/030.string-functions.md
@@ -216,6 +216,27 @@ URL(text)
 URL('https://www.example.com') => a clickable link for https://www.example.com
 ```
 
+## URLENCODE
+The URLENCODE function percent-encodes special characters in a string so it can
+be substituted as a query parameter into a URL.
+
+It is similar to JavaScript `encodeURIComponent()` function, except it encodes
+only characters that have a special meaning according to RFC 3986 section 2.2
+and also percent signs and spaces; other characters such as letters from
+non-Latin alphabets will not be encoded. Like `encodeURIComponent()`, it should
+be used only for encoding URL components, not whole URLs.
+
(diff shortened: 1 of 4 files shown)


mitya57:
Tested on Postgres, and it works.
Screenshot

wingkwong (changes requested):
There's a few problems in this PR.

- You can't just implement one of the DB types. If this is inevitable, we should hide `URLENCODE` option for other DB types to avoid invalid formula error. However, ideally this shouldn't happen.
- `URL_ENCODE` requires a super user role to  run `CREATE EXTENSION` in advance before using it. In real life case, the DB user (connected in NocoDB) may not have such permission and this action shouldn't happen in NocoDB in my opinion. I think this approach is a no-go.
- You may probably just use `REPLACE` or `REGEX_REPLACE` in `knex.raw` to encode those special characters specified in RFC 3986.

mitya57:
> You may probably just use `REPLACE` or `REGEX_REPLACE` in `knex.raw` to encode those special characters specified in RFC 3986.

Okay, I implemented that approach. I used the list of characters in RFC 3986 section 2.2, but I removed characters that `encodeURIComponent()` does not encode, and added percent sign and space.

Please review again.

mitya57:
Here is how it looks now:
Screenshot

mitya57:
Rebased on top of the latest `develop` branch to fix the conflicts.
@wingkwong, @pranavxc Please review again.

mitya57:
Fixed a stupid copy-paste error in tests, now they should pass.

mitya57:
Improved the documentation.

mitya57:
Ping please? Just rebased again.

wingkwong:
Re-running the tests.

wingkwong (approved):

Issue #5988, opened 2023-06-29
🔦 Feature: Monday as first day of week in calendar popup


NaleVex:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

Currently popup calendar view only shows Sunday as first day of week which is not right for most European side of the world and use Monday as first day instead.
![f1e44c0015438911](https://github.com/nocodb/nocodb/assets/20393079/496edb1b-9036-4757-b941-527367105df2)

### Suggested Solution

Add a project setting to change first day to Monday when using calendar popup while changing value in date type column.

frankMagoba:
Hi @o1lab can I work on this?

dstala:
@frankMagoba let me know if you still wish to pick this up? this should be made configurable using an ENV variable

frankMagoba:
Yes @dstala I wish to pick it up. Should I make it configurable via env?

dstala:
No. Lets just make monday as first value always. So that weekend dates sit together in the end.

PR #6887, opened 2023-11-06
feat: Monday as first day of week in calendar popup

gitstart:
## Change Summary
closes: https://github.com/nocodb/nocodb/issues/5988

Monday as first day of week in calendar popup
## Change type

- [x] feat: Monday as first day of week in calendar popup

## Test/ Verification

Provide summary of changes.

Screenshot 2023-11-03 at 09 59 57

---

### DIFF
--- a/packages/nc-gui/components/cell/DateTimePicker.vue
+++ b/packages/nc-gui/components/cell/DateTimePicker.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import dayjs from 'dayjs'
 import { isSystemColumn } from 'nocodb-sdk'
+import updateLocale from 'dayjs/plugin/updateLocale'
 import {
   ActiveCellInj,
   CellClickHookInj,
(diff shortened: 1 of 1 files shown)


pranavxc (changes requested):
The behaviour should be same for DatetimePicker component as well.


Screenshot 2023-11-06 at 5 02 59 PM


It works fine if there is a DatePicker since the change done is globally applied.

Screenshot 2023-11-06 at 5 03 24 PM

So I would prefer to move the datejs change globally somewhere which runs independently globally.

```javascript
dayjs.extend(updateLocale)
dayjs.updateLocale('en', {
  weekStart: 1,
})
```

packages/nc-gui/components/cell/DateTimePicker.vue
```diff
@@ -25,8 +26,11 @@ interface Props {
 }
 
 const { modelValue, isPk, isUpdatedFromCopyNPaste } = defineProps<Props>()
-
 const emit = defineEmits(['update:modelValue'])
+dayjs.extend(updateLocale)
+dayjs.updateLocale('en', {
+  weekStart: 1,
+})
```
pranavxc, line 33: I would prefer to move it to `packages/nc-gui/plugins/a.dayjs.ts` rather than repeating.

PR #6894, opened 2023-11-06
refactor: Include request context in app-hooks

pranavxc:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/controllers/form-columns.controller.ts
+++ b/packages/nocodb/src/controllers/form-columns.controller.ts
@@ -1,4 +1,11 @@
-import { Body, Controller, Param, Patch, UseGuards , Request} from '@nestjs/common';
+import {
+  Body,
+  Controller,
+  Param,
+  Patch,
+  Request,
+  UseGuards,
+} from '@nestjs/common';
 import { GlobalGuard } from '~/guards/global/global.guard';
 import { FormColumnsService } from '~/services/form-columns.service';
 import { Acl } from '~/middlewares/extract-ids/extract-ids.middleware';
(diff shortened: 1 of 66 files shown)


packages/nocodb/src/controllers/forms.controller.ts
```diff
@@ -44,6 +45,7 @@ export class FormsController {
       body,
       tableId,
       user: req.user,
+      req,
```
wingkwong, line 48: since we include `req` now, we don't need to pass `req.user` again.

packages/nocodb/src/controllers/form-columns.controller.ts
```diff
@@ -19,10 +26,13 @@ export class FormColumnsController {
   async columnUpdate(
     @Param('formViewColumnId') formViewColumnId: string,
     @Body() formViewColumnbody: FormColumnUpdateReqType,
+
+    @Request() req: any,
```
wingkwong, line 30: can use `RequestType` instead of `any`? (apply to all)

wingkwong (commented):
1. please add the reason why we need to add request context in PR summary.

packages/nocodb/src/controllers/forms.controller.ts
pranavxc: Will do the cleanup part in a follow-up pr

packages/nocodb/src/controllers/form-columns.controller.ts
pranavxc: I will update this part as well in a followup pr... and inside hook I will use a custom type since in some case only certain fields we are passing ( user, headers, etc... )

pranavxc:
> 1. please add the reason why we need to add request context in PR summary.

At the moment we are not using it in OSS but later it will be helpful to extract data for Audit and listeners.

PR #6935, opened 2023-11-09
fix: Issues related to filter

pranavxc:
## Change Summary

- Support blank operation in filter with different format
- Negated filter operation missing null values when filtering using LTAR(Belongs To)

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/db/BaseModelSqlv2.ts
+++ b/packages/nocodb/src/db/BaseModelSqlv2.ts
@@ -5013,6 +5013,29 @@ export function extractCondition(
       [logicOp, alias, op, value] =
         str.match(/(?:~(and|or|not))?\((.*?),(\w+)\)/)?.slice(1) || [];
     }
+
+    // handle isblank and isnotblank filter format
+    switch (op) {
+      case 'is':
+        if (value === 'blank') {
+          op = 'blank';
+          value = undefined;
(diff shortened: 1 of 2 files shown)


packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -5013,6 +5013,29 @@ export function extractCondition(
       [logicOp, alias, op, value] =
         str.match(/(?:~(and|or|not))?\((.*?),(\w+)\)/)?.slice(1) || [];
     }
+
+    // handle isblank and isnotblank filter format
+    switch (op) {
+      case 'is':
+        if (value === 'blank') {
+          op = 'blank';
```
dstala, line 5021: `(field,is,blank)` : what will happen if field actually had value `blank` and we wanted to filter it? we should not support this case IMO. `field,isblank` `field,isnotblank` should suffice

pranavxc: Actually for that `eq` is what user should have used

dstala: Ah, my bad. Got it

dstala (approved):

PR #6965, opened 2023-11-13
fix: regex to get default value

### LINKED ISSUE #6964: 🐛 Bug: error with default value configuration in date field (PG)

### Steps to reproduce ?

works, but unable to trigger duplicate as the configuration cdf stored is invalid

duplicate date field function pg_catalog.timezone(text, character varying) does not exist

### Desired Behavior

x

### Project Details

0.202.5

### Attachments

_No response_

mertmit:
## Change Summary

Closes #6964

## Change type


dstala (approved):

dstala (changes requested):
- [ ] to discuss if upgrader is required; existing columns would continue to work with incorrect cdf configured but if one tries to duplicate, it leads to error

PR #7162, opened 2023-12-05
fix: filter mimetypes to preview in browser

mertmit:
## Change Summary

Only preview image, pdf and plain text attachments.

## Change type


pyozzi-toss:
I checked, but it seems that bypass is impossible.

The bypass point I checked is /api/v1/db/storage/upload-by-url:
```
{
       url: string;
       fileName: string;
       mimetype?: string;
       size?: string | number;
     }
```
Since I was receiving a mimetype as a parameter, I tried to bypass the verification logic by including ['image', 'pdf', 'text/plain'].

pranavxc (approved):

PR #7164, opened 2023-12-05
fix: leftover issue with alternative clients

mertmit:
## Change Summary

Add bindings for places missing it

## Change type

### DIFF
--- a/packages/nocodb/src/db/sql-client/lib/pg/YugabyteClient.ts
+++ b/packages/nocodb/src/db/sql-client/lib/pg/YugabyteClient.ts
@@ -119,7 +119,9 @@ class YBClient extends PGClient {
         // column['unique'] = response.rows[i]['cst'].indexOf('UNIQUE') === -1 ? false : true;
 
         column.cdf = response.rows[i].cdf
-          ? response.rows[i].cdf.replace(/::[\w ]+$/, '').replace(/^'|'$/g, '')
+          ? response.rows[i].cdf
+              .replace(/::[\w (),]+$/, '')
+              .replace(/^'|'$/g, '')
           : response.rows[i].cdf;
 
         // todo : need to find column comment
--- a/packages/nocodb/src/db/sql-client/lib/pg/PgClient.ts
+++ b/packages/nocodb/src/db/sql-client/lib/pg/PgClient.ts
@@ -891,7 +891,9 @@ class PGClient extends KnexClient {
         // column['unique'] = response.rows[i]['cst'].indexOf('UNIQUE') === -1 ? false : true;
 
         column.cdf = response.rows[i].cdf
-          ? response.rows[i].cdf.replace(/::[\w ]+$/, '').replace(/^'|'$/g, '')
+          ? response.rows[i].cdf
+              .replace(/::[\w (),]+$/, '')
+              .replace(/^'|'$/g, '')
           : response.rows[i].cdf;
 
         // todo : need to find column comment


packages/nocodb/src/db/sql-client/lib/mysql/VitessClient.ts
```diff
@@ -131,7 +131,8 @@ class Vitess extends MysqlClient {
       args.databaseName = this.connectionConfig.connection.database;
 
       const response = await this.sqlClient.raw(
-        `select *, table_name as tn from information_schema.columns where table_name = '${args.tn}' ORDER by ordinal_position`,
+        `select *, table_name as tn from information_schema.columns where table_name = ?? ORDER by ordinal_position`,
```
pranavxc, line 134: Should be `?`

packages/nocodb/src/db/sql-client/lib/pg/YugabyteClient.ts
```diff
@@ -77,8 +78,12 @@ class YBClient extends PGClient {
       INNER JOIN pg_catalog.pg_class c ON (a.attrelid=c.oid)
       LEFT OUTER JOIN pg_catalog.pg_attrdef ad ON (a.attrelid=ad.adrelid AND a.attnum = ad.adnum)
       LEFT OUTER JOIN pg_catalog.pg_description dsc ON (c.oid=dsc.objoid AND a.attnum = dsc.objsubid)
-      LEFT JOIN pg_index i ON (a.attnum = any(i.indkey) and a.attrelid = i.indrelid and i.indrelid = '${args.tn}'::regclass AND i.indisprimary)
-      WHERE NOT a.attisdropped AND c.relname = '${args.tn}' and a.attnum > 0 ORDER BY a.attnum`);
+      LEFT JOIN pg_index i ON (a.attnum = any(i.indkey) and a.attrelid = i.indrelid and i.indrelid = :table::regclass AND i.indisprimary)
```
pranavxc, line 81: Do we need to escape `:` here to avoid treating as identifier instead of string ? 

 `:table\\:\\:regclass`

packages/nocodb/src/db/sql-client/lib/pg/YugabyteClient.ts
```diff
@@ -114,7 +119,7 @@ class YBClient extends PGClient {
         // column['unique'] = response.rows[i]['cst'].indexOf('UNIQUE') === -1 ? false : true;
 
         column.cdf = response.rows[i].cdf
-          ? response.rows[i].cdf.split('::')[0].replace(/'/g, '')
+          ? response.rows[i].cdf.replace(/::[\w ]+$/, '').replace(/^'|'$/g, '')
```
pranavxc, line 122: type can contain scale, precision etc. May be `.replace(/::[\w ]+$/, '')` won't be enough.

for example:

```
value::NUMERIC(1,2)
```

pranavxc (changes requested):

packages/nocodb/src/db/sql-client/lib/pg/YugabyteClient.ts
mertmit: We don't escape on packages/nocodb/src/db/sql-client/lib/pg/PgClient.ts - columnList and seem to be working

packages/nocodb/src/db/sql-client/lib/pg/YugabyteClient.ts
mertmit: I replaced regex with /::[\w (),]+$/ to support scale/precision as well (both pg and yugabyte)
Let me know if this is enough (I am trying to avoid split bc that way we need to handle :: in value as well, with regex we are making sure it is on ending)

pranavxc: Looks fine now

pranavxc (approved):

PR #7342, opened 2024-01-02
User Management sort bug fixes and user field UI/UX

rameshmane7218:
## Change Summary

- Fixed `User Management` role update issue on project level and sort role according to `Role Hierarchy`
- Added user initials as a profile pic in `User Field` dropdown options

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/components/project/AccessSettings.vue
+++ b/packages/nc-gui/components/project/AccessSettings.vue
@@ -10,6 +10,7 @@ import {
 } from 'nocodb-sdk'
 import type { Roles, WorkspaceUserRoles } from 'nocodb-sdk'
 import { isEeUI, storeToRefs, useUserSorts } from '#imports'
+import type { User } from '#imports'
 
 const basesStore = useBases()
 const { getBaseUsers, createProjectUser, updateProjectUser, removeProjectUser } = basesStore
@@ -85,7 +86,7 @@ const updateCollaborator = async (collab: any, roles: ProjectRoles) => {
         WorkspaceRolesToProjectRoles[currentCollaborator.workspace_roles as WorkspaceUserRoles] === roles &&
         isEeUI)
     ) {
-      await removeProjectUser(activeProjectId.value!, currentCollaborator as any)
+      await removeProjectUser(activeProjectId.value!, currentCollaborator as unknown as User)
       if (
         currentCollaborator.workspace_roles &&
         WorkspaceRolesToProjectRoles[currentCollaborator.workspace_roles as WorkspaceUserRoles] === roles &&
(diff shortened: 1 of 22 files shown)


packages/nc-gui/components/project/AccessSettings.vue
```diff
@@ -85,7 +85,7 @@ const updateCollaborator = async (collab: any, roles: ProjectRoles) => {
         WorkspaceRolesToProjectRoles[currentCollaborator.workspace_roles as WorkspaceUserRoles] === roles &&
         isEeUI)
     ) {
-      await removeProjectUser(activeProjectId.value!, collab)
+      await removeProjectUser(activeProjectId.value!, currentCollaborator as any)
```
wingkwong, line 88: fix the type instead of using `any`. (apply to all)

packages/nocodb-sdk/src/lib/enums.ts
```diff
@@ -247,6 +247,14 @@ export const OrderedWorkspaceRoles = [
   null,
 ];
 
+export const OrderedOrgRoles = [
+  OrgUserRoles.SUPER_ADMIN,
+  OrgUserRoles.CREATOR,
+  OrgUserRoles.VIEWER,
+  // placeholder for no access
+  null,
```
wingkwong, line 255: what about `OrgUserRoles.NO_ACCESS`?

wingkwong (changes requested):

packages/nocodb-sdk/src/lib/enums.ts
rameshmane7218: We only have 3 roles at Org level `Super`, `Creator` & `Viewer`

packages/nc-gui/components/project/AccessSettings.vue
rameshmane7218: Can I use `as unknown as User` to remove ts error because `removeProjectUser/updateProjectUser/createProjectUser` accept type `User` and current type is `Collaborators`

rameshmane7218: For now I added `as unknown as User` let me know if this is not a correct way

wingkwong: I'm fine with it

packages/nocodb-sdk/src/lib/enums.ts
wingkwong: I'm wondering why we need the null value here.

rameshmane7218: I followed the same approch that already used for `OrderedWorkspaceRoles`

```ts
export const OrderedWorkspaceRoles = [
  WorkspaceUserRoles.OWNER,
  WorkspaceUserRoles.CREATOR,
  WorkspaceUserRoles.EDITOR,
  WorkspaceUserRoles.COMMENTER,
  WorkspaceUserRoles.VIEWER,
  // placeholder for no access
  null,
];
```

rameshmane7218: @mertmit any idea why we used null as a placeholder for no access in `OrderedWorkspaceRoles `

mertmit: Initially we were depending OrderedWorkspaceRoles & OrderedProjectRoles to map roles between ws - project. Later as this approach was error prone I've started using WorkspaceRolesToProjectRoles instead, so it should be safe to get rid of that null in OrderedWorkspaceRoles, and it is not required at all for org level.

rameshmane7218: Okay thanks, I will remove null form both places

PR #7343, opened 2024-01-02
fix: Reload data after bulk update

### LINKED ISSUE #7296: 🐛 Bug: select multiple cells, clear - related formula column's data doesn't reload automatically

### Steps to reproduce ?

- CONCAT({Title}, " something")
- Select multiple cells in {Title} field, clear
- Formula display field doesn't get updated immediately. Needs a page reload

### Desired Behavior

Formula fields should immediately reflect new change

### Project Details

202.10

### Attachments

_No response_

pranavxc:
## Change Summary

- Closes #7296 

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/composables/useData.ts
+++ b/packages/nc-gui/composables/useData.ts
@@ -1,5 +1,5 @@
-import { UITypes } from 'nocodb-sdk'
 import type { ColumnType, LinkToAnotherRecordType, PaginatedType, RelationTypes, TableType, ViewType } from 'nocodb-sdk'
+import { UITypes } from 'nocodb-sdk'
 import type { ComputedRef, Ref } from 'vue'
 import {
   NOCO,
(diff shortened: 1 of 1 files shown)


packages/nc-gui/composables/useData.ts
```diff
@@ -403,6 +403,8 @@ export function useData(args: {
       if (row.rowMeta) row.rowMeta.saving = false
     }
 
+    // reload data since row update may change the other columns data( formula, QrCode, Barcode, Rollup, Checkbox, User, Auto Updated Datetime, etc...)
+    await callbacks?.loadData?.()
```
mertmit, line 407: In the block above we try to update those to avoid refreshing all page with every bulk update, isn't it possible to avoid calling this

dstala (approved):

packages/nc-gui/composables/useData.ts
pranavxc: If there is column which refers to the updated column(eg: Formula, QrCode, Barcode, etc...) it won't reload on clearing the cell. If we are clearing single cell it's updating row using the response data but for bulk update we don't have those information.

pranavxc: We can skip it when unnecessary by checking presence of any such column

mertmit (approved):
This is good for now, fixes the issue mentioned.

For future reference:
- We can later switch to silent refresh (page loader comes up every time)
- Also if there is a sort applied this can cause modified columns to reorder to other pages, we need to think about UX carefully on those (one alternative is we can refresh selected rows only with a list call with filtering pk's in array providing our pks)

dstala (approved):

PR #7464, opened 2024-01-19
feat: revise cache

mertmit:
## Change Summary

- deprecate getAll and delAll (they were using `KEYS pattern` which we would like to avoid)
- add bulk del support (less number of calls to Redis)
- set before all appendToList calls (we make use of existing child key so this is a must now)
- keep list of parents on child object
  - extra operations for functions
    - set: get existing __nc_list__ if any and keep it for new value 
    - appendToList: get existing __nc_list__ if any & append list being appended and set child
    - setList: get existing __nc_list__ if any & append list being set and set child
- use list of parents to delete CHILD_TO_PARENT
- use mget for getList (less number of calls to Redis)

## Change type


wingkwong (approved):
LGTM

PR #8228, opened 2024-04-09
Update copyright to 2024

oskarrough:
This is a tiny PR. In the sidebar bottom left it still said copyright 2023. This updates it to 2024.

screenshot 2024-04-09 at 08 59 19

### DIFF
--- a/packages/nc-gui/components/dashboard/Sidebar/UserInfo.vue
+++ b/packages/nc-gui/components/dashboard/Sidebar/UserInfo.vue
@@ -176,7 +176,7 @@ onMounted(() => {
     </NcDropdown>
 
     <template v-if="isMobileMode"></template>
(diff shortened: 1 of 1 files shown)


oskarrough:
> @oskarrough Do we need to change it?

Hah, that's the right question. No, maybe not.

That site also states:

> [...] AND WEBSITES? New versions or editions of works should contain the publication date of the new version or edition. For constantly evolving websites and blogs that contain works published over several years, the notice may include a range of years (e.g., 2011–2023), starting from the date of the oldest published elements and ending with the date of the newest published elements.

So, maybe the copyright notice could just be removed completely in the sidebar. Why would it need to be visible there at all times? It feels that the license is clearly stated in other places.

Personally when I see copyright years that are off by one or two, it always felt like "oh this software wasn't updated", which led me to create this PR. In any case, feel free to close!

PR #8291, opened 2024-04-17
Auto Upstall Enhancements: Added help script generation

rohittp0:
## Change Summary

- Created an advanced options menu where the users can set the number of `nocodb` container instances to run
- `noco.sh` now generates a script `help.sh` that helps the user in managing their NocoDB. It provides functionality to
   - Start / Stop / Restart all containers
   - Upgrade NocoaDB to the latest version
   - See logs for each container ( `nocodb`,  `postgres`,  `nginx`, `redis`, `watchtower` )
   - Scale the number of `nocodb` containers
   - Monitor the resource usage

## Change type

## Test/ Verification

The script has been tested in the provided VM and is verified to be working. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced the setup script to support running multiple instances of NocoDB.
	- Introduced a service management menu for better user interaction.
	- Improved SSL certificate handling for increased security.
- **Refactor**
	- Updated service configurations for optimized performance.
	- Switched to using Docker commands directly for specific operations, enhancing script efficiency.

### DIFF
--- a/docker-compose/setup-script/noco.sh
+++ b/docker-compose/setup-script/noco.sh
@@ -62,6 +62,42 @@ check_for_docker_compose_sudo() {
     fi
 }
 
+# Function to read a number from the user
+read_number() {
+    local number
+    read -p "$1" number
+
+    # Ensure the input is a number or empty
+    while ! [[ $number =~ ^[0-9]+$ ]] && [ -n "$number" ] ; do
+        read -p "Please enter a valid number: " number
+    done
+
+    echo $number
+}
+
+# Function to read a number within a range from the user
+read_number_range() {
+    local number, min, max
+
+    # Check if there are 3 arguments
+    if [ "$#" -ne 3 ]; then
+        number=$(read_number)
+        min=$1
+        max=$2
+    else
+        number=$(read_number "$1")
+        min=$2
+        max=$3
+    fi
+
+    # Ensure the input is in the specified range
+    while [[ -n "$number" && ($number -lt $min || $number -gt $max) ]]; do
+        number=$(read_number "Please enter a number between $min and $max: ")
+    done
+
+    echo $number
+}
+
 # *****************    HELPER FUNCTIONS END  ***********************************
 # ******************************************************************************
 
@@ -213,7 +249,18 @@ else
     message_arr+=("Watchtower: Disabled")
 fi
 
+echo "Show Advanced Options [Y/N] (default: N): "
+read ADVANCED_OPTIONS
 
+if [ -n "$ADVANCED_OPTIONS" ] && { [ "$ADVANCED_OPTIONS" = "Y" ] || [ "$ADVANCED_OPTIONS" = "y" ]; }; then
+    NUM_CORES=$(nproc)
+    echo  "How many instances of NocoDB do you want to run? (Maximum: ${NUM_CORES} (default: 1): "
+    NUM_INSTANCES=$(read_number_range 1 $NUM_CORE
+fi
+
+if [ -z "$NUM_INSTANCES" ]; then
+    NUM_INSTANCES=1
+fi
 
 # ******************************************************************************
 # *********************** INPUTS FROM USER END  ********************************
--- a/docker-compose/setup-script/noco.sh
+++ b/docker-compose/setup-script/noco.sh
@@ -686,6 +686,7 @@ monitoring_service() {
 
 # Main program loop
 while true; do
+    trap - INT
     show_menu
     echo "Enter your choice: "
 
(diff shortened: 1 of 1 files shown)


docker-compose/setup-script/noco.sh
```diff
@@ -213,7 +248,20 @@ else
     message_arr+=("Watchtower: Disabled")
 fi
 
+echo "Show Advanced Options [Y/N] (default: N): "
+read -r ADVANCED_OPTIONS
+
+if [ -n "$ADVANCED_OPTIONS" ] && { [ "$ADVANCED_OPTIONS" = "Y" ] || [ "$ADVANCED_OPTIONS" = "y" ]; }; then
+    NUM_CORES=$(nproc)
```
pranavxc, line 255: Does `nproc` command available in macos ?

image

If we are not planning to support Macos then it's fine (cc @o1lab).

o1lab: ya we wont need it on local machines

docker-compose/setup-script/noco.sh
```diff
@@ -477,37 +532,143 @@ server {
...
+    case \$log_choise in
+        1) $DOCKER_COMMAND compose logs -f nocodb ;;
+        2) $DOCKER_COMMAND compose logs -f db ;;
+        3) $DOCKER_COMMAND compose logs -f nginx ;;
+        4) $DOCKER_COMMAND compose logs -f redis ;;
+        5) $DOCKER_COMMAND compose logs -f watchtower ;;
+        6) $DOCKER_COMMAND compose logs -f ;;
+        0) return ;;
+        *) echo "Invalid choice. Returning to main menu." ;;
+    esac
```
pranavxc, line 601: If possible we can make these part dynamic based on users choice selection since few services are optional.

o1lab: I mentioned to @rohittp0 that if we can somehow show docker stats containers with their name and user can select that to show the logs.

docker-compose/setup-script/noco.sh
```diff
@@ -477,37 +532,143 @@ server {
...
+done
 EOF
-fi
 
+message_arr+=("Help script: help.sh")
 
-message_arr+=("Update script: update.sh")
+cat > ./update.sh <<EOF
+$DOCKER_COMMAND compose pull
+$DOCKER_COMMAND up -d --force-recreate
```
pranavxc, line 664: `$DOCKER_COMMAND up` or `$DOCKER_COMMAND compose up` ?

docker-compose/setup-script/noco.sh
pranavxc: May be we can use `docker-compose config` - https://github.com/docker/compose/issues/2683

docker-compose/setup-script/noco.sh
rohittp0: Typo, fixed it :)

docker-compose/setup-script/noco.sh
rohittp0: Now it reads the container info and also accounts for replicas,

```markdown
Select a container for logs:
1. db (1 replicas)
2. nginx (1 replicas)
3. nocodb (3 replicas)
4. redis (1 replicas)
5. watchtower (1 replicas)
A. All
0. Back to main menu
Enter your choice: 
```

```markdown
Select a replica for nocodb:
1. nocodb replica 1
2. nocodb replica 2
3. nocodb replica 3
A. All
Enter replica number: 
```

rohittp0:
I tried playing with the UI, this is what I came up with
image

The trade-off is that it comes with an additional dependency `dialog`. Should I add this ?

rohittp0:
From what I have tested this is working fine with both Windows terminal and the IDE terminal, but not sure if it will break on any other terminal ( if it doesn't have color support or something ).

pranavxc:
After local testing, I have several observations:

1.  It seems that the `nproc` command is not available by default on macOS. It may need to be installed if necessary to ensure compatibility.
2. During the execution of `help.sh`, it appears that pressing Ctrl + C does not result in the expected exit behaviour.
3. Continue execution only after pressing enter key while running `help.sh`
4. Regarding the proposed dialog UI changes (https://github.com/nocodb/nocodb/pull/8291#issuecomment-2065213251), Let's stick with current implementation since dialog looks slightly odd.
5. When choosing `logs` option some errors are there in MacOS terminal.
   Screenshot 2024-04-20 at 3 46 13 PM
6. Additionally if possible check `docker` service is in running or not and show a warning message.

rohittp0:
image

The warning will show up like this if docker is not running

o1lab (approved):

Issue #2759, opened 2022-07-20
NC_ADMIN_EMAIL & NC_ADMIN_PASSWORD not working if super user missing


mertmit:
**Steps To Reproduce**

1. Create an user (there should be at least one existing user other than super to reproduce)
2. Remove super user from database
3. Try to start instance using env variables
4. Error raised
```
Error occurred while updating/creating admin user
TypeError: Cannot read properties of undefined (reading 'email')
    at initAdminFromEnv (/home/srom/repos/nocodb/packages/nocodb/src/lib/meta/api/userApi/initAdminFromEnv.ts:100:33)
    at Noco.init (/home/srom/repos/nocodb/packages/nocodb/src/lib/Noco.ts:191:5)
    at /home/srom/repos/nocodb/packages/nocodb/src/run/docker.ts:30:14
```

**Expected Behavior**

If no super user present, one should be created using the env variables.

Issue #8230, opened 2024-04-09
🐛 Bug: Copying and pasting from Excel overwrites one additional row cell


halphalp:
### Please confirm if bug report does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. Copy values from Excel
2. Select a location in a NocoDB table to paste
3. Paste the values and note that the row after the last pasted row has its value erased

Essentially, if I copy 5 cells from Excel and paste them into rows 1-5 in a NocoDB table, the value in the column I'm pasting for row 6 will be erased if it exists.

### Desired Behavior

Only update the rows being pasted - not beyond.

### Project Details

Node: **v20.11.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.205.1**

dstala:
tried copying from a Google spreadsheet, unable to reproduce. Can you confirm if you by mistake didn't copy one extra empty cell? you can paste us a video of this issue. would help us.

Haven't tried using MS Excel though

o1lab:
Could you please try it in chrome or others once.

xtyuns:
the same question，When the content copied from excel exceeds 1 column, you will also encounter an error message from the front end.

step1: create a table
![r1](https://github.com/nocodb/nocodb/assets/41326335/0c0a2b91-a6d4-4f80-b688-9379a6df2854)

step2: copy from excel
![r2](https://github.com/nocodb/nocodb/assets/41326335/1d702191-e310-44dc-ae83-6b452c241e1c)

step3: paste in table
![r3](https://github.com/nocodb/nocodb/assets/41326335/4e05a5f5-adb6-478d-8543-fa4998a0be6f)

xtyuns:
The reason for this problem is that clipboardData always ends with `\n`.
![image](https://github.com/nocodb/nocodb/assets/41326335/cb4c9187-675d-4024-8bef-9061f4d9c797)

So when you select 2 columns from Excel copy and paste them into nocodb, it will produce an extra row with only one column
![image](https://github.com/nocodb/nocodb/assets/41326335/24ad7f07-ad0b-4319-8787-87f8f7a083c8)

PR #8317, opened 2024-04-22
Nc fix(nc-gui): Copying and pasting from Excel overwrites one additional row cell

rameshmane7218:
## Change Summary

- ref: #2759

## Change type

## Test/ Verification

Provide summary of changes.

## Summary by CodeRabbit

- **Bug Fixes**
    - Improved clipboard data handling in multi-select functionality by removing trailing newline characters.

### DIFF
--- a/packages/nc-gui/composables/useMultiSelect/index.ts
+++ b/packages/nc-gui/composables/useMultiSelect/index.ts
@@ -816,7 +816,13 @@ export function useMultiSelect(
     e.preventDefault()
 
     // Replace \" with " in clipboard data
-    const clipboardData = e.clipboardData?.getData('text/plain') || ''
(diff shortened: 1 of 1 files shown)


dstala (approved):

xtyuns:
ref: #8230 
At the same time, I want to prompt something: when copying content from an online document(like: google sheets), it does not have the behavior of appending `\n` at the end, but if the user happens to select some cells with empty content at this time, the changes in this PR will cause a cell to be lost.

Should it be considered: when the content of clipboardData has only one column and the end is a `\n`, a prompt box pops up to ask the user if the user needs to remove the end of one `\n` (or: whether to insert a new row?), I think this kind of behavior that shows asking users may be better.

PR #8385, opened 2024-05-03
fix: filter datetime (sqlite / mysql)

cangku:
## Change Summary

- fix datatime can't filter data, where is today, datetime is now, use date without timezone no error, has timezone will can't filter. under other conditions, it's necessary to consider whether the time point is the beginning or the end of a day.
﻿
- fix mysql can't use datetime filter.

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/db/conditionV2.ts
+++ b/packages/nocodb/src/db/conditionV2.ts
@@ -597,12 +597,12 @@ const parseConditionV2 = async (
             UITypes.LastModifiedTime,
           ].includes(column.uidt)
         ) {
-          let now = dayjs(new Date());
+          let now = dayjs(new Date()).utc();
           const dateFormatFromMeta = column?.meta?.date_format;
           if (dateFormatFromMeta && isDateMonthFormat(dateFormatFromMeta)) {
             // reset to 1st
-            now = dayjs(now).date(1);
-            if (val) genVal = dayjs(val).date(1);
+            now = dayjs(now).utc().date(1);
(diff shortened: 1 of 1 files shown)


cangku:
Is no one handling this fix? @dstala @pranavxc

pranavxc:
> Is no one handling this fix? @dstala @pranavxc

Sorry for the delay, we will review it tomorrow.

pranavxc (commented):
All looks good and I have few doubts, will test locally couple of scenarios and then I will proceed with merge.

packages/nocodb/src/db/conditionV2.ts
```diff
@@ -740,7 +759,9 @@ const parseConditionV2 = async (
                   // else
                   qb = qb.where(knex.raw('??::date = ?', [field, val]));
                 } else {
-                  qb = qb.where(knex.raw('DATE(??) = DATE(?)', [field, val]));
+                  qb = qb.where(
+                    knex.raw("DATE(??, 'localtime') = DATE(?)", [field, val]),
```
pranavxc, line 763: Does all databases(sqlite, mysql and mssql) take `localtime` as second argument ?

cangku: MySQL was excluded from the beginning. Here, I am only testing SQLite and have not considered MSSQL. I can modify this through conditional statements.

cangku: his only fixes the issue. I feel this part can be abstracted and optimized further.

PR #8411, opened 2024-05-06
fix: Cmd K not working for viewer roles

DarkPhoenix2704:
## Change Summary

- [x] Shows only the bases, user have access to
- [x] removes root scopes and load all project data by default

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/src/services/command-palette.service.ts
+++ b/packages/nocodb/src/services/command-palette.service.ts
@@ -23,7 +23,7 @@ export class CommandPaletteService {
       const { scope } = param.body;
 
       if (scope === 'root') {
-        const bases = await Base.list({ user: param.user });
+        const bases = await BaseUser.getProjectsList(param.user.id, param);
 
         for (const base of bases) {
           cmdData.push({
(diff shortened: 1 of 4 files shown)


packages/nocodb/src/services/command-palette.service.ts
```diff
@@ -20,95 +20,73 @@ export class CommandPaletteService {
...
-                base_id: base.id,
-              },
-            },
-          });
-        }
-      } else if (scope.startsWith('p-')) {
-        const allBases = [];
-
-        const bases = await Base.list({ user: param.user });
+      allBases.push(...bases);
```
pranavxc, line 27: Here no longer checking the palette scope, is it intentional?

cc: @mertmit

mertmit: we don't need scopes yet, looks good for now

mertmit (commented):
This is not exactly on same scope but requires a tiny fix so better to include it, whenever we open cmd+k from settings page it tries to load account_settings which is no longer a valid parent, we should load root instead.
Can you please get rid of the the logic related to account_settings from packages/nc-gui/composables/useCommandPalette/index.ts

mertmit (approved):

PR #8437, opened 2024-05-09
nc-fix/ SingleLineText Fields silently dropping larger than length values

rohittp0:
## Change Summary

Added proper errror handling to `SingleLineText`
## Change type

image


packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -4512,6 +4513,20 @@ class BaseModelSqlv2 {
...
   }
 
+  /*
+   *  Utility method to validate database constraints
+   */
+  protected validateConstraints(
+    column: Column<any>,
+    data: Record<string, any>,
+  ) {
+    if (column.clen < data[column.title]?.length) {
```
pranavxc, line 4523: use `dtxp` here - `column.clen` => `+column.dtxp`

pranavxc: And do we have to do this for all types? in some cases, this value won't be defined

packages/nocodb-sdk/src/lib/sqlUi/MysqlUi.ts
```diff
@@ -1015,6 +1011,11 @@ export class MysqlUi {
         break;
       case 'SingleLineText':
         colProp.dt = 'varchar';
+        colProp.validate = {
+          func: ['isLength'],
+          args: [''],
+          msg: ['Validation failed : isLength ({cn})'],
+        };
```
pranavxc, line 1018: Is it required? We are not passing the max value argument at all here 

Ref: https://www.npmjs.com/package/validator

packages/nocodb/src/db/BaseModelSqlv2.ts
rohittp0: If the value is not defined the expression will evaluate to false right ?

packages/nocodb-sdk/src/lib/sqlUi/MysqlUi.ts
rohittp0: Oh, I thought it would infer max length if not passed. Will remove it

packages/nocodb/src/db/BaseModelSqlv2.ts
rohittp0: Added an explicit check to make it clear

packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -4512,6 +4514,20 @@ class BaseModelSqlv2 {
...
   }
 
+  /*
+   *  Utility method to validate database constraints
+   */
+  protected validateConstraints(
+    column: Column<any>,
+    data: Record<string, any>,
+  ) {
+    if (column.dtxp && column.dtxp < data[column.title]?.length) {
```
pranavxc, line 4524: Scale can be there for integer type as well. And if the value is a number/boolean `length` property won't be there.

rohittp0: If this pattern looks good, I can add in checks for all the field types

pranavxc: in the case of number, we need to confirm what it means(scale value), like the number of digits or bits

pranavxc: You can fix it for string at the moment by checking `typeof data[column.title] === 'string'`

packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -4512,6 +4514,24 @@ class BaseModelSqlv2 {
...
+  /*
+   *  Utility method to validate database constraints
+   */
+  protected validateConstraints(
+    column: Column<any>,
+    data: Record<string, any>,
+  ) {
+    if (
+      typeof data[column.title] === 'string' &&
+      column.dtxp &&
```
pranavxc, line 4526: for `enum` type in MySQL this dtxp value will be comma separated string value so we have to skip that case as well

pranavxc (approved):

Issue #8347, opened 2024-04-25
🔦 Feature: support read-only and non-root docker containers


tcurdt:
### Please confirm if feature request does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

For security reasons some companies require containers to be run as non-root and sometimes even in read-only mode. This usually is no bigger problem but nocodb seems to have a very unusual docker setup.

Setting this should work without bigger problems:

```
services:
  nocodb:
    container_name: nocodb
    image: nocodb/nocodb:latest
    read_only: true
    user: 65534:65534
```

### Suggested Solution

TBH I am not quite sure why an un-tar would be necessary at container startup at runtime over build time.
If possible the image should be setup correctly at build time.

And if some folders need special permissions this here could be a work around:

```
ENV UID 65534
ENV GID 65534
RUN chown -R $UID:$GID /some/folder
```

salim-b:
> TBH I am not quite sure why an un-tar would be necessary at container startup at runtime over build time.

It is not. Simply removing the untar (also from the [other start scripts](https://github.com/search?q=repo%3Anocodb%2Fnocodb%20tar%20-xzf&type=code))

https://github.com/nocodb/nocodb/blob/08dfcfa7d6ac49fb9358109c0e6d21a20fe3746a/packages/nocodb/docker/start-litestream.sh#L27-L30

and adding the following line to the Dockerfile at [this position](https://github.com/nocodb/nocodb/blob/08dfcfa7d6ac49fb9358109c0e6d21a20fe3746a/packages/nocodb/Dockerfile#L70) should suffice.

```Dockerfile
RUN tar -xzf /usr/src/appEntry/app.tar.gz -C /usr/src/app/ && rm -f /usr/src/appEntry/app.tar.gz
```

I wonder what @pranavxc thinks.

pranavxc:
> > TBH I am not quite sure why an un-tar would be necessary at container startup at runtime over build time.
> 
> It is not. Simply removing the untar (also from the [other start scripts](https://github.com/search?q=repo%3Anocodb%2Fnocodb%20tar%20-xzf&type=code))
> 
> https://github.com/nocodb/nocodb/blob/08dfcfa7d6ac49fb9358109c0e6d21a20fe3746a/packages/nocodb/docker/start-litestream.sh#L27-L30
> 
[...]

PR #8439, opened 2024-05-09
refactor: avoid unfavorable tar -> untar in container image

salim-b:
## Change Summary

Before this change, the NocoDB app code was GZIP-compressed to a tar archive during container image generation and extracted again on the first *run* of the app. This is unfavorable for a containerized app since the extraction is repeated on every container restart (unless `/usr/src/appEntry` lives on persistent storage), which delays startup. Besides, adding a tar archive in the container image to be extracted at run-time actually *in*creases the storage requirements for the app at run-time.

This change removes the compression and decompression steps and instead directly copies the app code as-is. It is a prerequisite for https://github.com/nocodb/nocodb/issues/8347.

Additionally, some [ShellCheck](https://www.shellcheck.net/) suggestions were applied to make the start scripts more robust / POSIX-compliant.

## Change type

## Test/ Verification

Untested.


packages/nocodb/Dockerfile.local
```diff
@@ -49,7 +49,7 @@ RUN apk --update --no-cache add \
     curl \
     jq
 
-# Copy packaged production code & main entry file
+# Copy production code & main entry file
 COPY --from=builder /usr/src/appEntry/ /usr/src/appEntry/
```
pranavxc, line 53: We have to copy build and `node_modules` to `/usr/src/app/`

`L#31` can be removed and here we can add

```
COPY --from=builder /usr/src/app/ /usr/src/app/
```

salim-b: Ah sorry, I mixed up `/usr/src/app/` and `/usr/src/appEntry/` 🙈. Should be fine now.

pranavxc (approved):

PR #8525, opened 2024-05-18
feat: move webhook handler to worker

mertmit:
## Change Summary

- move webhook handler to worker

## Change type


packages/nocodb/src/modules/jobs/jobs/webhook-handler/webhook-handler.processor.ts
```diff
@@ -0,0 +1,31 @@
...
+      job.data: {
+        hookName: string;
+        prevData;
+        newData;
+        user: UserType;
+        viewId: string;
+        modelId: string;
+        tnPath: string;
+      }
+    */
```
pranavxc, line 28: You can define the type and pass as generic type

```
job: Job<JobDataType>
```

mertmit: Done

pranavxc (approved):

PR #8526, opened 2024-05-19
feat: support Litestream age encryption

salim-b:
## Change Summary

Follow-up to https://github.com/nocodb/nocodb/pull/8494 that allows users to opt in to [Litestream's integrated age encryption](https://litestream.io/reference/config/#encryption) by simply setting the environment variables `LITESTREAM_AGE_PUBLIC_KEY` and `LITESTREAM_AGE_SECRET_KEY`.

Additionally, the [second commit](https://github.com/nocodb/nocodb/pull/8526/commits/ad3967b9c8f631762f5b5fef294b315dc41bc3fc) fixes variable names to match the documentation.

## Change type

## Test/ Verification

Successfully tested in own cloud deployment.

The CLI tool [dasel](https://daseldocs.tomwright.me/)  is used to [query and update the Litestream YAML config file](https://github.com/nocodb/nocodb/pull/8526/files#diff-e9696c304210e6036dc512394949b4fd6092d1423471b7fe335f89070efeaf3cR25-R34) before Litestream is started. This should be a robust solution that doesn't interfere should users still want to use a custom container build replacing the default `litestream.yaml` config for some reason.


packages/nocodb/docker/start-litestream.sh
```diff
@@ -13,25 +13,44 @@ use_litestream() {
   && [ -z "${NC_MINIMAL_DBS}" ] \
   && [ -n "${LITESTREAM_S3_ENDPOINT}" ] \
   && [ -n "${LITESTREAM_S3_BUCKET}" ] \
-  && [ -n "${LITESTREAM_ACCESS_KEY_ID}" ] \
-  && [ -n "${LITESTREAM_SECRET_ACCESS_KEY}" ]
+  && [ -n "${LITESTREAM_S3_ACCESS_KEY_ID}" ] \
+  && [ -n "${LITESTREAM_S3_SECRET_ACCESS_KEY}" ]
```
mertmit, line 17: I see this part is documented as LITESTREAM_S3_ACCESS_KEY_ID & LITESTREAM_S3_SECRET_ACCESS_KEY in our documentation but we should keep others to avoid existing installations.

If possible we can replace it to support both of them:
```suggestion
  && ( ([ -n "${LITESTREAM_S3_ACCESS_KEY_ID}" ] && [ -n "${LITESTREAM_S3_SECRET_ACCESS_KEY}" ]) || \
       ([ -n "${LITESTREAM_ACCESS_KEY_ID}" ] && [ -n "${LITESTREAM_SECRET_ACCESS_KEY}" ]) )
```

cc: @pranavxc

salim-b: Those env vars were only introduced with https://github.com/nocodb/nocodb/pull/8494 and it was my fault that they didn't match the documentation. Before #8494, they [were actually named `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`](https://github.com/nocodb/nocodb/pull/8494/files#diff-e9696c304210e6036dc512394949b4fd6092d1423471b7fe335f89070efeaf3cL9).

If at all, it would probably make more sense to keep backwards-compatbility with `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` – but not `LITESTREAM_ACCESS_KEY_ID` and `LITESTREAM_SECRET_ACCESS_KEY` which were only part of stable NocoDB since v0.207.2.

pranavxc: @salim-b Then let's keep support to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and mark it as deprecated in docs.

salim-b: @pranavxc Ok, done in https://github.com/nocodb/nocodb/pull/8526/commits/2480c0865da3e084f2deea12e4dc6078a409dc86.

pranavxc (approved):

mertmit (approved):

PR #8541, opened 2024-05-21
fix: bigInt issues

### LINKED ISSUE #3849: 🐛 Bug: bigint is not correcting cast in Number

### Steps to reproduce ?

1. I have data like "363228830310334464" in mysql as bigint type.
2. I use Navicat query the data is display "363228830310334464".
3. But when I use nocodb connect with the same db, the data is cast to "363228830310334460", precision loss..

### Desired Behavior

I hope to get your answer as soon as possible

Thank you very much for your help

### Project Details

mysql 5.7
nocodb 0.97

### Attachements

_No response_

mertmit:
## Change Summary

Fixes #3849
- add bigInt support to mysql on driver level
- add support to bigInt for Number component

## Change type


dstala (changes requested):
https://github.com/nocodb/nocodb/issues/3849#issuecomment-2127259697

dstala (commented):
https://github.com/nocodb/nocodb/assets/86527202/dbfe8e7f-a949-4805-b716-886fe8372904

pranavxc:
In Webhook filter and frontend grid filter logic we need to cast value to Bigint while comparing.

PR #8555, opened 2024-05-23
fix: Litestream S3 - add support to deprecated env variables

### LINKED ISSUE #8549: 🐛 Bug: Persistence not backing up to S3 on latest version of Noco
### Please confirm if bug report does NOT exist already ?

### Steps to reproduce ?

Deploy as ECS task on FARGATE 

Set the following variables
NC_S3_BUCKET_NAME | AWS S3 bucket name for S3 storage plugin. |  
NC_S3_REGION | AWS S3 region for S3 storage plugin. |  
NC_S3_ACCESS_KEY | AWS access key ID for S3 storage plugin. |  
NC_S3_ACCESS_SECRET | AWS access secret for S3 storage plugin.

Deploy a new service 

Am I missing something , all my permissions are set correctly as this worked previously for me  with 

AWS_ACCESS_KEY | AWS access key ID for S3 storage plugin. |  
AWS_ACCESS_SECRET | AWS access secret for S3 storage plugin.

This is now deprecated 

### Desired Behavior

Folder with persistence should be created on S3

This worked fine on the older version

### Project Details

Noco DB used on AWS ECS Fargate

### Attachments

_No response_

pranavxc:
## Change Summary

- Closes #8549

## Change type

## Test/ Verification

Provide summary of changes.


mertmit (approved):

packages/nocodb/docker/start-litestream.sh
```diff
@@ -19,14 +26,16 @@ use_litestream() {
...
-  && [ -n "${LITESTREAM_S3_ENDPOINT}" ] \
   && [ -n "${LITESTREAM_S3_BUCKET}" ] \
   && [ -n "${LITESTREAM_S3_ACCESS_KEY_ID}" ] \
   && [ -n "${LITESTREAM_S3_SECRET_ACCESS_KEY}" ]
 }
 
 if use_litestream ; then
 
+  # set default bucket path if not provided
+  : "${LITESTREAM_S3_PATH:=nocodb}"
```
salim-b, line 37: To be consistent, I would rather set this in the [`Dockerfile` where the other Litestream-specific defaults are set](https://github.com/nocodb/nocodb/blob/f441403abadb91ad3d07cd9dd943b626777a8571/packages/nocodb/Dockerfile#L53).

packages/nocodb/docker/litestream.yml
```diff
@@ -6,6 +6,7 @@ dbs:
       - type: s3
         endpoint: ${LITESTREAM_S3_ENDPOINT}
         force-path-style: true
+        region: ${LITESTREAM_S3_REGION}
```
salim-b, line 9: @pranavxc I think setting `region` does only make sense when ***not*** setting `endpoint` at the same time since `endpoint` already includes the region, cf. https://litestream.io/reference/config/#s3-replica.

Did you test what happens if `LITESTREAM_S3_REGION` is unset and `LITESTREAM_S3_ENDPOINT` is set and vice versa? Does Litestream ignore empty config entries?

packages/nocodb/docker/start-litestream.sh
```diff
@@ -11,6 +11,13 @@ fi
 if [ -z "${LITESTREAM_S3_SECRET_ACCESS_KEY}" ] && [ -n "${AWS_SECRET_ACCESS_KEY}" ] ; then
   export LITESTREAM_S3_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
 fi
+if [ -z "${LITESTREAM_S3_PATH}" ] && -n [ "${AWS_BUCKET_PATH}" ] ; then
+  export LITESTREAM_S3_PATH="${AWS_BUCKET_PATH}"
+fi
+if [ -z "${LITESTREAM_S3_BUCKET}" ] && [ -n "${AWS_BUCKET}" ] ; then
+  export LITESTREAM_S3_BUCKET="${AWS_BUCKET}"
```
salim-b, line 18: I guess, we should also [document](https://docs.nocodb.com/getting-started/self-hosted/environment-variables/) that `AWS_BUCKET_PATH` and `AWS_BUCKET` are deprecated, same as for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.

salim-b (commented):
@pranavxc Some comments about your recent changes (I'm watching the these files via GitHub's atom feeds, so I ot notified about your recent changes 😉).

If you'd like, I can prepare a follow-up PR with my suggestions. 🙂

packages/nocodb/docker/start-litestream.sh
```diff
@@ -19,14 +26,16 @@ use_litestream() {
   && [ -z "${DATABASE_URL}" ] \
   && [ -z "${DATABASE_URL_FILE}" ] \
   && [ -z "${NC_MINIMAL_DBS}" ] \
-  && [ -n "${LITESTREAM_S3_ENDPOINT}" ] \
```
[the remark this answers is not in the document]
pranavxc: It fallbacks to the default region or endpoint if not set so kept as not mandatory

packages/nocodb/docker/start-litestream.sh
pranavxc: If we set it at docker file it's hard to add backward compatibility for env var `AWS_BUCKET_PATH`

packages/nocodb/docker/litestream.yml
pranavxc: I tested and it was fine, will test again to verify

packages/nocodb/docker/start-litestream.sh
pranavxc: Agree and thanks for pointing out.

packages/nocodb/docker/litestream.yml
salim-b: Note that I hardcoded `force-path-style: true` in the Dockerfile.

About `force-path-style`:

> Uses the path style which is required by non-AWS services. This is automatically enabled if `endpoint` is set.

So I guess we'd have to remove `force-path-style: true` if we wanna support `region`, no?

(In the end I don't think it's really worth to support `region` since it only brings a tiny bit of convenience for AWS and Backblaze users but complicates support for other S3-compatible hosters.)

packages/nocodb/docker/start-litestream.sh
```diff
@@ -11,6 +11,13 @@ fi
 if [ -z "${LITESTREAM_S3_SECRET_ACCESS_KEY}" ] && [ -n "${AWS_SECRET_ACCESS_KEY}" ] ; then
   export LITESTREAM_S3_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
 fi
+if [ -z "${LITESTREAM_S3_PATH}" ] && -n [ "${AWS_BUCKET_PATH}" ] ; then
```
salim-b, line 14: @pranavxc This seems to be invalid syntax, the `-n` test needs to be inside the brackets.

packages/nocodb/docker/litestream.yml
pranavxc: Even with `force-path-style:` region was working for me.

packages/nocodb/docker/start-litestream.sh
pranavxc: Possibly I will create a follow up PR with your suggested changes.

packages/nocodb/docker/start-litestream.sh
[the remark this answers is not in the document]
salim-b: What are the default region and endpoint? Where are they defined?I didn't spot anything about this in the [Litestream doc](https://litestream.io/reference/config/)...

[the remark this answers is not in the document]
pranavxc: Even I couldn’t find that details but it fallbacks to us-east-1 I guess

packages/nocodb/docker/litestream.yml
salim-b: > Even with `force-path-style:` region was working for me.

Ok, interesting...

> I kept it because setting region is easier than figuring out the endpoint.

I see^^

I still think it would be best to adhere to what the official Litestream doc says, hence remove `force-path-style: true` if we keep `region` :)

packages/nocodb/docker/start-litestream.sh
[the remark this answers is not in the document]
salim-b: Hm, ok. We should then update the env var documentation if we keep it that way.

packages/nocodb/docker/litestream.yml
pranavxc: Ya better to remove since it automatically going to set to true when we are configuring custom endpoint

packages/nocodb/docker/start-litestream.sh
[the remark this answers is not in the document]
pranavxc: I didn’t added that info since I couldn’t find that info anywhere in their docs

packages/nocodb/docker/start-litestream.sh
salim-b: Could you ping me when you submit that follow-up PR? 🙂

pranavxc: In between something, so once sorted out, I will work on it. I’m happy if you would like to open a PR as well.

salim-b: I've just submitted https://github.com/nocodb/nocodb/pull/8572 which should address everything we discussed here (expect for testing) :)

pranavxc: Thanks for creating the PR, soon I will test it and merge.

PR #8595, opened 2024-05-28
fix: Gallery View delete using expanded form

rohittp0:
## Change Summary

Fixed record not found error when deleting using expanded form of Gallery View
## Change type

### DIFF
--- a/packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
+++ b/packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
@@ -136,6 +136,14 @@ const pagination = computed(() => {
     extraStyle: `margin-left: ${scrollBump.value}px;`,
   }
 })
+
+async function deleteSelectedRowsWrapper() {
+  if (!deleteSelectedRows) throw new Error('Feature not implemented')
+
+  await deleteSelectedRows()
+  // reload table data
+  await reloadTableData({ shouldShowLoading: true })
+}
 </script>
 
 <template>
@@ -151,7 +159,7 @@ const pagination = computed(() => {
     :expand-form="props.expandForm"
     :row-height="rowHeight"
     :delete-row="deleteRow"
-    :delete-selected-rows="deleteSelectedRows"
+    :delete-selected-rows="deleteSelectedRowsWrapper"
     :delete-range-of-rows="deleteRangeOfRows"
     :update-or-save-row="updateOrSaveRow"
     :remove-row-if-new="removeRowIfNew"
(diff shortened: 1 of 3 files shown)


dstala (approved):

dstala:
Node: **v18.19.1**
Arch: **arm64**
Platform: **darwin**
Docker: **false**
RootDB: **pg**
PackageVersion: **0.207.2**

In grid + groupBy, triggering bulk delete for a record keeps loading forever. record is deleted on page reload
@rohittp0

packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
```diff
@@ -136,6 +136,14 @@ const pagination = computed(() => {
     extraStyle: `margin-left: ${scrollBump.value}px;`,
   }
 })
+
+async function deleteSelectedRowsWrapper() {
+  if (!deleteSelectedRows) throw new Error('Feature not implemented')
+
+  await deleteSelectedRows()
+  // reload table data
+  await reloadTableData({ shouldShowLoading: true })
```
mertmit, line 145: What is the expected UX here @dstala, throwing will cause error modal to be shown and make UI unusable until refresh if I am not missing something

packages/nc-gui/composables/useData.ts
```diff
@@ -732,7 +732,7 @@ export function useData(args: {
 
     await callbacks?.syncCount?.()
     await callbacks?.syncPagination?.()
-    await callbacks?.globalCallback?.()
+    callbacks?.globalCallback?.()
```
mertmit, line 735: What is the purpose of this change? This is a shared composable so is changing callback logic a good idea?

packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
rohittp0: It should not throw. the `deleteSelectedRows` should always be set. The check is there just to get rid of TS warning.

packages/nc-gui/composables/useData.ts
rohittp0: The type definition of `globalCallback` is `globalCallback?: (...args: any[]) => void`. It is not async. Removed the `await` to handle TS warning. I think the `await` was there by mistake.

mertmit: Please keep await and replace type definition to expect a promise instead, having await on a non-promise will not have any significant difference but counter-part can cause unexpected results (as we await wrapper function everything should be handled before going to next step)

packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
mertmit: Please use following as this is the common pattern we follow
```
if (!deleteSelectedRows) {
  return
}
```

mertmit (approved):

PR #8597, opened 2024-05-28
Convert duration to decimal before saving to db

rohittp0:
## Change Summary

During bulk operations duration field need to be parsed if the provide value is in time stamp notation ( `hh:mm:ss` )
## Change type


dstala (commented):
TS compilation error needs a fix
TSError: ⨯ Unable to compile TypeScript:
Error: src/db/BaseModelSqlv2.ts(28,8): error TS2307: Cannot find module 'nc-gui/utils/durationUtils' or its corresponding type declarations.

mertmit (changes requested):

packages/nocodb/src/db/util/durationUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  return {
+    h: Number.parseFloat(groups[1]) || 0,
+    mm: Number.parseFloat(groups[2]) || 0,
+    ss: Number.parseFloat(groups[3]) || 0,
+  };
+};
+
+export const convertDurationToSeconds = (val: string) => {
+  const { regex, type } = durationOptions.find(({ regex }) => regex.test(val));
```
rameshmane7218, line 80: @rohittp0 It might be a good idea to move the `durationUtils.ts` file from the frontend `/nc-gui` package to the `nocodb-sdk` and customize the convertMS2Duration function according to the use case. 
For instance, if the duration type is `undefined`, find it from `durationOptions`. This will ensure that there is only one source of truth and will reduce code duplication.

*Note*: Also after moving this utils file to `nocodb-sdk` make sure to import variable/function (`durationOptions`, `convertMS2Duration`, `convertDurationToSeconds`) in respective file of `nc-gui`

rameshmane7218 (changes requested):

packages/nocodb/src/db/util/durationUtils.ts
rohittp0: I found no existing usage for the Date Utils in the frontend. Is there any usage or should I remove it from frontend

image
image

rameshmane7218: It's already used in frontend directly without import as this is a part of global import (try searching globally).

Screenshot 2024-05-29 at 1 58 00 AM
Screenshot 2024-05-29 at 1 57 43 AM
Screenshot 2024-05-29 at 1 57 30 AM

rohittp0: Oh sure will refactor those manually

rameshmane7218 (changes requested):
Fix the changes requested via Slack.

Issue #8194, opened 2024-04-07
🐛 Bug: baseUpdate - Base Editor: Not allowed


TangXY123:
### Please confirm if bug report does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

1. There are two users A and B，they are "Organization Level Creator".
2. A create 2 bases, grants "creator" role of base1 to B, and grants "viewer"(non-creator) role of base2 to B.
3. Login the system with B, they the system throw out a exception,"baseUpdate - Base Editor: Not allowed"

![1](https://github.com/nocodb/nocodb/assets/48933981/dd5baab9-53da-45be-8a43-b6f46d8c1aec)
![2](https://github.com/nocodb/nocodb/assets/48933981/102e110f-2153-479e-82d6-8e56358aa879)
![3](https://github.com/nocodb/nocodb/assets/48933981/71fcf29f-1293-42b6-8180-24c5cffe5241)
![4](https://github.com/nocodb/nocodb/assets/48933981/457eaef8-384a-4486-88df-cd91ae6cb9e0)

### Desired Behavior

x

### Project Details

x

TangXY123:
This bug exists in version 0.204.4-0.205.0, at least

PR #8600, opened 2024-05-28
Fix "Base Editor: Not allowed" error on sign

rohittp0:
## Change Summary
closes: #8194 
The permission being checked in the frontend and that being checked by the corresponding controller were different. Updated frontend to align with backend.

## Change type

### DIFF
--- a/packages/nc-gui/store/bases.ts
+++ b/packages/nc-gui/store/bases.ts
@@ -310,7 +310,7 @@ export const useBases = defineStore('basesStore', () => {
   }
 
   async function updateIfBaseOrderIsNullOrDuplicate() {
-    if (!isUIAllowed('baseReorder')) return
+    if (!isUIAllowed('baseUpdate')) return
 
     const basesArray = Array.from(bases.value.values())
 


dstala (approved):

jessystecroix717:
u

mertmit:
I don’t think this resolves the issue, also explanation in body is wrong because we have different acl for frontend & backend.

We need to take an in-depth look on our handling here, so lets go through this:
- there is one root level which is org
- every base is within org
- org role is inherited only if no role set at base level
- base order is kept at base entity itself *** -> this is the crucial part
So logically a org creator sounds like it should be able to reorder the bases but it is not confirmed for an org creator to have access to all bases (unless he is super-admin)

This causes two issues:
- There will be some invisible bases for an org creator, so order will be unpredictable on change
- Even if you are org creator, if you have assigned a lower role for a specific base you won’t be able to change order as you don’t allowed to update

So the change only hide the initial issue, but if you try to reorder (base with editor role) using that role you can see the error.

cc @dstala

rohittp0:
@mertmit So are you suggesting removing base reorder permission for org-level creators? Or should that permission be replaced with base edit, as users with base edit access would be guaranteed to have reordered permission?

dstala:
@mertmit 
- allow users with ORG-Creator (OC) to re-order any bases (including the one's in which this user has base-viewer access).
- prevent ORG-Viewer (OV) from re-ordering bases.

Can base-reorder request from OC be given an exception on Base viewer permission level to still over ride base order information?

In cloud, we apply same principle to WS-Creators. WS-Creator (+ WS Owner) will be able to apply re-order to any base.

We understand that, such action would be
- visible to everyone. reorder applies to all
- and depending on the bases that user has access to, the end visible result might differ.

mertmit:
Yes we can provide an exception.
Even I would suggest creating a new endpoint for this operation reason being right now we calculate order from frontend and provide it with payload but some of the bases might not be visible for the user causing miss calculations, in the new endpoint we can have a payload like `{ id: string; before_id?: string; after_id?: string }` then we can calculate order based on provided anchor base.
This way we will avoid having duplicate orders for bases.

Issue #8235, opened 2024-04-10
🐛 Bug: Some rows doesn't have data when draging in the bottom right corner of the cell to copy data


TangXY123:
### Please confirm if bug report does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Some rows doesn't have data when draging in the bottom right corner of the cell to copy data. we have to drag two times.
![image](https://github.com/nocodb/nocodb/assets/48933981/582c456b-c18a-4ff2-8101-668b5406018f)
![image](https://github.com/nocodb/nocodb/assets/48933981/90b3ec6c-7e00-439f-94d9-a5af9cf65102)

### Desired Behavior

x

### Project Details

- docker build
- version 0.205.0
- mysql8.0

rohittp0:
[recording-27_05_2024 (1).webm](https://github.com/nocodb/nocodb/assets/45072928/135b8e96-49b3-4fc9-8143-34c4657fbba7)

Testing on the latest develop branch with MySQL. The behaviour was same with Postgres as well

TangXY123:
> recording-27_05_2024.1.webm 
> Testing on the latest develop branch with MySQL. The behaviour was same with Postgres as well

hi, rohittp0, I notice this issue may have been fixed in 0.207.1.(which is found in 0.205.0)

dstala:
@rohittp0 
try the staging base here https://staging.noco.to/#/wqd7e84kkpg8169/pwvt6deb9owszvl/ms6qjj4fk2bgk18
https://github.com/nocodb/nocodb/assets/86527202/f1756282-a099-4755-aacd-ca0bca1e2188

PR #8605, opened 2024-05-29
fix: use xc table for at import

mertmit:
## Change Summary

closes: #8235 

- Use ai ID
- Restrict editing of system columns

## Change type


dstala (commented):
- [ ] two fields with same name `Id`. scenario needs to be handled (known issue, as discussed during call)
Screenshot 2024-05-30 at 12 42 05 PM


- [ ] data import failed
Screenshot 2024-05-30 at 12 31 14 PM

mertmit:
Data import part is a side-effect of system field update restrictions, I will think of a solution.

For duplicate id -> we should decide approach:
- Rename our fields
- Rename incoming fields

need your input on this @dstala

dstala:
- Let ID field be a system keyword in NocoDB. No other field can have ID field
- Rename Airtable's field with title "ID" to "ID1".

packages/nc-gui/components/dlg/AirtableImport.vue
```diff
@@ -403,22 +403,30 @@ function downloadLogs(filename: string) {
...
+            <a-tooltip placement="top">
+              <template #title>
+                <span>{{ $t('title.comingSoon') }}</span>
+              </template>
+              <a-checkbox v-model:checked="syncSource.details.options.syncUsers" disabled>
+                {{ $t('labels.importUsers') }}
+              </a-checkbox>
+            </a-tooltip>
+          </div>
+          -->
```
pranavxc, line 429: Can be removed if no longer needed or going to enable in future.

packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -3218,6 +3218,12 @@ class BaseModelSqlv2 {
         { ignoreView: true, getHiddenColumn: true },
       );
 
+      if (!prevData) {
+        NcError.recordNotFound(
+          typeof id === 'string' ? id : JSON.stringify(id),
```
pranavxc, line 3223: If possible let's move the `stringify` part to `recordNotFound` method.

pranavxc (approved):

packages/nc-gui/components/dlg/AirtableImport.vue
mertmit: We've disabled for now as it is not functioning properly, there is a possibility to re-enable

pranavxc (approved):

PR #8609, opened 2024-05-29
Decode field name as utf-8

rohittp0:
## Change Summary

The field names in file attachments were being decoded incorrectly. Added proper `UTF-8` decoding to ensure non-english field names are handled properly.

## Change type

## Additional information / screenshots

[recording-29_05_2024.webm](https://github.com/nocodb/nocodb/assets/45072928/d38b2cb6-e951-4095-add6-a0d47cce16b1)

### DIFF
--- a/packages/nocodb/src/services/public-datas.service.ts
+++ b/packages/nocodb/src/services/public-datas.service.ts
@@ -307,7 +307,9 @@ export class PublicDatasService {
 
     for (const file of param.files || []) {
       // remove `_` prefix and `[]` suffix
-      const fieldName = file?.fieldname?.replace(/^_|\[\d*]$/g, '');
+      const fieldName = Buffer.from(file?.fieldname || '', 'binary')
+        .toString('utf-8')
+        .replace(/^_|\[\d*]$/g, '');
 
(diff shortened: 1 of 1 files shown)


pranavxc (approved):

pranavxc (approved):

packages/nocodb/src/services/public-datas.service.ts
```diff
@@ -307,7 +307,9 @@ export class PublicDatasService {
 
     for (const file of param.files || []) {
       // remove `_` prefix and `[]` suffix
-      const fieldName = file?.fieldname?.replace(/^_|\[\d*]$/g, '');
+      const fieldName = Buffer.from(file?.fieldname || '', 'binary')
+        .toString('utf-8')
+        .replace(/^_|\[\d*]$/g, '');
```
pranavxc, line 313: Even `base.title` and `model.title` can contain different language characters so let's handle that as well.

rohittp0: Those seem to be already using `utf-8` the request body is also properly decoded. It is only the `files` field that is not working.

PR #8666, opened 2024-06-06
feat: query isolation

mertmit:
## Change Summary

- Enforce base_id for every query

## Change type


dstala:
# SQLite
- after migration, base_id field was not populated in barcode, qrcode, formula, lookup, rollup, select options, relations table. was fine in calendar range table
```
TypeError: Cannot read properties of null (reading 'getRelatedTable')
    at /Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/helpers/getAst.ts:191:34
    at async /Users/rajuudava/NocoDB/nocodb/packages/nocodb/src/helpers/getAst.ts:189:21
[nodemon] app crashed - waiting for file changes before starting...
```

dstala (approved):

PR #8695, opened 2024-06-11
fix: move get users list to promise

mertmit:
## Change Summary

- Move get users list to promise
- Use obj args for _formulaQueryBuilder

## Change type


packages/nocodb/src/db/formulav2/formulaQueryBuilderv2.ts
```diff
@@ -58,16 +59,30 @@ const getAggregateFn: (fnName: string) => (args: { qb; knex?; cn }) => any = (
...
-  alias,
-  model: Model,
-  aliasToColumn: Record<string, () => Promise<{ builder: any }>> = {},
-  tableAlias?: string,
-  parsedTree?: any,
-  column: Column = null,
-) {
+async function _formulaQueryBuilder(params: {
+  baseModelSqlv2: BaseModelSqlv2;
+  _tree;
```
pranavxc, line 64: we can rename `_tree` as `tree`

pranavxc (approved):

Issue #7903, opened 2024-03-19
🔦 Feature: Improve object storage provider configuration


salim-b:
### Please confirm if feature request does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

Currently, NocoDB supports the following (S3-compatible) object storage providers besides Amazon S3 / AWS:

- Backblaze B2
- DigitalOcean Spaces
- Google Cloud Storage
- Linode Object Storage
- MinIO
- OVHcloud Object Storage
- Scaleway Object Storage
- UpCloud Object Storage
- Vultr Object Storage

This is great! But configuration is not.

The current configuration has the following deficiencies:

- ~~It is unclear to the user what happens (i.e. where file attachments are stored) when *multiple* of these providers are configured.~~ *Since NocoDB v0.255.2 it's not possible anymore to configure multiple providers at the same time.*
- ~~It is not possible to configure any of the above providers via environment variables. Only Amazon S3 can be configured via the [`NC_S3_*`](https://docs.nocodb.com/getting-started/self-hosted/environment-variables/) variables.~~ *Possible since NocoDB v0.255.0*
- ~~"App store" is not a good notion/metaphor to configure these plugins, but this is just my personal opinion.~~ *The "app store" has been deprecated with NocoDB v0.255.2 and will be removed in a future release. E-mail and storage are now configured via "Account" -> "Setup" which should be more intuitive .*

### Suggested Solution

- ~~It should be possible to configure *any* object storage provider via environment variables. Either the `NC_S3_*` variables could be used in a provider-independent way (with an additional `NC_S3_PROVIDER` variable indicating the provider) or per-provider environment variables could be introduced. The latter would give more flexibility with regard to multiple simultaneous object storage providers.~~ *Possible since NocoDB v0.255.0, see [the docs](https://docs.nocodb.com/getting-started/self-hosted/environment-variables/#storage).*

- It should be possible to choose the *preferred* object storage provider in case multiple providers are configured (both via UI and via an environment variable).

- Ideally, it should be possible to use multiple different object storage providers simultaneously and choose the provider per attachment-column (or per table or per base).

### Related

- https://github.com/nocodb/nocodb/issues/4993
- https://github.com/nocodb/nocodb/issues/3402

salim-b:
[...]

Issue #4993, opened 2023-02-01
🔦 Feature: Add S3 endpoint & S3 path style option to env vars


loulou2852:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

I'm using an alternative S3 provider (Contabo object storage), but actually, there is no custom endpoints in env vars. (ie NC_S3_ENDPOINT).
This allows integrating other S3 compatible providers in Nocodb.

Furthermore, Contabo use “path style” type of request. 
So maybe including an option to set a path style to true via NC_S3_PATH_STYLE var.

### Suggested Solution

Ideally,

Adding to environment vars NC_S3_ENDPOINT & NC_S3_PATH_STYLE, so that other S3 providers can connect to Nocodb.

### Additional Context

If necessary, I can provide access to a Contabo Bucket if needed in private for testing purpose.

loulou2852:
It can be an addition of another plugin similar to S3, with these two parameters added.

AzazKamaz:
For specifying endpoint there exists Minio plugin. [Minio](https://min.io) is basically self hosted S3 compatible storage, so you can point that plugin to any S3 you want

loulou2852:
![](https://user-images.githubusercontent.com/4818634/217372164-9a900eb7-cb61-4a2a-ae1b-50fe2b3bf5d1.png)

I tested with the MINIO one, the file is saved in the object sotrage, but due to lack of path style setting missing or something else, it can't detect the image within nocodb.

this is the 404 url I've got from Nocodb screenshot cell: [https://eu2.contabostorage.com:443/nocodb/nc/uploads/noco/Test/Content/file/f_J0q9uxQgR-xo0jg6.jpg](https://eu2.contabostorage.com:443/nocodb/nc/uploads/noco/Test/Content/file/f_J0q9uxQgR-xo0jg6.jpg)

This is a working url from contabstorage:

[https://eu2.contabostorage.com/84721fb304514bc6a1901ae7e2107411:nocodb/nc/uploads/noco/Test/Content/file/f_J0q9uxQgR-xo0jg6.jpg](https://eu2.contabostorage.com/84721fb304514bc6a1901ae7e2107411:nocodb/nc/uploads/noco/Test/Content/file/f_J0q9uxQgR-xo0jg6.jpg)

And this is the settings:

[...]

PR #8886, opened 2024-06-27
feat: add S3 endpoint input

jgillich:
## Change Summary

This change adds a **S3 endpoint** input value and environment variable for the S3 plugin.

Why? Endpoint can be set on the Minio plugin already, however that one does not support environment variables. Since the S3 protocol is supported by many providers, many of them not using Minio, it makes sense to have this option on the S3 plugin as well.

This change is backwards-compatible, the default endpoint is AWS.

## Change type

- [X] feat: (new feature for the user, not a new feature for build script)

## Test/ Verification

Tested and working with Scaleway object storage

Relevant issues: #4993, #7903

### DIFF
--- a/packages/nocodb/src/plugins/s3/S3.ts
+++ b/packages/nocodb/src/plugins/s3/S3.ts
@@ -5,6 +5,7 @@ import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
 import { Upload } from '@aws-sdk/lib-storage';
 import axios from 'axios';
 import { useAgent } from 'request-filtering-agent';
+import type { S3ClientConfig } from '@aws-sdk/client-s3';
 import type { IStorageAdapterV2, XcFile } from 'nc-plugin';
 import type { Readable } from 'stream';
 import { generateTempFilePath, waitForStreamClose } from '~/utils/pluginUtils';
@@ -128,13 +129,17 @@ export default class S3 implements IStorageAdapterV2 {
     // s3Options.accessKeyId = process.env.NC_S3_KEY;
     // s3Options.secretAccessKey = process.env.NC_S3_SECRET;
 
-    const s3Options = {
+    const s3Options: S3ClientConfig = {
       region: this.input.region,
       credentials: {
         accessKeyId: this.input.access_key,
         secretAccessKey: this.input.access_secret,
       },
     };
+
+    if (this.input.endpoint) {
+      s3Options.endpoint = this.input.endpoint;
+    }
 
     this.s3Client = new S3Client(s3Options);
   }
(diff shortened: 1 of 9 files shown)


jgillich:
I noticed the field is not being shown when upgrading from an older version, is something else needed to make that happen?

salim-b:
The newly introduced `NC_S3_ENDPOINT` env var should be [documented](https://docs.nocodb.com/getting-started/self-hosted/environment-variables/) (in [`packages/noco-docs/docs/020.getting-started/050.self-hosted/020.environment-variables.md`](https://github.com/nocodb/nocodb/blob/develop/packages/noco-docs/docs/020.getting-started/050.self-hosted/020.environment-variables.md)).

jgillich:
@salim-b Thanks, done

Also fixed merge conflicts. Sorry for ignoring Code Rabbit but all of its suggestions range from pointless to potentially dangerous. A stanard linter would do a much better job tbh

vjm:
i would love to see this merged in order to use minio -- is there a path to get this merged?

packages/nocodb/src/plugins/s3/S3.ts
```diff
@@ -128,14 +132,19 @@ export default class S3 implements IStorageAdapterV2 {
     // s3Options.accessKeyId = process.env.NC_S3_KEY;
     // s3Options.secretAccessKey = process.env.NC_S3_SECRET;
 
-    const s3Options = {
+    const s3Options: S3ClientConfig = {
       region: this.input.region,
+      endpoint: this.input.endpoint,
```
mertmit, line 137:
```suggestion
```

Please remove this line as you are checking & appending it on next block

packages/nocodb/src/plugins/s3/S3.ts
```diff
@@ -168,7 +177,10 @@ export default class S3 implements IStorageAdapterV2 {
       const data = await upload.done();
 
       if (data) {
-        return `https://${this.input.bucket}.s3.${this.input.region}.amazonaws.com/${uploadParams.Key}`;
+        let endpoint = this.input.endpoint
```
mertmit, line 180:
```suggestion
        const endpoint = this.input.endpoint
```
Please use const here instead of let

mertmit (changes requested):
Hi @jgillich thank you for your contribution, I've just rebased your code to latest develop and asked some quick changes. Please see those when possible so we can continue with the merge process.

jgillich:
@mertmit Thank you, I implemented your suggestions

mertmit (approved):

pranavxc (approved):

PR #8890, opened 2024-06-28
feat: export as on background

mertmit:
## Change Summary

- Revise Jobs
  - Added `nc_jobs` table for job meta
  - Added Job to models
  - Added JobMeta service & controller (avoided Job name here as they are used within jobs module)
  - Integrated it to automatically populate (ids will match, status will be updated, return of processor will be stored in result)
  - Defined payload type for each job (context & user are common properties for user jobs)
  - New endpoint for fetching jobs within base /api/v2/jobs/{baseId}
- Added support for filename override for download endpoints (as well as secure endpoint)
- Added data-export job (CSV only for now, but expects filetype for future)
  - New endpoint for triggering this job /api/v2/export/{viewId}/{exportAs}
- Added data export extension

## Change type


c2006878:
how to use this feature?

PR #8898, opened 2024-06-30
Aggregation tests

DarkPhoenix2704:


dstala:
- [ ] SQLite, number field with 3 records [42, 100, empty cell] => standard deviation is 23.7; hide record containing 0, standard deviation for [42, empty cell] should be 0. We continue to display `23.7`

DarkPhoenix2704:
Unit Tests exists for aggregation. 
We can ignore pw for this

PR #9011, opened 2024-07-16
TAS-610 Use Sofia Pro Font

expexpert:
## Change Summary

## Change type

- [ ] feat: Use Sofia Pro Font
- Task link
https://linear.app/taskmagic/issue/TAS-610/use-sofia-pro-font


dstala:
not planned unless we have specific issues with current font.

PR #9062, opened 2024-07-24
fix: use original url if signed url not available

### LINKED ISSUE #9057: 🐛 Bug: NocoDB v0.251.2 broke attachment display (at least with Backblaze B2)

### Steps to reproduce

I noticed a serious regression with NocoDB **v0.251.2**: When using Backblaze B2 (configured via plugin of the same name) to store attachment files, the files are uploaded fine but not displayed.

The screencast below illustrates the bug.

In this example, I've uploaded an image file named `penguins inside.png`. It gets uploaded to the configured Backblaze B2 bucket fine and the following value is written in the `attachment` field, which is also fine:

```json
[
  {
    "url": "https://rdb-attachments.s3.eu-central-003.backblazeb2.com/nc/uploads/noco/p028vhq05d3je77/m50bn1vm4bpdouo/c8sy6fecvfdtvmo/penguins%20inside_rheSh.png",
    "title": "penguins inside.png",
    "mimetype": "image/png",
    "size": 46981
  }
]
```

In the web browser console I see that the NocoDB UI tries to load the image from the following URLs which both fail (HTTP 404):

- `https://admin.rdb.vote/_ipx/q_50&blur_3&s_10x10/dltemp/GaFRWvXvWubEBdgS/1721783400000/nc/uploads/noco/p028vhq05d3je77/m50bn1vm4bpdouo/c8sy6fecvfdtvmo/penguins%20inside_rheSh.png`
- `https://admin.rdb.vote/_ipx/q_75/dltemp/GaFRWvXvWubEBdgS/1721783400000/nc/uploads/noco/p028vhq05d3je77/m50bn1vm4bpdouo/c8sy6fecvfdtvmo/penguins%20inside_rheSh.png`

I did not test any other S3-compatible object stores.

### Desired Behavior

Attachments should be displayed correctly (as with NocoDB v0.251.1 and below).

### Project Details

Node: **v20.15.1**
Arch: **x64**
Platform: **linux**
Docker: **false**
RootDB: **sqlite3**
PackageVersion: **0.251.2**

I also tried the `nocodb/nocodb-daily:0.251.1-20240723-1803` docker image, same result.

### Attachments

[Bildschirmaufzeichnung vom 24.07.2024, 01:04:03.webm](https://github.com/user-attachments/assets/5672b6bf-b0cb-4f45-8b46-879cfc261039)

mertmit:
## Change Summary

Fixes #9057

- use original url if signed url not available

## Change type

### DIFF
--- a/packages/nocodb/src/models/PresignedUrl.ts
+++ b/packages/nocodb/src/models/PresignedUrl.ts
@@ -94,7 +94,7 @@ export default class PresignedUrl {
     },
     ncMeta = Noco.ncMeta,
   ) {
-    const isUrl = param.pathOrUrl.startsWith('http');
+    const isUrl = /^https?:\/\//i.test(param.pathOrUrl);
 
     let path = isUrl
       ? decodeURI(new URL(param.pathOrUrl).pathname)
--- a/packages/nocodb/src/models/PresignedUrl.ts
+++ b/packages/nocodb/src/models/PresignedUrl.ts
@@ -88,13 +88,17 @@ export default class PresignedUrl {
 
   public static async getSignedUrl(
     param: {
-      path: string;
+      pathOrUrl: string;
       expireSeconds?: number;
       filename?: string;
     },
     ncMeta = Noco.ncMeta,
   ) {
-    let path = param.path.replace(/^\/+/, '');
+    const isUrl = param.pathOrUrl.startsWith('http');
+
+    let path = isUrl
+      ? decodeURI(new URL(param.pathOrUrl).pathname)
+      : param.pathOrUrl.replace(/^\/+/, '');
 
     const { expireSeconds = DEFAULT_EXPIRE_SECONDS, filename } = param;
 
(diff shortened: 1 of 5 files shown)


packages/nocodb/src/models/PresignedUrl.ts
```diff
@@ -88,13 +88,17 @@ export default class PresignedUrl {
...
     param: {
-      path: string;
+      pathOrUrl: string;
       expireSeconds?: number;
       filename?: string;
     },
     ncMeta = Noco.ncMeta,
   ) {
-    let path = param.path.replace(/^\/+/, '');
+    const isUrl = param.pathOrUrl.startsWith('http');
```
salim-b, line 97: @mertmit Shouldn't this be a bit more precise and check for the whole *scheme* case-insensitively, i.e. using a regex like

```js
const isUrl = /^https?:\/\//i.test(param.pathOrUrl);
```

mertmit: Thanks for suggestion I've applied it

pranavxc (approved):

PR #9079, opened 2024-07-26
fix: disable adding columns to all view on creation

DarkPhoenix2704:
## Change Summary

When a new column is added, it automatically gets added to all the views and enabled. 
This disables this. The column will be shown only for the view from where it is created, others has to be manually enabled from fieldList

## Change type


dstala:
new fields will be added to other views unless the view is shared. by design

Issue #9087, opened 2024-07-26
🐛 Bug: Thumbnail generation failing


DarkPhoenix2704:
### Please confirm that the bug report does ***not*** already exist

- [X] I confirm there is no existing issue for this bug.

### Steps to reproduce

> Original URL (valid): https://rdb-attachments.s3.eu-central-003.backblazeb2.com/nc/uploads/noco/p75yos47mu8fs8b/mx5tkjg13vlbjr3/ccpxnishddejjt2/wallhaven-lmgyxl_j7Sl6.jpg

> Thumbnail URL (invalid): https://rdb-attachments.s3.eu-central-003.backblazeb2.com/nc/thumbnails/noco/p75yos47mu8fs8b/mx5tkjg13vlbjr3/ccpxnishddejjt2/wallhaven-lmgyxl_j7Sl6.jpg/tiny.jpg

> I just checked all files in the corresponding `rdb-attachments` bucket, there are *no* thumbnails, i.e. they were never written to the object store.

### Desired Behavior

```
2024-07-26T15:08:49Z app[1781991a690708] cdg [info][Nest] 335  - 07/26/2024, 3:08:49 PM   ERROR [ThumbnailGeneratorProcessor] Failed to generate thumbnails
2024-07-26T15:08:49Z app[1781991a690708] cdg [info][Nest] 335  - 07/26/2024, 3:08:49 PM   ERROR [ThumbnailGeneratorProcessor] TypeError: Invalid URL
2024-07-26T15:08:49Z app[1781991a690708] cdg [info][Nest] 335  - 07/26/2024, 3:08:49 PM   DEBUG [GlobalExceptionFilter] Cannot GET /dashboard/dltemp/15v8By6F80pge0Yp/1722013800000/nc/uploads/noco/p75yos47mu8fs8b/mx5tkjg13vlbjr3/ccpxnishddejjt2/wallhaven-73zg19_KQzZs.png
2024-07-26T15:08:49Z app[1781991a690708] cdg [info][Nest] 335  - 07/26/2024, 3:08:49 PM   DEBUG [GlobalExceptionFilter] NotFoundException: Cannot GET /dashboard/dltemp/15v8By6F80pge0Yp/1722013800000/nc/uploads/noco/p75yos47mu8fs8b/mx5tkjg13vlbjr3/ccpxnishddejjt2/wallhaven-73zg19_KQzZs.png
[...]

PR #9101, opened 2024-07-29
fix: thumbnail errors & path

mertmit:
## Change Summary

- call generate thumbnail job only for image attachments (instead of filtering inside the job, this way we will get better statistics & less number of jobs)
- thumbnail errors & path
- temporarily use arraybuffer for S3 as well to calculate image dimensions

## Change type


DarkPhoenix2704 (approved):

salim-b:
Thumbnail generation still fails, see https://github.com/nocodb/nocodb/issues/9087#issuecomment-2259201650

PR #9137, opened 2024-08-02
fix: s3 add forcePathStyle config

DarkPhoenix2704:
## Change Summary

- Allows Minio to be configured as S3 Plugin from environmental variable

## Change type

### DIFF
--- a/packages/nocodb/src/helpers/NcPluginMgrv2.ts
+++ b/packages/nocodb/src/helpers/NcPluginMgrv2.ts
@@ -137,6 +137,7 @@ class NcPluginMgrv2 {
           endpoint: process.env.NC_S3_ENDPOINT,
           access_key: process.env.NC_S3_ACCESS_KEY,
           access_secret: process.env.NC_S3_ACCESS_SECRET,
+          force_path_style: process.env.NC_S3_FORCE_PATH_STYLE === 'true',
         }),
       });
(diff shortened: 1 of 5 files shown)


salim-b:
@DarkPhoenix2704  I think the newly introduced `NC_S3_PATH_STYLE` env var should be [documented](https://docs.nocodb.com/getting-started/self-hosted/environment-variables/) (in [`packages/noco-docs/docs/020.getting-started/050.self-hosted/020.environment-variables.md`](https://github.com/nocodb/nocodb/blob/develop/packages/noco-docs/docs/020.getting-started/050.self-hosted/020.environment-variables.md)).

The difference between "path-style" and "virtual-hosted–style" S3 requests is explained [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#path-style-access).

I'd propose to document `NC_S3_PATH_STYLE` as follows:

| Variable | Mandatory | Description | If Not Set |
| -------- | --------- | ----------- | ---------- |
| `NC_S3_FORCE_PATH_STYLE` | No | Whether to force [path-style requests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#path-style-access) for the S3 storage plugin. | |

dstala (approved):

Issue #9446, opened 2024-09-07
🐛 Bug: external mysql database the sum withing grouping stops being counted


dstala:
### Please confirm that the bug report does ***not*** already exist

- [X] I confirm there is no existing issue for this bug.

### Steps to reproduce

But the total sum of all groups is counted

-- as reported over discord. reproducible.

### Desired Behavior

group aggregators should work 

### Project Details

255.2

PR #9462, opened 2024-09-10
fix: aggregation not working in mariadb

DarkPhoenix2704:
## Change Summary

Closes https://github.com/nocodb/nocodb/issues/9446

## Change type

### DIFF
--- a/packages/nocodb/src/db/BaseModelSqlv2.ts
+++ b/packages/nocodb/src/db/BaseModelSqlv2.ts
@@ -1623,13 +1623,13 @@ class BaseModelSqlv2 {
 
         if (this.dbDriver.client.config.client === 'mysql2') {
           selectors.push(
-            this.dbDriver.raw(`JSON_UNQUOTE((??)) as ??`, [
+            this.dbDriver.raw('JSON_UNQUOTE(??) as ??', [
               jsonBuildObject,
               `${f.alias}`,
             ]),
           );
         } else {
-          selectors.push(this.dbDriver.raw(`(??) as ??`, [tQb, `${f.alias}`]));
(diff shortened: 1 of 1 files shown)


packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -1621,7 +1621,16 @@ class BaseModelSqlv2 {
 
         tQb.select(jsonBuildObject);
 
-        selectors.push(this.dbDriver.raw(`(??) as ??`, [tQb, `${f.alias}`]));
+        if (this.dbDriver.client.config.client === 'mysql2') {
+          selectors.push(
+            this.dbDriver.raw(`JSON_UNQUOTE((??)) as ??`, [
```
pranavxc, line 1626: Does 2 brackets required here  ? `((??))`  => `(??)`

And string literal not required as well since no js variables used.

packages/nocodb/src/db/BaseModelSqlv2.ts
```diff
@@ -1621,7 +1621,16 @@ class BaseModelSqlv2 {
...
+              jsonBuildObject,
+              `${f.alias}`,
```
pranavxc, line 1628: `'${f.alias}' `=> `f.alias`

pranavxc (approved):

packages/nocodb/src/db/BaseModelSqlv2.ts
DarkPhoenix2704: If the f.alias is number or some other special character, it may cause issue. That's why kept in ``

Issue #9450, opened 2024-09-09
🐛 Bug: migration that generates thumbnails fails with external storage providers


dolsem:
### Please confirm that the bug report does ***not*** already exist

- [X] I confirm there is no existing issue for this bug.

### Steps to reproduce

This line https://github.com/nocodb/nocodb/blob/3d3f97800bf6cd25bd768ca2223dffe4e2d782f5/packages/nocodb/src/modules/jobs/migration-jobs/nc_job_002_thumbnail.ts#L67C32-L67C80 makes the migration fail when the storage provider is not local, e.g. minio or S3, as `scanFiles()` returns undefined.

### Desired Behavior

Ideally support external providers - probably dont have to rely on globs if those are not supported.

### Project Details

Node: **v20.15.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.255.2**

PR #9463, opened 2024-09-10
feat: Implement scanFiles for gcs and Minio

DarkPhoenix2704:
## Change Summary

Implement scanFiles for gcs and Minio

Closes https://github.com/nocodb/nocodb/issues/9450

## Change type

### DIFF
--- a/packages/nocodb/src/plugins/gcs/Gcs.ts
+++ b/packages/nocodb/src/plugins/gcs/Gcs.ts
@@ -1,18 +1,18 @@
 import fs from 'fs';
 import { promisify } from 'util';
+import { Readable } from 'stream';
 import { Storage } from '@google-cloud/storage';
 import axios from 'axios';
 import { useAgent } from 'request-filtering-agent';
 import type { GetSignedUrlConfig, StorageOptions } from '@google-cloud/storage';
 import type { IStorageAdapterV2, XcFile } from '~/types/nc-plugin';
-import type { Readable } from 'stream';
 import { generateTempFilePath, waitForStreamClose } from '~/utils/pluginUtils';
 
 interface GoogleCloudStorageInput {
   client_email: string;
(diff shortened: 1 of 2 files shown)


packages/nocodb/src/plugins/gcs/Gcs.ts
```diff
@@ -176,13 +177,27 @@ export default class Gcs implements IStorageAdapterV2 {
...
-    return Promise.resolve(undefined);
+  public async scanFiles(globPattern: string): Promise<Readable> {
+    const stream = new Readable({
+      objectMode: true,
+      read() {},
+    });
+
+    const [files] = await this.storageClient.bucket(this.bucketName).getFiles({
+      prefix: globPattern.replace(/^\*/, ''),
+    });
```
mertmit, line 194: Would this work for huge number of files? Even so it is not a good idea to load all into memory at once.
Can we use getFilesStream instead?

mertmit (changes requested):

packages/nocodb/src/plugins/gcs/Gcs.ts
```diff
@@ -176,13 +177,27 @@ export default class Gcs implements IStorageAdapterV2 {
...
+
+    for (const file of files) {
+      stream.push(file.name);
+    }
+    stream.push(null);
```
dolsem, line 199: To address @mertmit's comment:
```suggestion
    this.storageClient.bucket(this.bucketName).getFilesStream({
      prefix: globPattern.replace(/^\*/, ''),
      autoPaginate: true,
    })
      .on('error', (err) => {
        stream.destroy(err);
      })
      .on('data', (file) => {
        stream.push(file.name);
      })
      .on('end', () => {
        stream.push(null);
      });
```

mertmit (approved):

PR #9509, opened 2024-09-18
feat: restrict concurrent operations

mertmit:
## Change Summary

- Restrict concurrent operations to 5 by default (we allow 10 meta connections so lets avoid occupying all of them)

## Change type


mytoy45:
Ok

PR #9537, opened 2024-09-23
feat: add sqlite support in the upstall script

sinanmohd:
## Change Summary

- closes nocodb/nocohub#3805

## Change type

- [x] feat: add sqlite support in the upstall script

### DIFF
--- a/docker-compose/1_Auto_Upstall/noco.sh
+++ b/docker-compose/1_Auto_Upstall/noco.sh
@@ -380,6 +380,25 @@ check_if_docker_is_running() {
 	fi
 }
 
+persistent_store_isdeleted() {
+	_persistent_store="$1"
+
+	if [ ! -d "$_persistent_store" ]; then
+		print_warning "Persistent store was deleted without stopping the containers"
+
+		for container in $($CONFIG_DOCKER_COMMAND ps | grep -Eo 'nocodb-[a-z]+-[0-9]+$'); do
+			if ! $CONFIG_DOCKER_COMMAND stop "$container" > /dev/null 2>&1; then
+				print_error "Failed to stop ${container}"
+				exit 1
+			fi
+		done
+
+		return 0
+	fi
+
+	return 1
+}
+
 # Main functions
 check_existing_installation() {
 	NOCO_FOUND=false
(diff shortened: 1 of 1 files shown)


docker-compose/1_Auto_Upstall/noco.sh
```diff
@@ -510,6 +514,9 @@ create_docker_compose_file() {
     if [ "${CONFIG_MINIO_ENABLED}" = "Y" ]; then
         gen_minio=1
     fi
+    if [ "${CONFIG_MINIO_ENABLED}" = "Y" ]; then
```
pranavxc, line 517: Should be `["${CONFIG_POSTGRE_SQLITE}" = "p"] || ["${CONFIG_POSTGRE_SQLITE}" = "P" ]`

docker-compose/1_Auto_Upstall/noco.sh
```diff
@@ -551,7 +566,10 @@ EOF
   cat >> "$compose_file" <<EOF
     networks:
       - nocodb-network
+EOF
 
+  if [ "$CONFIG_POSTGRE_SQLITE" = "P" ]; then
```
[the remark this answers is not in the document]
sinanmohd: this should be set a unique value by prompt  ?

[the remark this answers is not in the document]
pranavxc: Ya if you can handle at prompt level then it's great.

docker-compose/1_Auto_Upstall/noco.sh
```diff
@@ -172,6 +173,41 @@ prompt() {
...
+        for one in "$@"; do
+            resp_upper="${response^^}"
+
+            if [ "$response" = "$one" ]; then
+                echo "$one"
+                return
+            elif [ "$resp_upper" = "$one" ]; then
+                echo "$one"
+                return
+            fi
```
pranavxc, line 205: we can convert both values to upper case and then compare, this way we can avoid this duplicate code

```bash
            resp_upper="${response^^}"
            one_upper="${one^^}"

            if [ "$resp_upper" = "$one_upper" ]; then
                echo "$one"
                return
            fi
```

sinanmohd: thanks

pranavxc (changes requested):
Not part of this pr, let's add `-v` while removing docker containers to remove any attached volume


```
$CONFIG_DOCKER_COMMAND compose down
```
to

```
$CONFIG_DOCKER_COMMAND compose down -v
```

pranavxc (approved):
Code changes looks fine, will merge after couple more testing.

pranavxc (commented):
Getting following error while running script, initially it was working and later it started throwing following error.

Screenshot 2024-09-28 at 11 35 31 AM

```bash
bash <(curl -sSL https://raw.githubusercontent.com/nocodb/nocodb/ec578a824a38e64654346bc42d6ffee5f74c5f8f/docker-compose/1_Auto_Upstall/noco.sh) <(mktemp)
```

sinanmohd:
> Getting following error while running script, initially it was working and later it started throwing following error.
> Screenshot 2024-09-28 at 11 35 31 AM
> 
> ```shell
> bash <(curl -sSL https://raw.githubusercontent.com/nocodb/nocodb/ec578a824a38e64654346bc42d6ffee5f74c5f8f/docker-compose/1_Auto_Upstall/noco.sh) <(mktemp)
> ```

this happens when 
1. the container is running (the following outputs a path)
	CUSTOM_HOME="$($CONFIG_DOCKER_COMMAND inspect --format='{{index .Mounts 0}}' "$NOCO_ID" | cut -d ' ' -f 3)"
	PARENT_DIR="$(dirname "$CUSTOM_HOME")"

2. you are in $(dirname $NOCO_HOME)
3. NOCO_HOME is deleted

did you remove NOCO_HOME while the containers was still up ?

docker-compose/1_Auto_Upstall/noco.sh
```diff
@@ -45,486 +46,538 @@ print_warning() { print_color "$YELLOW" "WARNING: $1"; }
...
+
+	while true; do
+		read -r -p "$prompt_text" response
+		if [ -z "$response" ]; then
+			echo "$default_response"
+			return
+		fi
+
+		for one in "$@"; do
+			resp_upper="${response^^}"
```
pranavxc, line 199: Looks like this syntax is supported widely

Screenshot 2024-09-30 at 12 27 03 PM


so let's replace it with something like following which

```bash
response_upper=$(echo "$response" | tr '[:lower:]' '[:upper:]')
```

docker-compose/1_Auto_Upstall/noco.sh
```diff
@@ -45,486 +46,538 @@ print_warning() { print_color "$YELLOW" "WARNING: $1"; }
...
+			one_upper="${one^^}"
```
pranavxc, line 200: check previous comment

pranavxc (changes requested):

pranavxc (approved):

Issue #8958, opened 2024-07-07
🐛 Bug: No belongs-to column for link to same table


sebastianoe:
### Please confirm if bug report does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Steps to reproduce ?

Use case: create an inventory with boxes where a box might be placed in another.

1. Create a table `Box` which has a `has-many` column `Contained Boxes` that refers to the `Box` table.
2. See that there is no generated `belongs-to` column visible, as it would be if the `has-many` link referred to another table.

This makes it easy to see which boxes a box contains but makes it hard to see to which other box a specific box belongs to.
A `belongs-to` field is visible in the schema and a system column is apparently generated - but this one is neither visible by default nor can it be renamed nor is it appropriately formatted, as it only refers to the ID.

See the attached screenshots.

### Desired Behavior

Render an appropriate `belongs-to` column in the table so that it's easy to see to which other box a specific box belongs to.

### Project Details

Node: **v20.12.1**
Arch: **x64**
Platform: **linux**
Docker: **false**
RootDB: **mysql2**
PackageVersion: **0.251.0**

### Attachments

![image](https://github.com/nocodb/nocodb/assets/962041/61b9e2e0-457a-4ae5-b101-637d40e03219)

![image](https://github.com/nocodb/nocodb/assets/962041/2056272a-2a16-42bc-9096-f80776ed6bc4)

![image](https://github.com/nocodb/nocodb/assets/962041/3f6e6227-45fd-4b26-b7c9-5ac57a977d41)

![image](https://github.com/nocodb/nocodb/assets/962041/7d52dcec-644e-47c0-a76b-28694e9766d7)

dstala:
@sebastianoe Discussed with the team. We will fix this in our next release.

lusimeon:
Hi @dstala, this issue does not seems to be fixed in latest release (0.257.2). The `belongs-to` side of the relation appears in schema but not in views or fields list. Is it possible to reopen?

dstala:
BT field is hidden by default on self-has-many-link creation.
From the "Fields" menu in the toolbar, use `Show system fields` & enable BT field as required.

dstala:
You can close this thread if that helps. @lusimeon

lusimeon:
Ok, that helps, thanks for the advice. I can't close this thread because i'm not the author. @dstala

PR #9700, opened 2024-10-22
fix: show self referencing belongs to

lusimeon:
closes: #8958

## Change Summary

Change visibility of self referencing belongs to in ui, see #8958.

## Change type

## Test/ Verification

1. Create a self referencing "Links" column (with "Has Many" relation type).
2. Two columns should be created and displayed on ui, one for the `has-many` side, another for the `belongs-to` side of the relation.

## Additional information

According to previously implemented code comment, self referencing belongs to were purposely hidden from ui, but i don't figure out why, any idea?


lusimeon:
Closed due to https://github.com/nocodb/nocodb/issues/8958#issuecomment-2443182505

PR #9724, opened 2024-10-27
Fix the Docker with SQLite run command

lbm98:
## Change Summary

Small fix to the Docker with SQLite run command


salim-b:
This PR is obsolete meanwhile and can be closed, I think.

PR #9864, opened 2024-11-20
chore: sync various

mertmit:
## Change Summary

## Change type


dstala (approved):

packages/nocodb-sdk/src/lib/globals.ts
```diff
@@ -9,6 +9,37 @@ export enum ViewTypes {
...
+};
+
+export const stringToViewTypeMap: Record<string, ViewTypes> = Object.entries(
+  viewTypeStrings
+).reduce((acc, [key, value]) => {
+  acc[value] = Number(key);
+  return acc;
+}, {});
+
+export const ViewTypesNameMap = {
```
rameshmane7218, line 34: `viewTypeAlias` as already present in `nc-gui/**/viewUtils.ts`, also the alias map is created in `command-palette.service.ts` both ee and oss file

We have to keep sdk changes as it is and remove from other places

rameshmane7218 (approved):

DarkPhoenix2704 (approved):

PR #9932, opened 2024-12-02
fix(extensions): typings for KvStore

amandesai01:
## Change Summary

Add typing for extension's KV Store

## Change type
chore: (updating grunt tasks etc; no production code change)

### DIFF
--- a/packages/nc-gui/composables/useExtensions.ts
+++ b/packages/nc-gui/composables/useExtensions.ts
@@ -35,6 +35,13 @@ export interface ExtensionManifest {
   }
 }
 
+export interface IKvStore<T> {
+  get(key: string): T | null
+  set(key: string, value: T): Promise<void>
+  delete(key: string): Promise<void>
+  serialize(): Record<string, T>
+}
+
 abstract class ExtensionType {
   abstract id: string
   abstract uiKey: number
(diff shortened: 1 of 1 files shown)


packages/nc-gui/composables/useExtensions.ts
```diff
@@ -35,14 +35,21 @@ export interface ExtensionManifest {
   }
 }
 
+export interface IKvStore<T> {
+  get(key: string): T | null
+  set(key: string, value: T): Promise<void>
+  delete(key: string): Promise<void>
+  serialize(): Record<string, T>
+}
```
mertmit, line 43: Something like following might allow us to store different keys in different types which would be more convenient imo
```
export interface IKvStore<T extends Record<string, any>> {
  get<K extends keyof T>(key: K): T[K] | null;
  set<K extends keyof T>(key: K, value: T[K]): Promise<void>;
  delete<K extends keyof T>(key: K): Promise<void>;
  serialize(): Record<string, T[keyof T]>;
}
```

packages/nc-gui/composables/useExtensions.ts
```diff
@@ -273,9 +280,9 @@ export const useExtensions = createSharedComposable(() => {
       return updateExtension(this._id, { kv_store: this.data })
     }
 
-    delete(key: string) {
+    async delete(key: string) {
       delete this.data[key]
-      return updateExtension(this._id, { kv_store: this.data })
+      await updateExtension(this._id, { kv_store: this.data })
```
mertmit, line 285: it already returns a promise is this required ?

mertmit (approved):

PR #9977, opened 2024-12-06
fix: Prevent sum from failing if field is blank

### LINKED ISSUE #9876: 🐛 Bug: SUM fails if field is blank

### Steps to reproduce

1. Create records with number fields
2. Leave some of those fields blank on a record
3. Create a formula field which does a SUM or other math operation, it will fail and produce no output because the empty fields break the formula.

### Desired Behavior

 It should assume that a blank field equals zero. Or let the user choose some other default in the field properties window. Our current solution is to make a seperate field which checks if the other field is empty and if so, output zero. This is obviously annoying to do and less efficient from a database perspective.

### Project Details

Hosted on nocodb.com

### Attachments

_No response_

Bchenjamin:
## Change Summary

Resolves #9876

The following changes are in `formulaQueryBuilderv2.ts`:
- SUM works by recursively calling an evaluation function over the relevant columns to accumulate a final result
- In the case where there is more than one argument left, the argument in question (left) is defaulted to 0 using`COALESCE`
- In the base case (there is exactly one argument left), the argument is defaulted to 0 using `COALESCE`

## Change type

## Test/ Verification

Manual UI testing of the SUM forumla

I have attached screenshots of testing on the UI

Screenshot 2024-12-05 at 7 48 22 PM

Screenshot 2024-12-05 at 7 49 11 PM

### DIFF
--- a/packages/nocodb/src/db/formulav2/formulaQueryBuilderv2.ts
+++ b/packages/nocodb/src/db/formulav2/formulaQueryBuilderv2.ts
@@ -868,14 +868,26 @@ async function _formulaQueryBuilder(params: {
               {
                 type: 'BinaryExpression',
                 operator: '+',
-                left: pt.arguments[0],
+                left: {
+                  type: 'CallExpression',
+                  callee: { type: 'Identifier', name: 'COALESCE' },
+                  arguments: [ pt.arguments[0], { type: 'Literal', value: 0 } ],
+                },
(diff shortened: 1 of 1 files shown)


Bchenjamin:
@DarkPhoenix2704 Anyway you can review this and help me get it merged soon?

DarkPhoenix2704:
@pranavxc has much more context with formula

pranavxc (approved):

PR #10006, opened 2024-12-09
chore: confirm & fix artifact issue

mertmit:
Re #10005


Juneezee:
I can confirm that the artifacts were downloaded correctly: https://github.com/nocodb/nocodb/actions/runs/12240111307/job/34142157122

dstala (approved):

PR #10102, opened 2024-12-21
fix: order migration recovery

DarkPhoenix2704:
## Change Summary

## Change type


packages/nocodb/src/modules/jobs/migration-jobs/nc_job_007_recover_order_column.ts
```diff
@@ -0,0 +1,273 @@
...
+            .catch((e) => {
+              this.log(`Error processing model ${model.id}:`, e.message);
+            });
+        }
+      }
+
+      await queue.onIdle();
+
+      await ncMeta.disableUpgraderMode();
+
```
mertmit, line 181: lets have this in catch block as well (we can do same in 005 migration bc we disable cache on enabling & we need to disable this mode whether the result)

mertmit (approved):

PR #10118, opened 2024-12-25
Update ar.json

xximj:
Updated and reviewed Arabic translations for Nocodb.


dstala:
Please help us by applying these changes in crowdin
https://docs.nocodb.com/engineering/translation

xximj:
I tried using crowdin but it's not accepting the major file changes I made. Could you please replace the existing file in crowdin with my version? 

While my version isn't perfect, but it's a better base for Arabic translation than the current machine-translated one and we can build upon it.
[ar.json](https://github.com/user-attachments/files/18260885/ar.json)

Issue #10132, opened 2024-12-27
🐛 Bug: Error in Grid when pageInfo is undefined in list api


DarkPhoenix2704:
### Please confirm that the bug report does ***not*** already exist

- [X] I confirm there is no existing issue for this bug.

### Steps to reproduce

> For some reason, starting last week, one of the views broke and says No Records Found. All other views are working, only this one broke. What can be the cause or what should I look into? I didn't change anything critical since other views still work

![image](https://github.com/user-attachments/assets/f9cce503-6647-4cad-a33f-11831f187826)

![image](https://github.com/user-attachments/assets/a54f923d-e94a-4678-8007-0939c35b24d5)

Ref: https://discord.com/channels/661905455894888490/661905455894888493/1320823418463064135

### Desired Behavior

. 

### Project Details

Cloud

PR #10133, opened 2024-12-27
fix: Error in Grid when pageInfo is undefined in list api

DarkPhoenix2704:
## Change Summary

Closes https://github.com/nocodb/nocodb/issues/10132

## Change type

### DIFF
--- a/packages/nc-gui/composables/useInfiniteData.ts
+++ b/packages/nc-gui/composables/useInfiniteData.ts
@@ -237,7 +237,7 @@ export function useInfiniteData(args: {
 
       const data = formatData(response.list, response.pageInfo, params)
 
-      if (response.pageInfo?.totalRows) {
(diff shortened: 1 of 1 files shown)


packages/nc-gui/composables/useInfiniteData.ts
```diff
@@ -222,9 +235,9 @@ export function useInfiniteData(args: {
             },
           )
 
-      const data = formatData(response.list, response.pageInfo)
+      const data = formatData(response.list, response.pageInfo, params)
 
-      if (response.pageInfo.totalRows) {
+      if (response.pageInfo?.totalRows) {
```
pranavxc, line 240: What if totalRows is `0` ? Do we need to reset ?

DarkPhoenix2704: fixed

pranavxc (approved):

PR #10263, opened 2025-01-13
feat: better json editing ux 

AmitJoki:
## Change Summary

Make the JSON always editable in the modal. No inline editing of JSON.

1. In Grid View, you can trigger the modal by
    a. Typing on it.
    b. Pressing Enter
    c. Double clicking the cell.
    d. Pressing the expand button.
2. In other views
    a. Pressing the expand button
    b. Single click on the field.

The UI buttons (Cancel and Save and the expand/contract) buttons are unified like the rest of the app so they don't look out of place.

The JSON editor modal is resizable horizontally.

There was also a bug fix where the json editor wasn't focused upon expansion.

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/components/cell/Json.vue
+++ b/packages/nc-gui/components/cell/Json.vue
@@ -147,7 +147,12 @@ useSelectedCellKeyupListener(active, (e) => {
       // non-printing keys, ignore them
       break
     default:
-      // Otherwise it's a printing character, open the JSON modal for editing
+      // Otherwise it's a printing character, append it and open the JSON modal for editing
+      if (typeof localValue.value === 'string') {
+        localValue.value += e.key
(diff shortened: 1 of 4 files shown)


dstala (approved):

dstala:
- [ ] when row height is more, see if we can adapt and show more information?
Screenshot 2025-01-13 at 9 41 45 PM

AmitJoki:
@dstala More data will be shown based on the row height now fixing 1). It also fixes 2) Spacebar should also now always open up the expanded view.

mertmit (commented):
When opening the JSON editor by typing something we should append the character at the end of the string not start.
Also cursor should go to end.

See LongText behavior for reference.

PR #10267, opened 2025-01-13
fix: avoid create db call on connect existing external

### LINKED ISSUE #9990: 🐛 Bug: adding a new mysql data source requires db user with `create database` permission

### Steps to reproduce

1. create a mysql connection/integration using a db user with no `create database` permission
2. try to create a new data source using this connection to connection to an *existing* database
3. you will get the following error:
```sql
create database if not exists `blah_blah_db` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci - Access denied for user 'user_no_db_create_access'@'%' to database 'blah_blah_db'

```

### Desired Behavior

we should be able to use a user with no `create database` permission to access an existing database in mysql.
more specifically, I think this sql command `create database if not exists` should not be used. instead we should check for db existence with a different command first and if it does not exist then call `create database`.

### Project Details

Node: **v20.15.1**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.258.3**

### Attachments

_No response_

mertmit:
## Change Summary

Fixes #9990

## Change type

### DIFF
--- a/packages/nocodb/src/db/sql-client/lib/pg/PgClient.ts
+++ b/packages/nocodb/src/db/sql-client/lib/pg/PgClient.ts
@@ -672,7 +672,7 @@ class PGClient extends KnexClient {
     try {
       const { rows } = await this.sqlClient.raw(
         `SELECT datname as database FROM pg_database WHERE datistemplate = false and datname = ?`,
-        [args.database],
+        [args.databaseName],
       );
       result.data.value = rows.length > 0;
     } catch (e) {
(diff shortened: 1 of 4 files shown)


packages/nocodb/src/db/sql-client/lib/pg/PgClient.ts
```diff
@@ -672,9 +672,18 @@ class PGClient extends KnexClient {
...
       );
+
       result.data.value = rows.length > 0;
+
+      if (result.data.value && args.schema) {
+        const { rows: rows2 } = await this.sqlClient.raw(
+          `SELECT schema_name FROM information_schema.schemata WHERE schema_name = ?`,
+          [args.schema],
+        );
+        result.data.value = rows.length > 0 && rows2.length > 0;
```
pranavxc, line 685: `rows.length > 0 && rows2.length > 0` => `rows2.length > 0`

pranavxc (approved):

PR #10350, opened 2025-01-27
chore(github/workflow/bumper): init 

sinanmohd:
## Change Summary

avoid nix hash missmatch issues when pnpm-lock is updated

this pr allows us to confidently run `nix run github:nocodb/nocodb/<branch-name>`
to locally test any pr/branch at any time with a single command 

## Change type

## Test/ Verification

tested with nix run

### DIFF
--- a/.github/workflows/nix-bumper.yml
+++ b/.github/workflows/nix-bumper.yml
@@ -1,0 +1,26 @@
+name: "bumper"
+
+on:
+  push:
+    branches:
+      - "devlop"
+  pull_request_target:
+    branches:
+      - "devlop"
+
+jobs:
+  bumper:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: "write"
+      id-token: "write"
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Install Nix
+      uses: DeterminateSystems/nix-installer-action@main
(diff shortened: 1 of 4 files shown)


.github/workflows/nix-bumper.yml
```diff
@@ -0,0 +1,26 @@
+name: "bumper"
+
+on:
+  push:
+    branches:
+      - "devlop"
+  pull_request_target:
+    branches:
+      - "devlop"
```
mertmit, line 9: Is "devlop" intended or a typo?

.github/workflows/nix-bumper.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  bumper:
+    runs-on: ubuntu-latest
```
mertmit, line 13: Is it possible we make this run on self-hosted runners instead?
If a dependency is required we can sync & install

nix/bumper/bumper.sh
mertmit, line 1: I noticed few typos on messages, please fix them to keep things clean

.github/workflows/nix-bumper.yml
sinanmohd: no worries, nix takes care of all the deps
I'll move it to the self hosted runner

mertmit (approved):

PR #10443, opened 2025-02-07
fix: reject bad attachment payloads

mertmit:
## Change Summary

As in title

## Change type


mertmit:
included https://github.com/nocodb/nocodb/pull/10444

Issue #10513, opened 2025-02-15
🐛 Bug: Alt + Left Arrow Shortcut Causes Data Loss


jereze:
### Please confirm that the bug report does ***not*** already exist

- [x] I confirm there is no existing issue for this bug.

### Steps to reproduce

I am using Nocodb 0.260.2 on Firefox (macOS). While editing a record, pressing `Alt + Left Arrow` unexpectedly triggers a shortcut to open the previous record, causing any unsaved modifications to be lost.

This behavior is inconsistent with standard text editing across most applications, where `Alt + Left Arrow` moves the cursor to the previous word within the text field. Since I frequently use this shortcut, I have unintentionally lost significant changes multiple times.

I believe this is a bug. I have attached a video to illustrate the issue.

### Desired Behavior

When editing a record, pressing `Alt + Left Arrow` should move the cursor to the previous word within the text field, as expected in most text editors and web applications. It should not trigger navigation between records or cause data loss.

### Project Details

```
Node: **v22.13.0**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.260.2**
```

### Attachments

https://github.com/user-attachments/assets/285e899e-b168-40ce-b85e-3ad510d5899a

NosimusAI:
Fixed by Nosimus AI in this PR: https://github.com/nocodb/nocodb/pull/10514

dstala:
@NosimusAI Please sign CLA [here](https://github.com/nocodb/nocodb/pull/10514#issuecomment-2661087084)

jereze:
Thank you!

PR #10514, opened 2025-02-15
fix: Alt + Left Arrow Shortcut Causes Data Loss

NosimusAI:
Fix of https://github.com/nocodb/nocodb/issues/10513

## Change Summary

Target Check: The added condition skips grid keyboard handling if the event occurs in an input/textarea.

Preserves Default Behavior: This allows native text navigation shortcuts (like Alt + Arrow) to work as expected in form fields.

Non-Intrusive: Maintains existing grid navigation behavior when not editing text cells.

## Change type

### DIFF
--- a/packages/nc-gui/composables/useSelectedCellKeyupListener/index.ts
+++ b/packages/nc-gui/composables/useSelectedCellKeyupListener/index.ts
@@ -7,6 +7,13 @@ function useSelectedCellKeydownListener(
   { immediate = false, isGridCell = true }: { immediate?: boolean; isGridCell?: boolean } = {},
 ) {
   const finalHandler = (e: KeyboardEvent) => {
+    /**
+     * Add check for input/textarea elements
+     */
+    if (e.target instanceof HTMLInputElement || e.target instanceof HTMLTextAreaElement) {
+      return;
+    }
+
     if (cmdKActive()) return
 
     /**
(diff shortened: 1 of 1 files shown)


dstala:
Please sign CLA [here](https://github.com/nocodb/nocodb/pull/10514#issuecomment-2661087084)

packages/nc-gui/composables/useSelectedCellKeyupListener/index.ts
```diff
@@ -7,6 +7,17 @@ function useSelectedCellKeydownListener(
   { immediate = false, isGridCell = true }: { immediate?: boolean; isGridCell?: boolean } = {},
 ) {
   const finalHandler = (e: KeyboardEvent) => {
+    /**
```
rameshmane7218, line 10: @NosimusAI this `useSelectedCellKeydownListener` is used in multiple places and we cannot prevet all events as the issue is only related to expanded form keyboard navigation left and right 
ref: https://github.com/NosimusAI/nocodb/blob/a82578e8bb0dac1af9e9ba986c47157644bd127b/packages/nc-gui/components/smartsheet/expanded-form/index.vue#L441

you have to add this condition their not here

rameshmane7218 (changes requested):

rameshmane7218:
@NosimusAI closing this PR as we wanted to do release and I added fix in another PR #10551

PR #10651, opened 2025-02-26
fix: signed url not working for s3 based providers

DarkPhoenix2704:
## Change Summary

Signed Urls were not working for profilePics, workspacePics and organizationPics, if forcePathStyle enabled for S3 plugin

## Change type

### DIFF
--- a/packages/nocodb/src/models/PresignedUrl.ts
+++ b/packages/nocodb/src/models/PresignedUrl.ts
@@ -317,6 +317,8 @@ export default class PresignedUrl {
       }
 
       await Promise.all(promises);
-    } catch {}
+    } catch (e) {
+      console.error(e);
+    }
   }
 }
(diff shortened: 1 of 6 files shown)


packages/nocodb/src/models/PresignedUrl.ts
```diff
@@ -317,6 +317,8 @@ export default class PresignedUrl {
       }
 
       await Promise.all(promises);
-    } catch {}
+    } catch (e) {
+      console.error(e);
+    }
```
mertmit, line 322: lets use logger here

packages/nocodb/src/plugins/backblaze/Backblaze.ts
```diff
@@ -42,10 +43,18 @@ export default class Backblaze extends GenericS3 implements IStorageAdapterV2 {
 
     this.s3Client = new S3Client(s3Options);
   }
+
+  private DEFAULT_KEYS = [
+    'uploads',
+    'thumbnails',
+    ...(Object.values(PublicAttachmentScope) as string[]),
+  ];
```
mertmit, line 51: We can move this to common place (eg. packages/nocodb/src/constants/index.ts)

mertmit (changes requested):

mertmit (approved):

PR #10672, opened 2025-02-28
feat(nix/package/workflows/aiopusher): init

sinanmohd:
## Change Summary

Build and release aio images when master branch moves.

## Change type

## Test/ Verification

nix run github:nocodb/nocodb/aio-workflow#workflows.aiopusher -L

new images @ https://hub.docker.com/r/nocodb/nocodb_aio/tags are built using this


.github/workflows/nix-aiopusher.yml
```diff
@@ -0,0 +1,18 @@
+name: aiopusher
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
```
mertmit, line 7: Instead of push we can make this part of our nocodb release flow so it will follow our release cycle.

mertmit:
Regarding our Nix setup (I fore-mentioned but wanted to raise this concern again)

Develop branch is our `unstable` branch.
So instead all documentation and commands should default to master (following our release cycle).

Building develop branch for production usage is not a good idea, if we would like to avoid having a branch prefix we can think about having nocodb-nix repo as well (for node usage we provide nocodb-seed)

This way we can make sure nix is deployments are stable as well.

sinanmohd:
> Develop branch is our `unstable` branch. So instead all documentation and commands should default to master (following our release cycle).

I've updated the recommend branch to master in the docs-refactor pr
https://github.com/nocodb/nocodb/pull/10682/commits/39593dc3a9b13d5b9321816ac11e9a95b7bae19f

mertmit (approved):

PR #10740, opened 2025-03-06
Fix row filter split error

ssweber:
## Change Summary

This PR fixes an issue where the "Row Filtered" message wasn't appearing for rows that should have been filtered out. The root cause was that the filter comparison logic was trying to call string methods (like .split()) on non-string values, particularly arrays of objects from relation fields.

Added a safeToString helper function that properly converts field values to strings before they're used in string operations, ensuring the row validation works correctly for all data types.

This prevents the "TypeError: data[field].split is not a function" error that was preventing the isValidationFailed flag from being set correctly

## Change type

### DIFF
--- a/packages/nc-gui/utils/dataUtils.ts
+++ b/packages/nc-gui/utils/dataUtils.ts
@@ -173,6 +173,22 @@ export function validateRowFilters(_filters: FilterType[], data: any, columns: C
     return true
   }
 
+function safeToString(value) {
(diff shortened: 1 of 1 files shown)


ssweber:
Of if a new function isn't desired, we could make the string operations safe without introducing a new function

ssweber:
Sorry. I have a much simpler fix with less changes.

PR #10743, opened 2025-03-06
fix(nc-gui): Handle array fields in filter comparisons

ssweber:
## Change Summary

Previously, filter comparisons (lallof, anyof, etc.) expected string values but would break when encountering fields that return arrays of objects. This PR adds inline handling to extract IDs from arrays when performing these comparisons.

Changes:
- Added array check and ID extraction for 'allof', 'anyof', 'nallof', and 'nanyof' cases
- Maintains existing behavior for non-array fields

Example:
Before: [{id: 1}, {id: 2}] would cause comparison errors After: [{id: 1}, {id: 2}] is properly handled as "1,2" for comparisons

## Change type

### DIFF
--- a/packages/nc-gui/utils/dataUtils.ts
+++ b/packages/nc-gui/utils/dataUtils.ts
@@ -162,6 +162,16 @@ export const rowDefaultData = (columns: ColumnType[] = []) => {
   return defaultData
 }
 
+// Helper function just for getting ID(s) from possible LTAR field
+function getFieldValue(value) {
+  if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


ssweber:
Sorry for noise, I will do more research before pulling next time. With more reading I realize that this case is already handled further up in the logic.

```
if ([UITypes.User, UITypes.CreatedBy, UITypes.LastModifiedBy].includes(column.uidt!)) {
          const userIds: string[] = Array.isArray(data[field])
            ? data[field].map((user) => user.id)
            : data[field]?.id
            ? 
```

I think we have to add add Lookup in this list?

Issue #10741, opened 2025-03-06
🐛 Bug: Error "FilterChildrenList" -  Editor users not allowed to use filter groups.


VanioRaiser:
### Please confirm that the bug report does ***not*** already exist

- [x] I confirm there is no existing issue for this bug.

### Steps to reproduce

Create filters on a table that involve at least one filter group
Create a user with editor permissions
Log in as editor user
Look at the table you made filters for in step 1
Click the Filter dropdown
Get error: "filterChildrenList - editor : Not allowed"; anything within the filter group is blank, though the filter still seems to be applied correctly.

### Desired Behavior

Editor users should be able to create and see filters that include filter groups

### Project Details

Node: **v22.14.0**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **pg**
PackageVersion: **0.262.1**

### Attachments

![Image](https://github.com/user-attachments/assets/39582614-3814-491d-b54b-a0b9f110d235)
![Image](https://github.com/user-attachments/assets/1b4576a3-50ab-435a-b4e4-d004f91c30ef)

VanioRaiser:
https://github.com/nocodb/nocodb/issues/6224

PR #10751, opened 2025-03-07
fix: invalid filter acl

DarkPhoenix2704:
## Change Summary

Closes https://github.com/nocodb/nocodb/issues/10741

## Change type


AMEESHAARACKAL (approved):
Bug Fixed : Error "FilterChildrenList" is not occuring

PR #10757, opened 2025-03-07
fix: Fix user field filter behavior for 'Lookup' type

### LINKED ISSUE #10728: 🐛 Bug: Live Row Filtering not working with Lookup Field as 2nd AND

### 1) Create Table
Table Project:

- Title
- Tasks (Link HasMany)
- Team (User Field)

Table Team:

- Title
- Projects
- Status (Todo/Done Single Select)
- Team (Lookup Project->Team)

### 2) Create Filter
Where Team contains User
AND Status is Not Done

### 3) Activate Filter by changing Status to Done

Triggers `Page Loading Error`
`UI Error : TypeError: data[field].split is not a function`
I have another table where it says `TypeError: Q.split is not a function`

### Desired Behavior

For Live Row Filtering to work with a Lookup Table as a secondary AND filter

### Project Details

Node: **v22.12.0**
Arch: **x64**
Platform: **linux**
Docker: **false**
RootDB: **sqlite3**
PackageVersion: **0.262.1**

(deployed from develop branch)

### Attachments

https://github.com/user-attachments/assets/c301bd0a-425a-4535-bb9b-f5ee6ef0ef2f

ssweber:
Correctly handle 'Lookup' fields as user fields during filtering

## Change Summary
Closes #10728 

When a Lookup field was actually a User-type field, filtering broke.

## Change type

## Test/ Verification

I tested before/after. 3rd pr for this, but this is the smallest change and I think correct! Sorry for all the noise.


ssweber:
@pranavxc See what you think!

packages/nc-gui/utils/dataUtils.ts
```diff
@@ -327,7 +327,7 @@ export function validateRowFilters(_filters: FilterType[], data: any, columns: C
             break
         }
 
-        if ([UITypes.User, UITypes.CreatedBy, UITypes.LastModifiedBy].includes(column.uidt!)) {
+        if ([UITypes.User, UITypes.CreatedBy, UITypes.LastModifiedBy, UITypes.Lookup].includes(column.uidt!)) {
```
pranavxc, line 330: Only if **Lookup** pointing to **User** type we have to do this.

ssweber: You're right. I think this issue goes a bit deeper, and we'll want to test for our filter getting triggered with a Local condition + OR of all kinds to different Lookup records. I'll kick it around next week, or if it's something you already have a good idea how to resolve, please do.

ssweber: FYI. Baserow just straight up prevents you doing a filter to a Lookup record. You have create a Formula with totext(LookupField) to first convert it to text. I suppose they did that to prevent lookups with each filter check.

ssweber: Maybe we just catch the exception that it's not a string on the Else and then check if it's a UIType.Lookup (and thus a UITypes.User?)

ssweber: I’ll test this out tomorrow:

```
if (
  // Standard user field types
  [UITypes.User, UITypes.CreatedBy, UITypes.LastModifiedBy].includes(column.uidt!) || 
  // Lookup fields with specific operations and non-string object data
  (column.uidt === UITypes.Lookup && 
   ['anyof', 'nanyof', 'allof', 'nallof', 'empty', 'blank', 'notempty', 'notblank'].includes(filter.comparison_op) &&
   data[field] && 
   typeof data[field] !== 'string' && 
   typeof data[field] === 'object')
) {
  // Process with user field logic
  // ...
}

**Comment by ssweber** [comment]
@pranavxc want to take a look at this now?

Instead of checking if every Lookup points to a User (which you'd need to do for every Lookup column encountered), I'm checking the data structure at runtime. If `typeof data[field] !== 'string'`, we process it like a User field.

The old code was breaking when trying to call .split() on these fields, so I'm using that fact and hoping this fixes that without over complicating things.

**Comment by ssweber** [comment]
For testing I tested Lookup->User, and also Lookup -> text, multi-select, date. Seemed to be working. Is there test to run against filters?

File: packages/nc-gui/utils/dataUtils.ts [hunk 2]
```diff
@@ -328,7 +328,16 @@ export function validateRowFilters(_filters: FilterType[], data: any, columns: C
...
+        if (
+          // Standard User field types
+          [UITypes.User, UITypes.CreatedBy, UITypes.LastModifiedBy].includes(column.uidt!) || 
+          // Lookup fields that are likely User field types
+          (column.uidt === UITypes.Lookup && 
+           ['anyof', 'nanyof', 'allof', 'nallof', 'empty', 'blank', 'notempty', 'notblank'].includes(filter.comparison_op) &&
+           data[field] && 
+           typeof data[field] !== 'string' && 
+           typeof data[field] === 'object')
+        ) {
```
pranavxc, line 340: It’s still not enough—field values can be objects in different scenarios. To ensure it points to a UserType, we need to verify the target column. This requires fetching the linked table’s column and checking the type of the looked-up column.

packages/nc-gui/utils/dataUtils.ts
```diff
@@ -169,7 +169,7 @@ export const isRowEmpty = (record: Pick<Row, 'row'>, col: ColumnType): boolean =
   return !isValidValue(record.row[col.title])
 }
 
-export function validateRowFilters(_filters: FilterType[], data: any, columns: ColumnType[], client: any) {
+export function validateRowFilters(_filters: FilterType[], data: any, columns: ColumnType[], client: any, metas?: Record<string, any> = {}) {
```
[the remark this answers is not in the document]
ssweber: made it mandatory

pranavxc (approved):

ssweber:
Thank you

Issue #9061, opened 2024-07-24
🔦 Feature: New Record Form/Row on Grid - Group By, column data auto filled by associated colum that been "Group By"


amincahcepu:
### Please confirm that the feature request does ***not*** already exist

- [X] I confirm there is no existing issue for this feature request.

### Use case

Hello, 

I have condition when i prefer adding new records in grid - group by.

Assumption
2 Table, A and B, 
relation is 1 to n

On B Grid, I group using A column

Current condition
When expanding a group, there will be button of "New Record Form"/"New Record Grid"
If I choose form, all fields are empty,
If I choose grid,  because all fields is empty, then it will not shown group by

### Suggested solution

 I hope system can fill the group by- field form, with group by key .
' because when the group by is alot,  it is pain to search the group key again.
if this feature exist, it will speed up the form fill.

I know,  I can add new B records using Grid A, using B FK Column,  but using that workaround, I cannot see more column of B data(limitied).

loom video link below

### Additional context

[https://www.loom.com/share/fe5710d0f51e419ba11f8bbbcf291769?sid=3afe3fad-8c76-4c88-ab74-2f19ac59d0e1](url)

ssweber:
I want to ping this issue as well. This would be a huge time saver, and is somewhat surprising behavior currently.

In Grid view with Grouping enabled, automatically assign the Group field value to new records created via 'New Record - Grid' or 'New Record - Form'. Currently, users must manually input the Group field value, which is tedious and error-prone. Additionally, new records without the Group field value immediately disappear from view, causing confusion and disrupting workflow.

https://github.com/user-attachments/assets/7cfe0f49-49a9-4ddd-8460-bbdf18531105

ssweber:
@DarkPhoenix2704 I read in a different issue that Grid 2.0 is in progress, and it uses old code for grouping.

My question, if I wanted to investigate implementing this feature… would a pr be welcome?

DarkPhoenix2704:
Hi @ssweber,
The Grid 2.0 has already been released. However, for groupby functionality we're still using the older version of the grid and working on implementing the new one.

This issue isn't related to groupby but is rather a generic one. The same can be reproduced in Gallery view, when users apply filters in gallery view and click on a new record, values should be autofilled based on those filters. If you'd like to work on this issue, I can assign it to you.

ssweber:
Hi @DarkPhoenix2704,

[...]

PR #10902, opened 2025-03-18
Fix for LTAR Record Grouping in NocoDB

ssweber:
This change fixes an issue in NocoDB where records grouped under a "Link to Another Record" (LTAR) relationship weren't automatically associating with their parent record when using either "New Record - Form" or "New Record - Grid" interfaces. The key improvements include:

    Added new helper functions:
         isGroupBtLTAR: Detects if a column is a "Belongs To" LTAR type
         getBtLTAR: Retrieves parent record information for grouped records

    Enhanced empty row creation:
         When adding a new row in a grouped view, the system now automatically associates it with the parent record
         Initializes an ltarState property to track LTAR relationships
         Uses promises to fetch and link parent records asynchronously

    Fixed column type filtering:
         Updated the condition that determines which columns should receive default values
         Properly handles LTAR columns in grouped views

This change ensures that when users add records within a grouped view based on a "Belongs To" relationship (whether using form or grid entry), the parent-child association is automatically established without requiring manual selection.

## Change Summary

## Change type

- [x ] feat: (new feature for the user, not a new feature for build script)
- [x ] fix: (bug fix for the user, not a fix to a build script)

I correctly handle when Record isn't Belongs-to. 
1) We need a good way of finding quickly when a parent column isn't unique
2) Is there a simpler way of identifying parent column by Group.column_text, and then id of parent record?
3) Need help when using 'New Record - Form'. It correctly adds, but how do you 'sync' them so that if you click `+` it shows it's linked? And sometimes there is a delay before it shows up (like with nested groups).

### DIFF
--- a/packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
+++ b/packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
@@ -175,7 +175,7 @@ function addEmptyRow(group: Group, addAfter?: number, metaValue = meta.value) {
     if (
       curr.key !== '__nc_null__' &&
       // avoid setting default value for rollup, formula, barcode, qrcode, links, ltar
-      (!curr.column_uidt == UITypes.LinkToAnotherRecord && !isGroupBtLtar(curr.column_name))
+      !(curr.column_uidt == UITypes.LinkToAnotherRecord && !isGroupBtLTAR(curr.column_name)) &&
       ![UITypes.Rollup, UITypes.Lookup, UITypes.Formula, UITypes.Barcode, UITypes.QrCode, UITypes.Links].includes(curr.column_uidt)
     ) {
       acc[curr.title] = curr.key
--- a/packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
+++ b/packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
@@ -36,6 +36,10 @@ const router = useRouter()
 
 const meta = inject(MetaInj, ref())
 
+const { getMeta } = useMetas()
+
+const { api } = useApi()
+
 const view = inject(ActiveViewInj, ref())
 
 const isPublic = inject(IsPublicInj, ref(false))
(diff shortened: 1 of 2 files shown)


ssweber:
fixes #9061

ssweber:
@DarkPhoenix2704 Could you take a look at this?

packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
```diff
@@ -98,6 +102,70 @@ const addRowExpandOnClose = (row: Row) => {
   eventBus.emit(SmartsheetStoreEvents.CLEAR_NEW_ROW, row)
 }
 
+function isGroupBtLTAR(column_name: string, metaValue = meta.value): boolean {
+  const ltarColumn = metaValue?.columns?.find(col => col.title === column_name);
+  // Return true if it's a BT LTAR column, otherwise false
+  if (!ltarColumn?.colOptions) return false;
+  return isBt(ltarColumn);
+}
```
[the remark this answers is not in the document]
ssweber: Not sure how this works, but it exists :)

packages/nc-gui/components/smartsheet/grid/Table.vue
```diff
@@ -787,10 +787,6 @@ async function saveEmptyRow(rowObj: Row) {
 function addEmptyRow(row?: number, skipUpdate = false) {
   const rowObj = callAddEmptyRow?.(row)
 
-  if (!skipUpdate && rowObj) {
-    saveEmptyRow(rowObj)
-  }
-
```
ssweber, line 793: It doesn't seem like this is needed. Deleting it actually fixed the problem I was having with Grid-mode not accepting the new row.

packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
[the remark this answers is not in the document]
ssweber: ok I did that

packages/nc-gui/components/smartsheet/grid/GroupByTable.vue
```diff
@@ -107,8 +175,8 @@ function addEmptyRow(group: Group, addAfter?: number, metaValue = meta.value) {
     if (
       curr.key !== '__nc_null__' &&
       // avoid setting default value for rollup, formula, barcode, qrcode, links, ltar
-      !isLinksOrLTAR(curr.column_uidt) &&
-      ![UITypes.Rollup, UITypes.Lookup, UITypes.Formula, UITypes.Barcode, UITypes.QrCode].includes(curr.column_uidt)
+      !(curr.column_uidt == UITypes.LinkToAnotherRecord && !isGroupBtLTAR(curr.column_name)) &&
```
ssweber, line 178: I think I messed this up and will push a change for it.

ssweber: Should be

if it’s a LinktoAnotherRecord and Isn’t a BelongsTo Ltar

ssweber:
Working.

I need help with fixing this:
https://github.com/user-attachments/assets/0d9425ff-5c42-4f48-8c7e-e4685b2706a6

ssweber:
@pranavxc , @dstala I think this is in a good spot now, unless either of you have suggestions on a better way than an api-call to get a Groups parent Id of a Bt LTAR.

This feature is highly desired by my team. It was the first “surprise” when we setup a parent Projects -> has many -> Tasks. It’s natural to group-by Projects in the Tasks table and then be confused why the new line disappears after clicking “New Record” underneath the grouping.

Only explaining this because I see the team is very busy 😀 Thank you.

ssweber:
Any feedback on this pr?

ssweber:
@DarkPhoenix2704 Do you have 5 minutes to give any feedback on this? I've tested and it's working as intended (and has logic to avoid modifying existing behavior)

dstala:
@ssweber we are working on major changes for group-by -- to introduce canvas based rendering + infinite scroll. @pranavxc will soon check this out to see if this can be integrated in anyway.

packages/nc-gui/components/smartsheet/grid/Table.vue
pranavxc: Currently, newly added rows are not saved unless the user modifies a cell value, which is not the expected behaviour.

cc @dstala 


https://github.com/user-attachments/assets/a29853fa-97db-4ea6-85cf-3d4f47d7a77b

ssweber:
Thanks @dstala . I'll poke around the changes brought in via nc-canvas-group and see how to update this pr to bring the same functionality to the canvas version.

pranavxc (commented):
Check the following comment - https://github.com/nocodb/nocodb/pull/10902/files#r2036589784

packages/nc-gui/components/smartsheet/grid/Table.vue
ssweber: Good catch. I just figured out how to toggle-off the experimental groupby. I'll take a look at this over the weekend and see how to restore the expected behavior.

ssweber: I think the problem I had was `addEmptyRow` really needed to be made async, and then `await` the `callAddEmptyRow`. Otherwise ltarstate is empty when the row is saved, since we're doing api calls.

This doesn't work:

```
async function saveEmptyRow(rowObj: Row) {
  // the ltarState will be empty here, but logs properly under rowObj
  await updateOrSaveRow?.(rowObj, '', rowObj.rowMeta.ltarState || {})
}
function addEmptyRow(row?: number, skipUpdate = false) {
  const rowObj = callAddEmptyRow?.(row)
  if (!skipUpdate && rowObj) {
    saveEmptyRow(rowObj)
  }
```

Should I ditch this version and see about adding it properly to the canvas-groupby instead? Or will the team look it into it?

ssweber: As you can see I'm out of my depth with vue/javascript

ssweber:
I think I'll close this one and reopen with one when I have a bit more time that integrates the LTAR information on the initial Groupby loading, so that there is no delay on Grid/Form creation due to an API call. I'll target the new canvas-groupby.

Issue #11118, opened 2025-04-16
🐛 Bug: row moved even when field modified is not filtered


dstala:
### Please confirm that the bug report does ***not*** already exist

- [x] I confirm there is no existing issue for this bug.

### Steps to reproduce

https://www.loom.com/share/e12ccc57603c4c6383c538e1544a7355?sid=5a6a3170-6cd2-4a1a-922b-de55717605c7

### Desired Behavior

Row should move out of view only if its filtered

### Project Details

x

AMEESHAARACKAL:
The value is filtering because of the exact date filter , **but it is not filtered out at the first place when a new record is created (the default value field ,may have been autocreated using Today/Current Date)** , It gets filtered when a record's (which has exact date not equal ) any field is edited 

- [ ] The root bug is that filtering is not performed when a new record is created (which is a display value field , that has default value)

https://github.com/user-attachments/assets/4df8c9cf-a4c4-4647-9139-7015edeb6b35

Steps to Reproduce :

1. Create a default value field as display field eg:Date Time
2. Create other fields
3. Add filter - also include the default value field  - use **exact date** in the filter but a different date 
4. Add new records  _The records will be added but not filtered out_
5. Change value in any other field in a **record whose display value != exact date**
_The row gets filtered_

PR #11121, opened 2025-04-16
fix: grid related issues

DarkPhoenix2704:
## Change Summary

> expand record -- navigation breaks after current page limit
https://github.com/nocodb/nocohub/issues/5112
> group by field label overflow needs fix
> Count & aggregation font same size? They appear not to be in same line
> Closes https://github.com/nocodb/nocodb/issues/11118
> row moved line is partial

## Change type


AMEESHAARACKAL (approved):
sync pr

PR #11143, opened 2025-04-18
fix (nocodb): json cdf when passed value is number or boolean

fendy3002:
## Change Summary

fix (nocodb): json cdf when passed value is number or boolean

## Change type


AMEESHAARACKAL (approved):
sync pr - https://github.com/nocodb/nocohub/pull/5144

PR #11196, opened 2025-04-24
fix: webhookPluginList permission

DarkPhoenix2704:
## Change Summary

Closes https://github.com/nocodb/nocohub/issues/5185#top

## Change type

### DIFF
--- a/packages/nocodb/src/utils/acl.ts
+++ b/packages/nocodb/src/utils/acl.ts
@@ -38,6 +38,7 @@ const permissionScopes = {
     'pluginTest',
     'pluginRead',
     'pluginUpdate',
+    'webhookPluginList',
 
     // Misc
     'commandPalette',
@@ -330,6 +331,7 @@ const rolePermissions:
       userUpdate: true,
       userDelete: true,
       generateResetUrl: true,
+      webhookPluginList: true,
       userInviteResend: true,
       upload: true,
       uploadViaURL: true,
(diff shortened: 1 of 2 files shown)


AMEESHAARACKAL (approved):
closes https://github.com/nocodb/nocodb/pull/11196 

- Error message is not thrown
- Webhook can be created for both Super Admin and Organizational Level Creator

https://github.com/user-attachments/assets/62f001fe-b701-4b58-a9cb-0180b8672bc3

packages/nocodb/src/utils/acl.ts
```diff
@@ -330,6 +331,7 @@ const rolePermissions:
       userUpdate: true,
       userDelete: true,
       generateResetUrl: true,
+      webhookPluginList: true,
```
pranavxc, line 334: This might make the plugin config accessible to users with the Creator role.

cc @dstala

DarkPhoenix2704: The webhook_url and channel name are stored along with webhook config. So even if a superAdmin created a webhook with slack/teams, the webhook_url would be accessible to base_creator.

dstala: For me, user with creator access should be able to create webhook. if there is a way not to expose URL & just let user select from the dropdown - it would be preferred.

DarkPhoenix2704: > if there is a way not to expose URL & just let user select from the dropdown - it would be preferred.

We dont have any unique id per channel and we do not have the channel name as unique. So we won't be able to drop the webhook_url.

PR #11357, opened 2025-05-13
[fix] - Portuguese Correction

rovierr:
Visualizaç~eos: Tantao pública -> Visualizações: Tanto pública


dstala:
Please use crowdin for i18n corrections. details here
https://docs.nocodb.com/engineering/translation/#using-crowdin

PR #11384, opened 2025-05-15
feat: introduce `NC_SMTP_REJECT_UNAUTHORIZED`

salim-b:
## Change Summary

Allows to configure the SMTP plugin's `rejectUnauthorized` attribute via the environment variable `NC_SMTP_REJECT_UNAUTHORIZED`.

Furthermore fixes parsing `NC_SMTP_SECURE` and `NC_SMTP_IGNORE_TLS` – before *any* value was interpreted as `true` incl. e.g. `NC_SMTP_SECURE="false"`.

## Change type

## Test/ Verification

None.

Currently, after the SMTP plugin is configured via environment variables, the plugin's UI page looks as follows:

Screenshot from 2025-05-15 20-58-55

The <kbd>From domain</kbd> field (corresponding to the plugin's `name` attribute) cannot be configured via environment variables, although it is [marked as *required*](https://github.com/nocodb/nocodb/blob/cbe16b378a5f1f10a976a360c89c66b20f7d98f2/packages/nocodb/src/plugins/smtp/index.ts#L43).

The <kbd>Reject unauthorized</kbd> toggle also can't be configured via environment variables.

This PR should make configuration of the latter possible via the environment variable `NC_SMTP_REJECT_UNAUTHORIZED`.

First, I wanted to also introduce `NC_SMTP_DOMAIN` to configure the SMTP plugin's `name` attribute. But then I realized that setting this has no effect on sent e-mails. What is its intended purpose again?


pranavxc (approved):

salim-b:
Thanks for merging, @pranavxc!

Do you also know why setting the <kbd>From domain</kbd> field (corresponding to the SMTP plugin's `name` attribute) has no effect?

salim-b:
> It’s used as the name property, and according to the **Nodemailer** documentation, it serves the following purpose: - https://nodemailer.com/extras/smtp-server#options-reference
> Screenshot 2025-05-17 at 5 53 01 PM

Ok, I tried to dig into that whole SMTP banner topic and as it turns out it doesn't apply to how NocoDB uses Nodemailer here: in SMTP *client* mode to connect to a third-party SMTP server.

The `name` option is only relevant when Nodemailer is used to run your own SMTP server (the `name` option then specifies the hostname that the SMTP server announces in its own banner).

So I think the `name` option should be removed from NocoDB since it has no effect and is just confusing.

PR #11639, opened 2025-06-11
feat: fill handler improvement

fendy3002:
## Change Summary

feat: fill handler improvement

## Change type


dstala:
- [ ] percent & currency — write two cells with 0 & 1. Select both & drag — the behaviour is different compared to number / decimal

dstala (approved):

PR #11734, opened 2025-06-23
feat: revise audit architecture

mertmit:
## Change Summary

## Change type


dstala:
- [ ] after clicking load earlier, if i press `esc` — it doesn’t close modal. i need to click again on the modal and then click `esc`

Issue #11813, opened 2025-07-04
🐛 Bug: Columns order on CSV export


mrekin:
### Please confirm that the bug report does ***not*** already exist

- [x] I confirm there is no existing issue for this bug.

### Steps to reproduce

Hi!
Didn't found similar issue.

I have simple small table with <10 columns and rows.
I order columns and sort rows.

But exported CSV file has another tables order and I can't understand how to get needed result.
Table. It's not depend if I export from table or from view.

NocoDB: 0.263.8, docker installation.

![Image](https://github.com/user-attachments/assets/fdb7c275-b53c-4196-aa2d-33a4c4e6ff46)

CSV

![Image](https://github.com/user-attachments/assets/393dc711-1f60-453b-b8e8-ccca220980f7)

### Desired Behavior

Exported CSV has similar data order as source view/table.

### Project Details

NocoDB: 0.263.8, docker installation.
Database: PG

PR #11865, opened 2025-07-10
fix: use view column order when downloading csv

fendy3002:
## Change Summary

fix: use view column order when downloading csv. Ref: https://github.com/nocodb/nocodb/issues/11813

## Change type


fendy3002:
already synced by mert

PR #11932, opened 2025-07-20
fix: validate plugin input before update to prevent invalid plugin save

### LINKED ISSUE #11751: 🔦 Feature: Test E-mail / Storage configuration automatically after pressing Save

### Use case

After configuring NocoDB Storage / E-mail settings and pressing Save there is no connectivity validation, only if you press Test before saving.

NocoDB then tries to use the following configuration when using the web interface, and it can sometimes throw errors that are missleading(TLS errors, socket hangup, and so on).

Example: badly S3 Storage configured can cause TLS errors(write EPROTO … ssl3_get_record:wrong version number) when uploading files or exporting tables as CSV. I was blaming my connectivity with the postgres database, even tho SSL was off.

### Suggested solution

After pressing the Save button, a test should be made before and passed in order to save the Storage / E-mail configuration.

### Additional context

_No response_

ArVaViT:
## Change Summary

Fixes [#11751](https://github.com/nocodb/nocodb/issues/11751)

This PR ensures that plugin input is validated using `NcPluginMgrv2.test()` **before saving**, preventing invalid configurations from being stored.

Previously, a plugin could be saved without testing — leading to invalid configurations that only cause errors later during execution.

## Change type

- [x] fix (bug fix for the user, not a fix to a build script)

### Code added:
```ts
const pluginInfo = await Plugin.get(param.pluginId);
if (pluginInfo?.title && pluginInfo.category) {
  await NcPluginMgrv2.test({
    title: pluginInfo.title,
    category: pluginInfo.category,
    input: param.plugin.input,
  });
}

### DIFF
--- a/packages/nocodb/src/services/plugins.service.ts
+++ b/packages/nocodb/src/services/plugins.service.ts
@@ -37,6 +37,14 @@ export class PluginsService {
     req: NcRequest;
   }) {
     validatePayload('swagger.json#/components/schemas/PluginReq', param.plugin);
+    const pluginInfo = await Plugin.get(param.pluginId);
+    if (pluginInfo?.title && pluginInfo.category) {
+      await NcPluginMgrv2.test({
+        title: pluginInfo.title,
+        category: pluginInfo.category,
+        input: param.plugin.input,
+      });
+    }
 
     const plugin = await Plugin.update(param.pluginId, param.plugin);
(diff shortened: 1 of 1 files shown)


ArVaViT:
CI failing on latest develop too. Doesn't seem related to this PR.

PR #12002, opened 2025-07-30
chore: Revert Dockerfile changes since it's breaking arm/v7 build

pranavxc:
## Change Summary

Ref: https://github.com/nocodb/nocodb/actions/runs/16566645108/job/46849371081

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nocodb/Dockerfile
+++ b/packages/nocodb/Dockerfile
@@ -1,46 +1,44 @@
-# syntax=docker/dockerfile:1
-
-####################
+###########
 # Litestream Builder
-####################
-FROM golang:bookworm AS lt-builder
+###########
+FROM golang:bullseye AS lt-builder
 
 WORKDIR /usr/src/
 
-# Install buildtime dependencies
-RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


mertmit (approved):

salim-b:
@pranavxc I don't think you need to revert all the changes. Changing `FROM golang:bookworm` back to `FROM golang:bullseye` should suffice. Also, I would add a comment explaining that `golang:bookworm` fails to build.

Ideally, I'd also try with `golang:1.24` same as [upstream](https://github.com/benbjohnson/litestream/blob/main/Dockerfile#L1) to see if this resolves ARM/v7 builds.

pranavxc:
> @pranavxc I don't think you need to revert all the changes. Changing `FROM golang:bookworm` back to `FROM golang:bullseye` should suffice. Also, I would add a comment explaining that `golang:bookworm` fails to build.
> 
> Ideally, I'd also try with `golang:1.24` same as [upstream](https://github.com/benbjohnson/litestream/blob/main/Dockerfile#L1) to see if this resolves ARM/v7 builds.

@salim-b  Since we are not exactly sure which part is causing the issue, we reverted all changes for now as we want to roll out a bug fix release.

PR #12057, opened 2025-08-07
Nc fix(nc-gui): Skip toast error during link creation if table is not configured

rameshmane7218:
## Change Summary

## Change type

## Test/ Verification

Provide summary of changes.

### DIFF
--- a/packages/nc-gui/composables/useColumnCreateStore.ts
+++ b/packages/nc-gui/composables/useColumnCreateStore.ts
@@ -80,12 +80,18 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
 
     const additionalValidations = ref<ValidationsObj>({})
 
+    const avoidShowingToastMsgForValidations = ref<{ [key: string]: boolean }>({})
+
     const setAdditionalValidations = (validations: ValidationsObj) => {
       additionalValidations.value = { ...additionalValidations.value, ...validations }
     }
 
     const removeAdditionalValidation = (key: string) => {
       delete additionalValidations.value[key]
+    }
+
+    const setAvoidShowingToastMsgForValidations = (validations: { [key: string]: boolean }) => {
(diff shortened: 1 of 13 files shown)


packages/nc-gui/composables/useColumnCreateStore.ts
```diff
@@ -345,8 +352,15 @@ const [useProvideColumnCreateStore, useColumnCreateStore] = createInjectionState
       try {
         if (!(await validate())) return
       } catch (e: any) {
+        let skipToast = false
         const errorMsgs = (e.errorFields || [])
-          .filter((f) => f?.name !== 'cdf')
+          .filter((f) => {
+            if (avoidShowingToastMsgForValidations.value[f?.name ?? '']) {
+              skipToast = true
+            }
```
pranavxc, line 360: If there are two fields and one is configured to be skipped while the other is not, what should the behaviour be in that case?

rameshmane7218: Toast message will be show for those field which are not added in `avoidShowingToastMsgForValidations`

rameshmane7218: `skipToast` flag is just used to avoid `message.error(t('msg.error.formValidationFailed'))`

```ts
let skipToast = false
const errorMsgs = (e.errorFields || [])
  .filter((f) => {
    if (avoidShowingToastMsgForValidations.value[f?.name ?? '']) {
      skipToast = true
    }

    return f?.name !== 'cdf' && !avoidShowingToastMsgForValidations.value[f?.name ?? '']
  })
  .map((e: any) => e.errors?.join(', '))
  .filter(Boolean)
  .join(', ')

if (errorMsgs) {
  message.error(errorMsgs)
  return
}

if (skipToast) return

if (!fromKanbanStack?.value || (fromKanbanStack.value && !e.outOfDate)) {
  message.error(t('msg.error.formValidationFailed'))
  return
}
```

PR #12139, opened 2025-08-20
fix: track and skip duplicates on remote import

mertmit:


packages/nocodb/src/modules/jobs/jobs/export-import/import.service.ts
```diff
@@ -2044,10 +2045,20 @@ export class ImportService {
...
+      idMap,
+      dataStream,
+      destBase,
+      destProject,
+      destModel,
+      req,
+      skipDuplicates = false,
+    } = param;
+
+    const dataHashMap: Map<string, boolean> = new Map();
```
fendy3002, line 2061: can use `Set` instead `Map`

packages/nocodb/src/modules/jobs/jobs/export-import/import.service.ts
```diff
@@ -2109,7 +2120,18 @@ export class ImportService {
                   }
                 }
               }
-              chunk.push(row);
+
+              if (skipDuplicates) {
+                // check if the row already exists
+                const rowHash = hash(row);
```
fendy3002, line 2126: if the row contains primary it won't work

mertmit: I couldn't get this, can you elaborate?

fendy3002: I mean if the row contains primary key value, the hash will be unique. Otherwise it works

fendy3002: let's defer for now since I just realized this is nocodb changes. we can discuss this later

PR #12435, opened 2025-09-30
feat: nc-db-max-connection-pool-env

### LINKED ISSUE #9973: 🔦 Feature: Configure database max. connection pool size through an environment variable

- [X] I confirm there is no existing issue for this feature request.

### Use case

If you have a powerful database instance/setup that can handle hundreds of connections (or otherwise), it makes sense to have the ability to scale the connection pool size, so that you can utilize and harness the database fully.

### Suggested solution

When constructing the default configuration for the db, if the environment variable for configuring the max. pool size is set (say e.g. `DB_MAX_POOL_SIZE`) set the _max_ property on _pool_ object to this value. Else just leave the default value as is (in this case `10`).

### Additional context

Should this get into the codebase, I'm assuming that it'll be important to keep in mind to update [docs.nocodb.com](https://docs.nocodb.com/getting-started/self-hosted/environment-variables) environment variable section to reflect this addition.

hemantmm:
closes: #9973

## Change Summary

✅ Added support for configuring database connection pool size via an environment variable.
✅ Updated the default database configuration to read `NC_DB_MAX_POOL_SIZE`.
✅ If `NC_DB_MAX_POOL_SIZE` is set, the max property of the connection pool uses this value; otherwise, it defaults to 10.

## Change type

## Test/ Verification

Provide summary of changes.

The changes allow users to:

- Scale database connection pools dynamically by setting `NC_DB_MAX_POOL_SIZE` in their environment.
- Retain default connection pool size of 10 if the environment variable is not set.
- All pool configurations now use a consistent and safe parsing approach for the environment variable,

### DIFF
--- a/packages/nocodb/src/db/sql-migrator/lib/templates/mysql.template.ts
+++ b/packages/nocodb/src/db/sql-migrator/lib/templates/mysql.template.ts
@@ -14,6 +14,7 @@ export default {
             password: 'password',
             database: 'default_dev',
           },
+          pool: { min: 0, max: +process.env.DB_MAX_POOL_SIZE || 10 },
           meta: {
             tn: 'nc_evolutions',
             dbAlias: 'primary',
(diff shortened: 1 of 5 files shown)


o1lab:
Better to prefix `NC_` as this is related `NC_DB` configuration

PR #12468, opened 2025-10-03
fix: handle programmatic errors in api

DarkPhoenix2704:
## Change Summary

fix: handle programmatic errors in api

## Change type


DarkPhoenix2704:
TBD: Will be synced manually

PR #12550, opened 2025-10-16
feat: list and manage user's mcp tokens in account settings

DarkPhoenix2704:
## Change Summary

feat: list and manage mcp tokens in account level

## Change type


DarkPhoenix2704:
Will be synced manually

Issue #12580, opened 2025-10-22
🐛 Bug: The "does not contain any of" filter doesn't work.


p0ebounty:
### Please confirm that the bug report does ***not*** already exist

- [x] I confirm there is no existing issue for this bug.

### Steps to reproduce

1. You need two tables that will be linked via the "Links" column.
2. Next, in the child table, add a "Single Select" field and create some values.
3. After that, in the parent table, add a "Lookup" field based on the Single Select field from the child table.
4. And then create a Filter on the LookUp field (does not contain any of) to exclude some values ​​from the field from the child table.

### Desired Behavior

Rows that you excluded based on the filter condition should not be displayed.

However, this type of filter does not work.

### Project Details

Nocodb - 0.265.1 - docker
DB - Postgres

### Attachments

https://github.com/user-attachments/assets/afdfd135-5d63-46eb-9fd4-6b45d86caf20

vanishajr:
Hey, @p0ebounty ,

I have opened a Pull Request here ( #12581 )
Please take a look when you have a moment.
Thank you!

p0ebounty:
Hello @vanishajr, thank you for your quick response.

Your video in the commit does indeed show the fixed issue. However, I also decided to test the same configuration on my server instance—I created two new tables and configured the same data architecture. The bug didn't appear.

In my video, where I showed an example using the Leads and Events tables, there were quite a lot of records (14K for leads and 21K for events), and the bug was still there. Perhaps it's due to the volume of records?

I tested the error on another instance of mine, which has the same configuration of these two tables and also has a large number of records—the bug is still there.

But the bug doesn't appear on smaller tables.

@fendy3002

vanishajr:
> Hello [@vanishajr](https://github.com/vanishajr), thank you for your quick response.
> 
> Your video in the commit does indeed show the fixed issue. However, I also decided to test the same configuration on my server instance—I created two new tables and configured the same data architecture. The bug didn't appear.
> 
> In my video, where I showed an example using the Leads and Events tables, there were quite a lot of records (14K for leads and 21K for events), and the bug was still there. Perhaps it's due to the volume of records?
> 
> I tested the error on another instance of mine, which has the same configuration of these two tables and also has a large number of records—the bug is still there.
> 
> But the bug doesn't appear on smaller tables.
> 
> [@fendy3002](https://github.com/fendy3002)

Thanks for the review! I will test the filter behavior on larger databases to try and reproduce the issue .

[...]

PR #12581, opened 2025-10-22
fix: does not contain any of filter

vanishajr:
## Change Summary

Closes: #12580  

Where the "does not contain any of" filter did not correctly exclude values on Lookup fields in NocoDB. The root cause was the missing negated operators in the negatedMapping object used by the filter processing logic. This fix adds the mappings for nanyof and nallof to their positive counterparts (anyof and allof), enabling the proper evaluation of negated filters for Lookup fields.

## Change type
fix: Bug fix for user-facing filter behavior on Lookup fields.

## Test/ Verification
- Verified reproduction steps manually by linking two tables (Leads and Events) with Single Select and Lookup fields.- 
- Applied the "does not contain any of" filter on the Lookup field and confirmed correct rows were included/excluded after the fix.
- Ran existing automated tests related to filters and confirmed no regressions.

https://github.com/user-attachments/assets/9f98a909-2aa4-4f9d-b2bd-71b4bf12a991

### DIFF
--- a/packages/nocodb/src/db/field-handler/utils/handlerUtils.ts
+++ b/packages/nocodb/src/db/field-handler/utils/handlerUtils.ts
@@ -26,6 +26,8 @@ export const negatedMapping = {
   neq: { comparison_op: 'eq' },
   blank: { comparison_op: 'notblank' },
   notchecked: { comparison_op: 'checked' },
+  nanyof: { comparison_op: 'anyof' },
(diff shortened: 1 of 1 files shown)


dstala:
@vanishajr Please sign here
https://github.com/nocodb/nocodb/pull/12581#issuecomment-3434078969

fendy3002 (approved):

PR #12724, opened 2025-12-11
fix(api-v3): return full linked records instead of counts for UITypes.Links

### LINKED ISSUE #12717: 🐛 Bug: Endpoint does not return the full list of paginated linked records inline

### Steps to reproduce

Create two tables. Use the "Link" field. Fetch data via api.

### Desired Behavior

The nested field is just a count not an array of nested objects like described in the release notes here:

https://nocodb.com/docs/changelog/2025.06.0#api-v3-beta--fewer-calls-cleaner-design

Embedded Relation Field Data
v2: When you listed records, relation fields returned only a count of linked records.
v3: The same endpoint now returns the full list of paginated linked records inline, so you can fetch related details in a single call—no more chasing separate endpoints.

```
{
  "records": [
    {
      "id": 1,
      "fields": {
        "Title": "post 1",
        "CreatedAt": "2025-12-09 08:03:40+00:00",
        "UpdatedAt": "2025-12-09 08:04:57+00:00",
        "post_interactions": 1
      }
    }
  ],
  "nestedNext": null
}
```

If this were fixed, my [other issue](https://github.com/nocodb/nocodb/issues/12693) would probably be resolved.

### Project Details

NocoDB Cloud

### Attachments

_No response_

iamgerwin:
## Summary
- Updates `getAst.ts` to treat `UITypes.Links` the same as `LinkToAnotherRecord` when generating the AST for nested field queries
- Updates `data-v3.service.ts` to handle `UITypes.Links` columns by looking for linked data with the `_nc_lk_` prefix used by `getProto()`
- Transforms the linked records to v3 format and filters out count values that shouldn't be in the response

## Root Cause
Previously, API v3 endpoints returned numeric counts for `UITypes.Links` columns instead of the actual paginated linked records as documented. This was because:
1. `getAst.ts` had special handling to return `value = 1` (count) for `UITypes.Links` instead of building the AST for nested fields
2. `data-v3.service.ts` only handled `UITypes.LinkToAnotherRecord` in the transformation, not `UITypes.Links`

## Test plan
- [x] Create two tables with a Link column between them
- [x] Add records with linked relationships
- [x] Test API v3 endpoint `/v3/{baseId}/tables/{tableId}/records`
- [x] Verify linked records are returned as arrays of objects, not counts
- [x] Test with HAS_MANY, MANY_TO_MANY, BELONGS_TO, and ONE_TO_ONE relationships

## Files Changed
- `packages/nocodb/src/helpers/getAst.ts` - Use `isLinksOrLTAR()` instead of checking only for `UITypes.LinkToAnotherRecord`
- `packages/nocodb/src/services/v3/data-v3.service.ts` - Handle `UITypes.Links` columns in transformation and LTAR column filtering

Fixes #12717

### DIFF
--- a/packages/nocodb/src/helpers/getAst.ts
+++ b/packages/nocodb/src/helpers/getAst.ts
@@ -235,7 +235,7 @@ const getAst = async (
     const nestedFields =
       query?.nested?.[col.title]?.fields || query?.nested?.[col.title]?.f;
     if (nestedFields && nestedFields !== '*') {
-      if (col.uidt === UITypes.LinkToAnotherRecord) {
+      if (isLinksOrLTAR(col)) {
         const colOpt = await col.getColOptions<LinkToAnotherRecordColumn>(
           context,
         );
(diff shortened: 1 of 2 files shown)


fendy3002:
closing this since this is not in plan for implementation.

PR #12744, opened 2025-12-17
fix(nc-gui): correct target base default and refresh sidebar on table duplicate

### LINKED ISSUE #12738: 🐛 Bug: Table copy problem

### Steps to reproduce

I'm in the zzz base, and when I copy a table from the fff base, the target base is also zzz, but the copied table appears in fff. However, if I'm in the fff base and copy a table to zzz, it works normally.

Furthermore, I'm in zzz Base, and I copy a table to fff Base, but fff doesn't update in real time; I need to refresh the page to see the table.

### Desired Behavior

Regardless of the current base, when copying a table, it must be correctly displayed in the target base.

### Project Details

/

### Attachments

_No response_

iamgerwin:
## Summary
- Fix table duplicate dialog to correctly default target base to the table's own base instead of the currently viewed base
- Add sidebar refresh for target base after successful table duplication to another base

## Problem
When copying a table from Base A while viewing Base B:
1. The target base incorrectly defaulted to Base B (the active route's base) instead of Base A (the table's own base)
2. After copying to a different base, the target base's sidebar didn't update in real-time - required a page refresh

## Solution
- Initialize `targetBase` from the table's actual base (`props.table.base_id`) via a computed property that looks up the base in `basesStore`
- After successful duplication to another base, call `tablesStore.loadProjectTables()` to refresh the target base's table list in the sidebar

## Changes
- `packages/nc-gui/components/dlg/Table/Duplicate.vue`
  - Added `tableBase` computed property to get the table's base from basesStore
  - Initialize `targetBase` in `onMounted()` to `tableBase.value`
  - Added `loadProjectTables()` and `refreshCommandPalette()` calls after successful cross-base duplication

Resolves #12738

### DIFF
--- a/packages/nc-gui/components/dlg/Table/Duplicate.vue
+++ b/packages/nc-gui/components/dlg/Table/Duplicate.vue
@@ -25,9 +25,13 @@ const { $e, $poller } = useNuxtApp()
 
 const basesStore = useBases()
 
+const { bases } = storeToRefs(basesStore)
+
 const { createProject: _createProject, loadProjects } = basesStore
 
-const { openTable } = useTablesStore()
+const tablesStore = useTablesStore()
+
+const { openTable } = tablesStore
 
 const baseStore = useBase()
 
@@ -53,7 +57,11 @@ const { getFeature } = useEeConfig()
 const wsDropdownOpen = ref(false)
 const baseDropdownOpen = ref(false)
 const targetWorkspace = ref(activeWorkspace)
-const targetBase = ref(activeBase.value)
+
+// Get the table's base from basesStore - this fixes the bug where target base
+// incorrectly defaulted to activeBase (current route's base) instead of the table's own base
+const tableBase = computed(() => bases.value.get(props.table.base_id!) ?? activeBase.value)
+const targetBase = ref<BaseType | null>(null)
 
 const targetTableMeta = computedAsync(async () => {
   return getMeta(props.table.id!)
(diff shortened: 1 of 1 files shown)


packages/nc-gui/components/dlg/Table/Duplicate.vue
```diff
@@ -53,7 +57,11 @@ const { getFeature } = useEeConfig()
 const wsDropdownOpen = ref(false)
 const baseDropdownOpen = ref(false)
 const targetWorkspace = ref(activeWorkspace)
-const targetBase = ref(activeBase.value)
+
+// Get the table's base from basesStore - this fixes the bug where target base
+// incorrectly defaulted to activeBase (current route's base) instead of the table's own base
+const tableBase = computed(() => bases.value.get(props.table.base_id!) ?? activeBase.value)
```
rameshmane7218: revert this all other change as in UI we show only active base table (single base tables at time), so obviously table which we are duplicating will be part of active base

packages/nc-gui/components/dlg/Table/Duplicate.vue
```diff
@@ -25,9 +25,13 @@ const { $e, $poller } = useNuxtApp()
 
 const basesStore = useBases()
 
+const { bases } = storeToRefs(basesStore)
```
rameshmane7218: remove this

rameshmane7218:
Closing this PR as the review changes were not addressed. The fix has been added in PR #12789.

PR #12961, opened 2026-01-28
fix: mcp queryRecords tools schema error & data-v3 validateDataListQueryParams support mcp sort (array type)

rongfengliang:
mcp queryRecords tools schema error & data-v3 validateDataListQ
## Change Summary

mcp queryRecords tools schema error & data-v3 validateDataListQueryParams support mcp sort (array type)

## Change type

- [x] fix: (bug fix  for mcp queryRecords tools schema error & data-v3 validateDataListQueryParams support mcp sort (array type))

### DIFF
--- a/packages/nocodb/src/mcp/mcp.service.ts
+++ b/packages/nocodb/src/mcp/mcp.service.ts
@@ -206,7 +206,7 @@ export class McpService {
             .array(
               z.object({
                 field: z.string().describe('Field Name'),
-                description: z.enum(['asc', 'desc']).describe('Sort Direction'),
+                direction: z.enum(['asc', 'desc']).describe('Sort Direction'),
               }),
             )
             .optional(),
--- a/packages/nocodb/src/services/v3/data-v3.service.ts
+++ b/packages/nocodb/src/services/v3/data-v3.service.ts
@@ -320,22 +320,29 @@ export class DataV3Service {
   ) {
     const columns = param.modelInfo.columns;
     if (param.query.sort) {
+      
       let fieldsArr: string[] = [];
-      if (typeof param.query.sort !== 'string') {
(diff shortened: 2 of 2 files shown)


let-gh:
Confirming this is still needed — the `mcp.service.ts` half of this PR, at least.

We hit the same bug independently: `queryRecords` accepts a `sort` argument and silently ignores it, returning rows in default order with no error. The cause is the one this PR fixes — the input schema declares the direction key as `description`:

https://github.com/nocodb/nocodb/blob/2026.07.0/packages/nocodb/src/mcp/mcp.service.ts#L206-L213

while `extractSortsObject` reads `s.direction`, so the direction is always `undefined` and the ordering is dropped.

Two things that might help move this along:

1. **The second half of this PR appears to have landed already.** `validateDataListQueryParams` in `services/v3/data-v3.service.ts` now has a `typeof param.query.sort === 'object'` branch that accepts the array form, so as far as I can tell only the schema-key rename is still outstanding — which should make this a one-word diff.

2. **It fails silently rather than erroring** because the direction check short-circuits on falsy values:
   ```ts
   parsedSort.some((s) => s.direction && !['asc', 'desc'].includes(s.direction))
   ```
   An `undefined` direction skips validation entirely. `field` *is* validated, so a bad field name errors while a missing direction doesn't — which is what made this hard to diagnose from the client side. MCP clients can't work around it either, since the unknown `direction` key gets stripped by the zod schema before it reaches the service.

Verified against `2026.06.2`; `mcp.service.ts` is identical at `2026.06.2`, `2026.07.0`, and on the default branch, so current releases are affected.

@rongfengliang this is showing as `CONFLICTING` — any chance you could rebase? Happy to open a rebased minimal PR with just the rename if you'd rather not, just say the word and I'll credit this one.

PR #12970, opened 2026-01-29
fix: insert webhook all fields

### LINKED ISSUE #12831: 🐛 Bug: Webhook receives incomplete data (missing lookup and link fields) when using v3 API

### Steps to reproduce

1. Create a table with LinkToAnotherRecord or Lookup fields
  2. Set up a webhook with after.insert trigger
  3. Insert a record using v3 API: POST /api/v3/tables/{tableId}/records
  4. Check the webhook payload - link and lookup fields are missing or null

  Root cause:
  In packages/nocodb/src/db/BaseModelSqlv2/insert.ts line 392, readByPk is called without query parameters, so it doesn't fetch related fields. However, v3 API returns complete data to users via chunkList (line 684 in data-v3.service.ts), creating inconsistency.

### Desired Behavior

Webhook should receive the same complete data that the API returns to users, including:
  - LinkToAnotherRecord fields
  - Lookup fields
  - All nested relations

### Project Details

  NocoDB used as docker: true
  NocoDB version: latest
  Database used in NC_DB URL: pg
  Project was created by clicking: New Project by connecting to external database
  Database on which spreadsheet is created: pg

  Additional Notes

  - This affects both v2 and v3 APIs
  - Webhook filter conditions fail with Cannot read properties of null (reading '0') when referencing these missing fields
  - Suggested fix: Use chunkList instead of readByPk in insert.ts line 392, similar to how v3 API fetches data

### Attachments

_No response_

fendy3002:
## Change Summary

fix: insert webhook all fields. Fixes https://github.com/nocodb/nocodb/issues/12831

## Change type

### DIFF
--- a/packages/nocodb/src/db/BaseModelSqlv2.ts
+++ b/packages/nocodb/src/db/BaseModelSqlv2.ts
@@ -3104,7 +3104,21 @@ class BaseModelSqlv2 implements IBaseModelSqlV2 {
         const chunk = pkAndData.slice(i, i + readChunkSize);
         const pksToRead = chunk.map((v) => v.pk);
 
-        const oldRecords = await this.chunkList({ pks: pksToRead });
+        const oldRecordChunkList = await this.chunkList({ pks: pksToRead });
+
+        // get ast
+        const { ast, parsedQuery } = await getAst(this.context, {
+          model: this.model,
+          query: {},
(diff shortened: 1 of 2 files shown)


mertmit:
@claude review

mertmit (approved):

PR #13052, opened 2026-02-14
fix: preserve colOptions on metadata-only PATCH for SingleSelect/MultiSelect

### LINKED ISSUE #13044: 🐛 Bug: PATCH on SingleSelect column wipes colOptions even for metadata-only updates
## Bug report

Thank you for maintaining NocoDB – we're big fans and rely on it heavily for our project. We recently reported #12972 (Checkbox PATCH HTTP 500), which was promptly fixed in PR #12974. We really appreciate the quick turnaround! This new issue appears to be the same class of bug, but affecting `colOptions` instead of `cdf`.

### Steps to reproduce

1. Create or have a SingleSelect column with dropdown options configured
2. Send a PATCH request that only updates metadata (e.g. `description`):
   ```bash
   curl -X PATCH \
     'https://<nocodb>/api/v2/meta/columns/<column_id>' \
     -H 'xc-token: <token>' \
     -H 'Content-Type: application/json' \
     -d '{"description": "some description"}'
   ```
3. Check the column's `colOptions` – all dropdown options are gone

### Expected behavior

A PATCH that only sets `description` (or other metadata fields) should **not** affect `colOptions`. The existing dropdown options should be preserved.

### Actual behavior

All `colOptions.options` are silently wiped. The column remains a SingleSelect but with an empty options list. No error is returned – the API responds with 200.

### Likely root cause

PR #12974 fixed the same class of issue for `cdf`: `getColumnPropsFromUIDT` was overwriting `cdf` with defaults even when the PATCH body didn't include it. The same pattern likely applies to `colOptions` – when the PATCH body doesn't include `colOptions`, the column update code path still processes (and clears) the existing options.

### Environment

| Key | Value |
|---|---|
| NocoDB version | 0.301.2 |
| Hosting | Self-hosted on Fly.io |
| Database | External PostgreSQL (Neon), connected as external data source |
| API version | v2 (`/api/v2/meta/columns/{id}`) |

### Additional context

- The bug is reproducible with **any** PATCH payload on a SingleSelect column, not just `description`. Even `{"title": "same_title"}` wipes the options.
- This does not seem to affect columns on the default SQLite data source – we only observed it on an external PostgreSQL data source (though we haven't tested exhaustively).
- Related: #12972, PR #12974

veeceey:
Fixes #13044

### Problem

When sending a PATCH request that only updates metadata fields (like `description` or `title`) on a SingleSelect/MultiSelect column without including `colOptions`, the existing dropdown options get lost.

This happens because the SingleSelect/MultiSelect code path in `columnUpdate` checks `colBody.colOptions?.options` to decide whether to process the options block. When `colOptions` isn't provided, the entire block is skipped -- which means either:
- The metadata update is silently dropped (no `Column.update` or `updateMetaAndDatabase` call happens)
- In certain flows, `Column.update` deletes the existing options but doesn't re-insert them since there are none in `colBody`

This is the same class of bug that was fixed for `cdf` in #12974.

### Fix

After calling `getColumnPropsFromUIDT`, check if `colBody.colOptions?.options` is missing but the existing column has options. If so, carry over the existing options to `colBody`. This way the full update flow proceeds normally, the options are preserved, and metadata-only changes (description, title, etc.) are actually applied.

### Reproduction

```bash
# Create a SingleSelect column with options, then:
curl -X PATCH \
  'https://<nocodb>/api/v2/meta/columns/<column_id>' \
  -H 'xc-token: <token>' \
  -H 'Content-Type: application/json' \
  -d '{"description": "some description"}'

# Before fix: colOptions wiped / description update silently dropped
# After fix: colOptions preserved, description updated
```


veeceey:
@CLAassistant check

fendy3002 (approved):

PR #13073, opened 2026-02-17
chore: remove unused node-sql-parser dependency

guan404ming:
## Summary

- Remove unused `node-sql-parser` from `packages/nocodb/package.json`

## Why

`node-sql-parser` is listed as a dependency but never imported anywhere in the codebase. It is also a large package, adding unnecessary weight to the install.

## Test plan

- [x] Verify `pnpm install` completes without errors
- [x] Verify no runtime errors from missing dependency

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ app/tests/coverage/
 .bower-*/
 .idea/
 coverage/
+.pnpm-store/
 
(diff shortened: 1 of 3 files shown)


guan404ming:
Hi @fendy3002 could you help take a look at this, thanks!

PR #13082, opened 2026-02-19
fix: respect requested fields in shared form linked record list

bittoby:
## Change Summary

Fix linked record fields showing "-" and missing attachment images in shared form views.

In the public `relDataList` endpoint, `extractOnlyPrimaries` was hardcoded to `true`, causing `getAst` to return only primary key and display value columns — ignoring the fields requested by the frontend. This made shared forms show "-" for extra fields (number, date, currency) and no attachment images in the linked record picker, while the form builder preview worked correctly.

The fix aligns with the existing authenticated API pattern in `data-table.service.ts` by conditionally setting `extractOnlyPrimaries` based on whether fields are explicitly requested. Also removes an unreachable dead code block that attempted to patch the AST after early return.

## Change type

## Test/ Verification

1. Create Table A with Text, Number, Date, Currency, and Attachment fields. Add records with data and images.
2. Create Table B with a Links field pointing to Table A.
3. Create a Form view on Table B.
4. Click the Links field in the form builder, enable **"Show Extra Fields"** in the right sidebar.
5. Verify extra fields and attachment images display correctly in the form builder preview.
6. Share the form, open the link in an incognito browser.
7. Click the Links field — extra fields now show actual values and attachment images display correctly.

**File changed:** `packages/nocodb/src/services/public-datas.service.ts`

**Before:** `extractOnlyPrimaries: true` (hardcoded)
**After:** `extractOnlyPrimaries: !(param.query?.f || param.query?.fields)` (matches `data-table.service.ts:473` pattern)

### DIFF
--- a/packages/nocodb/src/services/public-datas.service.ts
+++ b/packages/nocodb/src/services/public-datas.service.ts
@@ -654,7 +654,7 @@ export class PublicDatasService {
     const { ast, dependencyFields } = await getAst(context, {
       query: param.query,
       model,
-      extractOnlyPrimaries: true,
+      extractOnlyPrimaries: !(param.query?.f || param.query?.fields),
     });
 
     const listArgs: DependantFields & {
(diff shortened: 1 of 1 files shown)


bittoby:
@o1lab @pranavxc could you please review this PR? thanks

bittoby:
@mertmit could you please review this PR? 
thank you

mertmit:
Hi @bittoby,

Thanks for the contribution! Unfortunately, we can’t accept this PR in its current form.

The hardcoded `true` is intentional. For publicly shared views we restrict relation fields to only the display value and ID. The actual issue here is the UI showing empty images and dash placeholders and we’d gladly review and accept a PR that addresses that instead.

Allowing additional fields requires internal discussion. We’ll update the issue once we’ve aligned on an approach, so please wait for confirmation from the team before working on it to avoid unnecessary effort.

Thanks again!

PR #13340, opened 2026-03-24
feat(nc-gui,nocodb): allow configurable file size limit for imports

### LINKED ISSUE #12572: 🔦 Feature: Allow custom CSV maximum size for importing

### Use case

I want to be able to upload a CSV greter than 25MB without having to do it manually via APIs

### Suggested solution

Have an environment variable to customize the maximum size of the CSV

### Additional context

There actually are (closed) existing issues, all of which were dismissed referencing to a variable name that does not do what it is supposed to:
- https://github.com/nocodb/nocodb/issues/10595
- https://github.com/nocodb/nocodb/issues/10596 (somehow points to the wrong environment variables, `NC_ATTACHMENT_FIELD_SIZE`, `NC_NON_ATTACHMENT_FIELD_SIZE`, which does not affect the maximum size of a CSV that needs to be imported)
- https://github.com/nocodb/nocodb/issues/11345 which says it's a reverse proxy issue, but in my case the reverse proxy allows up to 1GiB of data to be uploaded

I think this is where the check happens in the frontend:

https://github.com/nocodb/nocodb/blob/ea4b44f0d14a21d06a874a9cadaafa134d07cea9/packages/nc-gui/components/dlg/QuickImport.vue#L463

costajohnt:
## Summary

- Add `NC_IMPORT_FILE_SIZE` environment variable (in MB, default: 25) to configure the maximum file size allowed for CSV/JSON/Excel imports
- Replace the hardcoded 25MB limit in `QuickImport.vue` with the configurable value from the backend via the existing `appInfo` pipeline
- Error message now dynamically reflects the configured limit

## Details

The value flows through the existing env-var-to-frontend pattern used by `NC_ATTACHMENT_FIELD_SIZE`:

1. Backend constant in `constants/index.ts` with `Math.max(1, ...)` guard
2. Exposed via `appInfo` in `utils.service.ts`
3. Typed in `useGlobal/types.ts`, defaulted in `useGlobal/state.ts`
4. Consumed in `QuickImport.vue` with fallback to 25MB

**Note:** The unit for `NC_IMPORT_FILE_SIZE` is MB (e.g., `NC_IMPORT_FILE_SIZE=50` means 50 MB). This differs from `NC_ATTACHMENT_FIELD_SIZE` which uses bytes. MB was chosen because the import dialog comparison already operates in MB and it's more intuitive for operators. Happy to change to bytes if the team prefers consistency.

**Note:** Server-side enforcement of this limit is not yet implemented (the existing hardcoded check was also frontend-only). This could be a follow-up.

closes: #12572

## Test plan

- [x] Set `NC_IMPORT_FILE_SIZE=5`, upload a 6MB CSV → error: "less than 5MB"
- [x] Unset the env var (default 25MB), upload the same 6MB CSV → imports successfully
- [x] Set `NC_IMPORT_FILE_SIZE=0` or negative → clamps to 1MB minimum
- [x] Verify backward compatibility: no env var set behaves exactly like before (25MB limit)

### DIFF
--- a/packages/nc-gui/components/dlg/QuickImport.vue
+++ b/packages/nc-gui/components/dlg/QuickImport.vue
@@ -460,9 +460,10 @@ const beforeUpload = (file: UploadFile, fileList: UploadFile[]) => {
     showMaxFileLimitError.value = true
   }
 
-  const exceedLimit = file.size! / 1024 / 1024 > 25
+  const importFileSizeLimitMB = Math.max(1, appInfo.value.ncImportFileSize || 25)
+  const exceedLimit = file.size! / 1024 / 1024 > importFileSizeLimitMB
   if (exceedLimit) {
-    message.error(`File ${file.name} is too big. The accepted file size is less than 25MB.`)
+    message.error(`File ${file.name} is too big. The accepted file size is less than ${importFileSizeLimitMB}MB.`)
   }
   return !exceedLimit || Upload.LIST_IGNORE
 }
--- a/packages/nc-gui/composables/useGlobal/state.ts
+++ b/packages/nc-gui/composables/useGlobal/state.ts
@@ -140,6 +140,7 @@ export function useGlobalState(storageKey = 'nocodb-gui-v2'): State {
     type: 'nocodb',
     version: '0.0.0',
     ncAttachmentFieldSize: 20,
+    ncImportFileSize: 25,
     ncMaxAttachmentsAllowed: 10,
     isCloud: false,
     automationLogLevel: 'OFF',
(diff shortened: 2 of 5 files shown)


costajohnt:
NC_IMPORT_FILE_SIZE=5

Screenshot 2026-03-23 at 8 38 07 PM

Variable not set, defaults to 25

Screenshot 2026-03-23 at 8 42 01 PM

dindrick:
Just want to mention that implementing this PR would be super helpful for my org! We'd love to have the ability to import larger datasets.

DarkPhoenix2704:
We have already implemented in this last release. You can find the docs [here](https://nocodb.com/docs/self-hosting/environment-variables#storage)

PR #13383, opened 2026-03-27
fix text overflow inside table in api tokens page

### LINKED ISSUE #13382: 🐛 Bug: Text Overflow in API Tokens Page inside Token Table

### Steps to reproduce

1. Navigate to the Account > Tokens page
2. Resize the browser window to a width of <= 1000px.
3. "Creator" and "Token" columns: the email string and the token string overflow into each other.
Image

4. The rightmost "Action" column is pushed out of the viewport or becomes partially invisible.
Image

### Desired Behavior

- At widths < 1000px, the table should either wrap text or provide a horizontal scrollbar.
- The "Action" column must remain accessible at all standard desktop resolutions.

- Even if the user isn't on a mobile phone, they might be using a browser window that isn't maximized. So, a table should not break at 1000px

### Project Details

Base id: p4daavnwj5or3w5
Observed in: https://app.nocodb.com
Browser: Firefox.
Screen Resolution: at < 1000px width.

### Attachments

_No response_

kirank55:
## Change Summary
Fixes: https://github.com/nocodb/nocodb/issues/13382 Text Overflow in API Tokens Page inside Token Table

## Change type

Provide summary of changes.
1. Reduced padding to make sure the "Action" column remains in the viewport in all standard desktop resolutions.
2. Defined a responsive width to avoid the email string and the token string from overflowing into each other

https://github.com/user-attachments/assets/fa303504-9566-40fd-90d3-e496f1ae355f

### DIFF
--- a/packages/nc-gui/components/account/Token.vue
+++ b/packages/nc-gui/components/account/Token.vue
@@ -273,17 +273,17 @@ const handleCancel = () => {
           }"
         >
           <div class="h-full w-full overflow-y-auto rounded-md">
-            <div class="flex w-full pl-5 bg-nc-bg-gray-extralight border-1 rounded-t-md">
+            <div class="flex justify-between w-full pl-5 bg-nc-bg-gray-extralight border-1 rounded-t-md">
               <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-2/9" data-rec="true">{{
                 $t('title.tokenName')
               }}</span>
-              <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-2/9 text-start" data-rec="true">{{
+              <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-1/9 lg:w-2/9 text-start" data-rec="true">{{
                 $t('title.creator')
               }}</span>
               <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-3/9 text-start" data-rec="true">{{
                 $t('labels.token')
               }}</span>
-              <span class="py-3.5 pl-19 text-nc-content-gray-muted font-medium text-3.5 w-2/9 text-start" data-rec="true">{{
+              <span class="py-3.5 pl-5 lg:pl-19 text-nc-content-gray-muted font-medium text-3.5 w-2/9 text-start" data-rec="true">{{
                 $t('labels.actions')
               }}</span>
             </div>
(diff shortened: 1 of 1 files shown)


packages/nc-gui/components/account/Token.vue
rameshmane7218, line 1: @kirank55 Discard all your changes and replace all `GeneralTruncateText` with `NcTooltip` like below. No need to do any extra changes
```vue
<NcTooltip class="truncate" show-on-truncate-only>
  <template #title>
    {{ el.created_by }}
  </template>

  {{ el.created_by }}
</NcTooltip>
```

packages/nc-gui/components/account/Token.vue
```diff
@@ -273,17 +273,17 @@ const handleCancel = () => {
           }"
         >
           <div class="h-full w-full overflow-y-auto rounded-md">
-            <div class="flex w-full pl-5 bg-nc-bg-gray-extralight border-1 rounded-t-md">
+            <div class="flex justify-between w-full pl-5 bg-nc-bg-gray-extralight border-1 rounded-t-md">
               <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-2/9" data-rec="true">{{
                 $t('title.tokenName')
               }}</span>
-              <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-2/9 text-start" data-rec="true">{{
+              <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-1/9 lg:w-2/9 text-start" data-rec="true">{{
```
rameshmane7218: Add `pl-2`

packages/nc-gui/components/account/Token.vue
```diff
@@ -273,17 +273,17 @@ const handleCancel = () => {
...
                 $t('title.creator')
               }}</span>
               <span class="py-3.5 text-nc-content-gray-muted font-medium text-3.5 w-3/9 text-start" data-rec="true">{{
```
rameshmane7218: Add `pl-2`

packages/nc-gui/components/account/Token.vue
```diff
@@ -355,7 +355,7 @@ const handleCancel = () => {
                     </NcTooltip>
                   </div>
                 </span>
-                <span class="text-nc-content-gray-muted font-medium text-3.5 text-start w-2/9">
+                <span class="text-nc-content-gray-muted font-medium text-3.5 text-start lg:w-2/9 w-1/9 truncate">
```
rameshmane7218: Add `pl-2`

packages/nc-gui/components/account/Token.vue
rameshmane7218: Add `pl-2`

rameshmane7218 (changes requested):

packages/nc-gui/components/account/Token.vue
kirank55: Hi, thank you for getting back. 
I have discarded all changes and replaced GeneralTruncateText with NcTooltip. 
However, the "Action" in the table head was still going out of the viewport because of extra padding. So I have reduced padding to make sure the "Action" column remains in the viewport. @rameshmane7218

rameshmane7218 (approved):

packages/nc-gui/components/account/Token.vue
rameshmane7218: Okay

Issue #13578, opened 2026-04-20
🐛 Bug: AI Base Builder Fails on “Create Base”


Raniabenmoussa:
### Please confirm that the bug report does ***not*** already exist

- [x] I confirm there is no existing issue for this bug.

### Steps to reproduce

1. Go to Create New Base
2. Select Build with AI
3. Choose any use case (e.g., Inventory Management)
4. Click “Suggest Tables & Views”
5. Then click “Create Base” 
6. An error message appears:
“Something didn’t work as expected. Please try again.”
The base is not created.

### Desired Behavior

The system should successfully generate and create the base with the suggested tables and views, without errors.

### Project Details

NocoDB version: Cloud (app.nocodb.com)

### Attachments

<img width="1919" height="870" alt="Image" src="https://github.com/user-attachments/assets/b6a05a28-5daa-4313-9b47-c42b87cdf397" />

dstala:
transient issue looks like (or was fixed). it's working now. re-open this ticket if still facing issue.

PR #13587, opened 2026-04-21
fix(ai): use correct activeProjectId in callAiSchemaApi

rakshith1928:
## Change Summary
In `useNocoAi.ts`, the `callAiSchemaApi` function was reading `workspaceStore.activeProjectId` which does not exist on workspaceStore. `activeProjectId` is already correctly destructured from `basesStore` at the top of the composable. This caused `baseId` to be undefined in call sites like `generateTables`, `predictViews`, and `predictFilters`, making the API call fail silently.

## Change type

## Test / Verification
Exercise any flow that calls `callAiSchemaApi` with no explicit `customBaseId` (e.g. AI "Suggest Tables", "Predict Views", "Predict Filters" from within a base) — `baseId` should now resolve correctly from `basesStore.activeProjectId` and the API call should succeed instead of silently bailing.

## Additional information
The same pattern is used correctly in `callAiUtilsApi` which uses `activeProjectId.value` — this change brings `callAiSchemaApi` in line with it. `workspaceStore` exposes `activeWorkspaceId` (used correctly by `callAiSchemaCreateApi`) but has never had `activeProjectId`.

### DIFF
--- a/packages/nc-gui/composables/useNocoAi.ts
+++ b/packages/nc-gui/composables/useNocoAi.ts
@@ -71,7 +71,7 @@ export const useNocoAi = createSharedComposable(() => {
 
   const callAiSchemaApi = async (operation: string, input: any, customBaseId?: string, skipMsgToast = false) => {
     try {
-      const baseId = customBaseId || workspaceStore.activeProjectId.value
+      const baseId = customBaseId || activeProjectId.value
 
       if (!aiIntegrationAvailable.value || !baseId) {
         return


rameshmane7218 (commented):
The fix itself is correct — `workspaceStore.activeProjectId` doesn't exist and the destructured `activeProjectId` from `basesStore` is the right source. Good catch on the inconsistency with `callAiUtilsApi`.

**However, this PR does not actually fix #13578.**

The AI Base Builder's "Create Base" flow (reproduced in the linked issue) does not go through `callAiSchemaApi`. It calls `callAiSchemaCreateApi` → `$api.ai.schemaCreate(...)` (see `useNocoAi.ts:499` where `createSchema` is invoked, and `:430` for `predictSchema`). Both already use `workspaceStore.activeWorkspaceId`, not `activeProjectId`, so this code path is untouched by the diff.

The "Something went wrong" error in #13578 is coming from the backend `schemaCreate` endpoint failing — this is a server-side bug, not the frontend `baseId` resolution being patched here. #13578 should be investigated separately as a backend `ai.schemaCreate` failure.

The frontend fix here is still worth merging on its own — it prevents a latent bug in the other `callAiSchemaApi` call sites (`generateTables`, `predictViews`, `predictFilters`, etc.).

rameshmane7218 (approved):

rameshmane7218:
@rakshith1928 Merging this pr, you can create different pr to fix issue #13578

rakshith1928:
@rameshmane7218 Thanks for the detailed review! You're right I traced through the full flow and confirmed that callAiSchemaCreateApi already uses workspaceStore.activeWorkspaceId correctly, so my fix doesn't touch that path at all.
But I dug into the #13578 failure further and the POST /api/v2/ai/workspaces/:workspaceId/bases endpoint returning 500 is definitely server-side, the frontend is sending a valid payload with title, tables, relationships, and views after proper validation. The OSS endpoint  POST /api/v2/ai/workspaces/:workspaceId/bases is not actually there in the codebase maybe its a cloud side issue.

PR #13771, opened 2026-05-06
fix(gui): auto-link record when creating from Has Many "+" button (V2 Link)

### LINKED ISSUE #13770: 🐛 Bug: V2 Link — creating record from Has Many "+" button does not auto-link to parent
## Problem

After upgrading Link fields from V1 (FK-based) to V2 (junction table), clicking the "+" button in a Has Many cell to create a new child record **no longer auto-links** the child to the parent record. The child record is created successfully, but the relationship is not established.

This is a regression from V1 behavior, where the new record was automatically linked to the parent.

### Two symptoms

1. **Expanded form does not pre-fill the BT field.** In V1, when opening the expanded form from a HM cell's "+" button, the reverse Belongs To field was pre-filled with the parent record. In V2, this field is empty.

2. **Saving the record does not establish the link.** Even if the user manually fills the form and saves, the junction table row linking the child to the parent is not created.

### Root cause

**Symptom 1 (pre-fill):** \`dataUtils.ts\` \`populateInsertObject()\` (L113-130) handles BT FK pre-fill by finding \`fk_child_column_id\` in the child table's columns and assigning the parent's PK value. After V2 upgrade, the FK column no longer exists in \`meta.columns\`, so \`childCol\` is \`undefined\` and the assignment silently fails.

**Symptom 2 (link not created):** \`LinkedItems.vue\` \`onCreatedRecord()\` (L227-235) handles the post-creation flow. For **existing parent records** (\`isNew.value = false\`), it only closes the dialog (\`vModel.value = false\`) without calling \`link()\` or \`nestedAdd()\`. In V1, this worked because \`populateInsertObject\` had already written the FK value into the child record. In V2, with no FK column, no link is established at all.

### Source reference

**\`LinkedItems.vue\` L227-235:**
\`\`\`js
if (!isNewRecord.value) return

if (!isNew.value) {
  vModel.value = false   // ← only closes dialog, no link() call
} else {
  await addLTARRef(record, injectedColumn?.value as ColumnType)
  loadChildrenList(false, state.value)
}
\`\`\`

Note: the \`else\` branch (new parent record) correctly calls \`addLTARRef\`, but the \`if\` branch (existing parent record) does not.

**\`dataUtils.ts\` L113-130:**
\`\`\`js
const childCol = meta.columns!.find((c) => colOpt.fk_child_column_id === c.id)
// childCol is undefined after V2 upgrade — FK column was deleted
\`\`\`

### V1 vs V2 behavior

| Step | V1 (FK-based) | V2 (junction table) |
|------|--------------|---------------------|
| Open expanded form from HM "+" | BT field pre-filled with parent | BT field empty |
| Save new record | FK column written → link established | No FK column → no link |
| Junction table | N/A | Row should be inserted, but isn't |

### Webhook timing issue

There are two possible fix approaches, and they have different webhook implications:

**Approach A (minimal — PR #13771):** Call \`link(record)\` in \`onCreatedRecord()\` after the record is created. This is simple but has a **webhook timing problem**: the record creation webhook fires when the record is saved, **before** \`link()\` is called. If the webhook reads the link/relation field, it will be empty at trigger time. In V1, the FK was written as part of the INSERT, so webhooks always saw the relation.

**Approach B (ideal — backend fix):** Fix \`populateInsertObject()\` / \`nestedInsert()\` to handle V2 Link fields, so the junction table row is created **in the same transaction** as the record. This would maintain V1's webhook behavior where the relation exists at trigger time. However, this requires deeper backend changes.

### Suggested fix

**Short-term (Approach A):** In \`LinkedItems.vue\` \`onCreatedRecord()\`, explicitly call \`link(record)\`:

\`\`\`diff
  if (!isNew.value) {
+   await link(record)
    vModel.value = false
  }
\`\`\`

**Long-term (Approach B):** Fix the backend \`populateInsertObject\` and/or \`nestedInsert\` to detect V2 Link fields and create junction table entries as part of the record creation flow, preserving webhook timing guarantees.

### Related

- #13762 — V2 Link field filter only matches display value, no way to filter by record ID (same V1→V2 upgrade)

### Environment

- NocoDB EE 2026.04.5 (Docker)
- PostgreSQL 16.6
- All link fields upgraded from V1 to V2

androidfans:
## Summary

Fixes #13770

After V1→V2 Link field upgrade, creating a new record from a Has Many cell's "+" button no longer auto-links the child to the parent.

**Root cause:** \`LinkedItems.vue\` \`onCreatedRecord()\` closes the dialog without calling \`link()\` for existing parent records, relying on \`populateInsertObject()\` to write the FK value. After V2 upgrade, FK columns are deleted and replaced by junction tables, so this implicit link path no longer works.

**Fix:** Explicitly call \`link(record)\` (already available from \`useLTARStoreOrThrow()\`) for existing parent records. This inserts the junction table row via the \`nestedAdd\` API, matching the behavior of manually linking an existing record.

### Webhook timing caveat

This fix calls \`link()\` **after** the record is created, meaning the create webhook fires before the link is established. In V1, the FK was part of the INSERT statement, so webhooks always saw the relation at trigger time.

A more thorough fix would be to update \`populateInsertObject()\` / \`nestedInsert()\` in the backend to handle V2 Link fields and create the junction table entry as part of the record creation flow. This PR is a minimal frontend-only fix; the backend approach is noted in #13770 as a follow-up.

## Changes

- \`packages/nc-gui/components/virtual-cell/components/LinkedItems.vue\`: Add \`await link(record)\` in the existing-parent branch of \`onCreatedRecord()\`

### DIFF
--- a/packages/nc-gui/components/virtual-cell/components/LinkedItems.vue
+++ b/packages/nc-gui/components/virtual-cell/components/LinkedItems.vue
@@ -227,6 +227,7 @@ const onCreatedRecord = async (record: any) => {
   if (!isNewRecord.value) return
 
   if (!isNew.value) {
+    await link(record)
     vModel.value = false
   } else {
     await addLTARRef(record, injectedColumn?.value as ColumnType)


androidfans:
wrong fix

PR #13775, opened 2026-05-06
feat: add eq_id/neq_id/in_id/nin_id operators for Link field filtering by record ID

### LINKED ISSUE #13762: 🐛 Bug: V1→V2 upgrade breaks existing API filters  -- V2 Link field filter only matches display value, no way to filter by record ID
## Problem

After upgrading V1 (FK-based) link fields to V2 (junction table), the `where` API filter on Link/LTAR fields only matches against the related table's **display value** (primary field), with no option to filter by record **Id**.

This is a breaking change from V1, where we could filter by the FK column directly (e.g. `(author_id,eq,5)`). After V2 upgrade, the FK column is deleted, and the only filter syntax available is `(author,eq,John Smith)` which matches the display value.

**The V2 upgrade UI does not warn about this breaking change.** Users upgrading existing link fields have no indication that their API filters will stop working.

### Source reference

In \`packages/nocodb/src/db/field-handler/handlers/ltar/ltar.general.handler.ts\`, all three relation types (HM, BT, MM) hardcode:

\`\`\`js
fk_column_id: parentModel?.displayValue?.id,
\`\`\`

There is no code path to filter by the related record's Id.

### Why this is a problem

1. **Display values are not guaranteed unique — and NocoDB doesn't enforce uniqueness.** The primary display field has no unique constraint by design. Consider a \`Books\` table linked to an \`Authors\` table where the display value is \`name\`. If two authors share the same name (e.g. "John Smith"), \`(author,eq,John Smith)\` matches both — there is no way to target a specific author. This isn't a user configuration error; NocoDB intentionally allows non-unique display values, so the filter system must support disambiguating by Id.

2. **Breaking change from V1.** In V1, \`Books\` had a physical \`author_id\` FK column. API consumers could filter precisely: \`(author_id,eq,5)\`. After V2 upgrade, this column is deleted and all existing API integrations break silently.

3. **Workaround is extremely complex and still broken.** The only workaround we found:
   - Create a Formula field \`{Id}\` on every related table (because Lookup cannot reference the system \`Id\` field directly)
   - Create a Lookup field on every referencing table to pull this formula value
   - Change all API filter code to use the Lookup field instead

   This means **2 new fields per relation** across the entire schema. For a project with 10+ tables and 30+ link fields, this is unsustainable.

   **Even this workaround is incomplete:** Lookup fields do not support the \`in\` operator (returns HTTP 500), so batch queries like \`(author_id,in,1,2,3)\` must be rewritten as verbose \`(author_id,eq,1)~or(author_id,eq,2)~or(author_id,eq,3)\` chains. This is fragile, hits URL length limits with large lists, and is not a sustainable solution.

### Tested approaches that don't work

| Syntax | Result |
|--------|--------|
| \`(author,eq,5)\` | 0 rows — matches display value "name", not Id |
| \`(author,eq,{Id:5})\` | 0 rows |
| \`(author.Id,eq,5)\` | Error: "Column alias not found" |
| \`@("author", eq, 5)\` (v3 where syntax) | 0 rows |
| \`(author_lookup_id,in,1,2,3)\` (Lookup workaround) | HTTP 500 — \`in\` not supported on Lookup fields |

### Expected behavior

1. **V2 upgrade UI should warn** that existing FK-based API filters will break after upgrade.
2. Link field filters should support filtering by the related record's Id. Possible approaches:
   - \`(author.Id,eq,5)\` — dot notation to access related record fields
   - A dedicated operator like \`(author,eq_id,5)\`
   - At minimum, allow Lookup to reference the system \`Id\` field directly (eliminating the need for the Formula intermediary), and support \`in\` on Lookup fields

### Environment

- NocoDB EE 2026.04.5 (Docker)
- PostgreSQL 16.6
- All link fields upgraded from V1 to V2

androidfans:
## Summary

Fixes #13762

V2 Link fields can only be filtered by display value (\`(author,eq,John)\`), with no way to filter by record ID. This is a breaking change from V1 where FK columns allowed \`(author_id,eq,5)\`.

### New operators

| Operator | Syntax | Description |
|----------|--------|-------------|
| \`eq_id\` | \`(author,eq_id,5)\` | Equal by related record ID |
| \`neq_id\` | \`(author,neq_id,5)\` | Not equal by ID |
| \`in_id\` | \`(author,in_id,1,2,3)\` | In list of IDs |
| \`nin_id\` | \`(author,nin_id,1,2,3)\` | Not in list of IDs |

Existing operators (\`eq\`, \`neq\`, \`in\`, etc.) continue to match by display value — fully backward compatible.

### Implementation

The \`_id\` suffix is detected in the LTAR filter handler. When present:
1. \`comparison_op\` is stripped to the base operator (\`eq_id\` → \`eq\`)
2. \`fk_column_id\` uses the related table's \`primaryKey\` instead of \`displayValue\`
3. The rest of the filter pipeline works unchanged

### Changes (4 files, ~30 lines)

- \`query-filter-lexer.ts\`: register operators in COMPARISON_OPS
- \`handlerUtils.ts\`: add negated mappings (\`neq_id\`→\`eq_id\`, \`nin_id\`→\`in_id\`)
- \`ltar.general.handler.ts\`: detect \`_id\` suffix → use primaryKey → strip suffix before parseConditionV2
- \`links.general.handler.ts\`: route \`_id\` operators to LtarGeneralHandler (instead of RollupGeneralHandler)

### Test plan

1. \`(field,eq_id,5)\` → returns records linked to ID 5
2. \`(field,in_id,1,2,3)\` → returns records linked to IDs 1, 2, or 3
3. \`(field,neq_id,5)\` → excludes records linked to ID 5
4. \`(field,eq,DisplayName)\` → still works (backward compat)

### DIFF
--- a/packages/nocodb-sdk/src/lib/parser/queryFilter/query-filter-lexer.ts
+++ b/packages/nocodb-sdk/src/lib/parser/queryFilter/query-filter-lexer.ts
@@ -37,6 +37,10 @@ export const COMPARISON_OPS = <const>[
   'isWithin',
   'btw',
   'nbtw',
+  'eq_id',
+  'neq_id',
+  'in_id',
+  'nin_id',
 ];
 
 export const COMPARISON_OPS_ALIAS = <const>[
--- a/packages/nocodb/src/db/field-handler/handlers/links/links.general.handler.ts
+++ b/packages/nocodb/src/db/field-handler/handlers/links/links.general.handler.ts
@@ -14,7 +14,8 @@ export class LinksGeneralHandler extends GenericFieldHandler {
     options: FilterOptions,
   ) {
     // V2 MO/OO: single-record semantics — filter by display value (like BT)
-    if (isBtLikeV2Junction(column)) {
+    // _id operators: always route to LTAR handler for primary key filtering
+    if (isBtLikeV2Junction(column) || filter.comparison_op?.endsWith('_id')) {
       return new LtarGeneralHandler().filter(knex, filter, column, options);
     }
 
--- a/packages/nocodb/src/db/field-handler/handlers/ltar/ltar.general.handler.ts
+++ b/packages/nocodb/src/db/field-handler/handlers/ltar/ltar.general.handler.ts
@@ -51,6 +51,15 @@ export class LtarGeneralHandler extends GenericFieldHandler {
         : RelationTypes.HAS_MANY;
     }
 
+    // _id operators: filter by related record's primary key instead of display value
+    const isIdOp = filter.comparison_op?.endsWith('_id') ?? false;
+    const effectiveFilter = isIdOp
+      ? new Filter({
+          ...filter,
+          comparison_op: filter.comparison_op.replace(/_id$/, ''),
+        })
+      : filter;
+
     if (relationType === RelationTypes.HAS_MANY) {
       const childTableAlias = getAlias(aliasCount);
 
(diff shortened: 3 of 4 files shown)


androidfans:
in_id has bug

Issue #8763, opened 2024-06-15
🔦 Feature: Search within a VIEW (Global)


dstala:
### Please confirm if feature request does NOT exist already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

The objective of this feature is to provide users with the ability to perform a search within a specific view, enabling them to find records or fields quickly without the need to filter or manually browse through extensive lists of data.

### Suggested Solution

Option for the user to search all the fields instead of specifics from the toolbar search menu

### Additional Context

![Screenshot 2024-06-15 at 7 27 57 AM](https://github.com/nocodb/nocodb/assets/86527202/7025b326-a526-42f5-9f17-f185f02ea486)

emre1e:
#8489 same issue

dstala:
Sub-issue
[🔦 Feature: Search highlighting matched fields](https://github.com/nocodb/nocohub/issues/4428)

machiav3lli:
Is there any blocker for this feature? or maybe someone who can point out to the relevant code?

wizzard0:
Ugly workaround is to create a formula field concatenating every other field, then search using that field.

jettero-jet:
+1 on this. We recently migrated from Monday.com to self-hosted NocoDB (v0.301.2) for our real estate CRM with 500+ records across multiple tables. Global search across all fields was one of the most-used features in Monday.com. Currently having to set up individual column filters with "contains" is a significant productivity loss. This feature would be a game changer for CRM and project management use cases.

drewrox2009:
I have posted a pull request to fix this but it hasn't been reviewed yet. Just a bump where it may get seen

kuhwede:
same issue

Issue #5617, opened 2023-05-10
🔦 Feature: Simple search/query option for Shared/embed view 


compasssites:
### Please confirm if feature request does NOT exists already ?

- [X] I confirm there is no existing issue for this

### Describe the usecase for the feature

In a shared or embed view (wihtout login), it woudl be nice to have a easy search field (like an Omnisearch) to filter/query from the data easily. Its difficult to find a name if one has few hundred rows. We can use Filter dropdown , but thats not intuitive and common people (tech challenged) might not find it or make an effort. A simple search/query option would be nice. 

### Suggested Solution

A simple search/query option which searches from all the fields from that view would be nice.

springfielddatarecovery:
We would also like this

drewrox2009:
I have posted a pull request to fix this but it hasn't been reviewed yet. Just a bump where it may get seen

PR #13883, opened 2026-05-22
feat: implement All Columns search in toolbar

drewrox2009:
## Summary
- Adds "All Columns" search mode to the smartsheet toolbar
- When search field is set to "all", query matches across all searchable columns using OR logic
- Prevents Ctrl+F shortcut from interfering with the smart text panel search

## Changes
- `packages/nc-gui/composables/useSmartsheetStore.ts` — Build multi-column search query when `search.field === "all"`, joining valid column queries with `~or`
- `packages/nc-gui/components/smartsheet/toolbar/SearchData.vue` — Skip Ctrl+F handler when focus is inside the smart text panel

## References
- ref: #5617
- ref: #8763

### DIFF
--- a/packages/nc-gui/components/smartsheet/toolbar/SearchData.vue
+++ b/packages/nc-gui/components/smartsheet/toolbar/SearchData.vue
@@ -62,15 +62,16 @@ const isSearchResultVisible = computed(() => {
 })
 
 const columns = computed(() => {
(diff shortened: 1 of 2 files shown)


drewrox2009:
I have posted a pull request to fix this but it hasn't been reviewed yet. Just a bump where it may get seen

PR #13918, opened 2026-05-26
fix: preserve column order in csv export

### LINKED ISSUE #13768: 🐛 Bug: CSV export does not reflect the column order after rearranging columns in the UI

### Steps to reproduce

1. add new column
2.  change order
3. export to csv

### Desired Behavior

csv have the same order of column 

### Project Details

i used hosted version
https://app.nocodb.com/

### Attachments

_No response_

aakashgautam-git:
## Change Summary

Fix CSV export column ordering to match the visible UI column order after rearranging columns.

Previously, empty table CSV exports bypassed view ordering and used the raw database column order instead.

Also ensures export filters/sorts are preserved across recursive paginated export calls.

Closes #13768

## Change type

* [x] fix: (bug fix for the user, not a fix to a build script)

## Test/ Verification

Verified the following scenarios locally:

* Rearranged columns in UI are exported in the same order in CSV
* Empty table CSV exports now preserve UI column order
* Non-empty table exports continue working correctly
* Export filters/sorts persist correctly across pagination
* Excel and JSON export behavior remains unchanged

Root cause:

* Non-empty CSV exports already preserved order because Papaparse derived headers from reordered row objects
* Empty table exports followed a separate path using `getQueriedColumns()`, which bypassed view ordering when `fieldsSet?.size` existed

Fix:

* Removed the conditional branch bypassing view ordering
* Ensured view ordering is always respected when a view exists
* Forwarded `param` in recursive export pagination calls to preserve filters/sorts consistently


fendy3002 (changes requested):
Please add unit test

fendy3002:
@aakashgautam-git uh apologize that I forgot we do not do unit tests anymore. However I have tried your patch myself and it works, please rollback the test and jest config and I'll merge.

once again apologize that I forgot about it

fendy3002 (approved):

Issue #9024, opened 2024-07-18
🐛 Bug: impossible to submit new record form with defaut values


jbricout007:
### Please confirm that the bug report does ***not*** already exist

- [X] I confirm there is no existing issue for this bug.

### Steps to reproduce

I have a table with some mandatory fields, wich have default values such as zero (numeric fields)
I try to add a new record by clicking on "new record-form"
The button is not available : I have to change the default value to submit the record
The problem does not appear  with "new record-grid" function

### Desired Behavior

It normally has to be possible to sumbit the new form without changing default values to add a new record

### Project Details

Node: **v18.18.2**
Arch: **x64**
Platform: **linux**
Docker: **false**
RootDB: **pg**
PackageVersion: **0.204.0**

### Attachments

![Form_before_changing_mandatory_value_with_default_value_KO](https://github.com/user-attachments/assets/8dee5a4a-482b-4cc1-9ef6-4f132730a389)
![Form_after_changing_mandatory_value_with_default_value_OK](https://github.com/user-attachments/assets/5ab62542-5f4e-4a2e-9f19-95cac79437d9)

dstala:
@jbricout007 if i create a table from nocodb & set a field with default value - i am able to create new record by using form mode. are you connecting to external db in this case? if so, can you let us know if the field has any additional constraints set?

dstala:
also, can you upgrade to latest version of nocodb. easy for us to support that way

salim-b:
I *think* I observed the same issue in older NocoDB versions: In a table with a column having a `DEFAULT` value as well as a `NOT NULL` constraint set from an external PostgreSQL data source, a new row could only be added via form view when a value was explicitly set for that column (otherwise an error was displayed that the column cannot be NULL). Via grid view, it always worked.

At least in the latest stable release (v0.251.1), this issue doesn't seem to be present anymore (at least I can't reproduce it anymore). Hope that helps narrowing it down.

jbricout007:
> @jbricout007 if i create a table from nocodb & set a field with default value - i am able to create new record by using form mode. are you connecting to external db in this case? if so, can you let us know if the field has any additional constraints set?

hi @dstala : an upgrade will take place in a few times. Perhaps it will be corrected seeing the explanation of @salim-b 
Yes we are connected to an external database (postgresql) but the field doses not have anay constraint : we just parameter in nocodb "NOT null" field and a default value for that field (0)

dstala:
[...]

PR #13929, opened 2026-05-28
chore(deps): patch remaining Dependabot advisories via pnpm overrides

pranavxc:
## Summary

Follow-up to PR #13892. Targets the **77 Dependabot advisories** still open against `develop` — same approach: pin minimum patched versions via `pnpm.overrides` without bumping declared deps across manifests.

Patch/minor-only bumps; major-version jumps with peer/API risk are deferred.

### New direct-import / runtime overrides

| Tested | Package | Patched range | Behaviour risk | What to smoke-test |
|---|---|---|---|---|
| ✅ | `ajv` (8.20.0) | `<8.18.0 → ^8.18.0` | None — patch within 8.x | Any request hitting `validatePayload(...)` (most POST endpoints) |
| ✅ | `webpack` (5.107.x) | `<5.104.1 → ^5.104.1` | None — patch within 5.x | Build steps via `pnpm build` |
| ✅ | `diff` (8.0.4) | `>=6.0.0 <8.0.3 → ^8.0.3` | nc-gui declares `^7.0.0` → bumped to 8.x. `diffWords` return shape `[{value, added, removed}]` preserved through 6 → 7 → 8 | Doc revision / audit diff UI |
| ✅ | `picomatch` (4.0.4) | `>=4.0.0 <4.0.4 → ^4.0.4` | None — patch within 4.x | File watchers, glob matching |
| ✅ | `glob` (11.1.0) | `>=11.0.0 <11.1.0 → ^11.1.0` | None — minor within 11.x | Local file-storage attachment download, migrations |
| ✅ | `js-yaml` (4.1.1) | `>=4.0.0 <4.1.1 → ^4.1.1` | None — patch within 4.x | Markdown/docs with frontmatter |
| ✅ | `serialize-javascript` (7.0.x) | `<7.0.5 → ^7.0.5` | None — patch within 7.x | nc-gui build artifacts |
| ✅ | `unhead` (2.1.13+) | `<2.1.13 → ^2.1.13` | None — patch/minor within 2.x | Frontend page meta / SEO |

### New transitive overrides

| Tested | Package | Patched range | Smoke test scope |
|---|---|---|---|
| ✅ | `path-to-regexp` (0.1.13 **exact pin** for Express 4) | `<0.1.13 → 0.1.13` | Any HTTP request — exercised by all routes. **Exact pin** because `>=0.1.13` resolves to 8.4.2, which is not callable as default and breaks Express 4 |
| ✅ | `follow-redirects` (1.16.0) | `<=1.15.11 → ^1.16.0` | axios HTTP requests (webhook firing, integrations) |
| ✅ | `js-cookie` (3.0.7) | `>=3.0.0 <=3.0.5 → ~3.0.7` | Frontend cookies. Scoped to 3.0.x so the 2.2.1 consumer is left alone |
| ✅ | `flatted` (3.4.2) | `<3.4.2 → ^3.4.2` | `pnpm lint` (eslint cache) |
| ✅ | `tar-fs` (3.1.2) | `>=3.0.0 <3.1.1 → ^3.1.1` | sqlite3 build, nuxt streaming |
| ✅ | `@tootallnate/once` (3.0.1) | `>=3.0.0 <3.0.1 → ^3.0.1` | http-proxy / agent-based requests |
| ✅ | `uuid` (12.0.1) | `>=12.0.0 <12.0.1 → ^12.0.1` | Anywhere uuid v4/v7 is generated (sessions, IDs) |

## Skipped — would break behaviour

| Package | Reason |
|---|---|
| `nodemailer 8.x` | We run 7.0.11. Major jump risks SMTP behaviour |
| `nuxt 3.21.6` | Already deferred in previous audits |
| `@nestjs/core 11.x` | nocodb uses NestJS 10.4.x; bumping core alone breaks peers |
| `file-type 21.x` | We're on 20.4.1; major jump risks consumer APIs |
| `tmp 0.2.6` | `tmp@0.2.7` (latest patched) is in the 7-day pnpm quarantine. Lockfile keeps `0.2.5` for now; will be picked up automatically once 0.2.7 quarantine expires, or we add `tmp` to `minimumReleaseAgeExclude` for an emergency bypass |
| `nanotar` | Advisory has no `first_patched_version` — upstream needs to ship a fix |
| `lodash` 4.18.0 | Advisory's `first_patched_version: 4.18.0` does not exist (lodash latest is 4.17.21) — advisory data issue |
| `playwright` | Dev dep only; advisory targets the test runner, not runtime |

## Lessons from the parallel nocohub audit

A bare `<X → >=X` override is dangerous because pnpm resolves `>=X` to the LATEST matching version. If a different major exists past X, the override can silently over-reach into an incompatible major. We hit this in [nocohub PR #9024](https://github.com/nocodb/nocohub/pull/9024):

- `path-to-regexp@<0.1.13: ">=0.1.13"` → resolved 8.4.2 → broke Express 4 (`TypeError: pathRegexp is not a function`)
- `@xmldom/xmldom@<0.8.13: ">=0.8.13"` → resolved 0.9.10 → broke @node-saml (`TypeError: errorHandler is no longer supported`)

Both were fixed by replacing `>=` with `^` / `~` / exact pin to cap the upper bound. The same pattern is applied throughout this PR.

## Test plan

- [ ] CI: `pnpm install --frozen-lockfile` reproduces locked versions
- [ ] CI: backend typecheck passes
- [ ] CI: frontend build succeeds (nc-gui)
- [ ] CI: backend unit tests pass
- [ ] CI: playwright e2e passes
- [ ] Smoke: webhook firing (handlebars/axios/form-data paths)
- [ ] Smoke: attachment upload + CSV import (multer)
- [ ] Smoke: SAML/OIDC/Google sign-in (passport-saml/xmldom/jws)
- [ ] After merge: confirm advisory counts drop via `gh api repos/nocodb/nocodb/dependabot/alerts --state open`

## Local verification done

- `pnpm install` runs clean (existing peer-warning drift unchanged from `develop`)
- All new override targets smoke-tested at the API level (require + minimal API call)
- `path-to-regexp@0.1.13` matches `/a/:b` correctly (Express 4's expected callable default)


fendy3002 (approved):

package.json
```diff
@@ -103,8 +103,23 @@
...
+      "ajv@<8.18.0": "^8.18.0",
+      "webpack@<5.104.1": "^5.104.1",
+      "unhead@<2.1.13": "^2.1.13",
+      "serialize-javascript@<7.0.5": "^7.0.5",
+      "glob@>=11.0.0 <11.1.0": "^11.1.0",
+      "diff@>=6.0.0 <8.0.3": "^8.0.3",
+      "picomatch@>=4.0.0 <4.0.4": "^4.0.4",
+      "js-yaml@>=4.0.0 <4.1.1": "^4.1.1",
+      "js-cookie@>=3.0.0 <=3.0.5": "~3.0.7",
+      "follow-redirects@<=1.15.11": "^1.16.0",
```
mertmit, line 117: There is a duplicate entry above, see if intended (follow-redirects)

mertmit (approved):

PR #13950, opened 2026-05-31
fix: allow ordering linked records

### LINKED ISSUE #11485: 🔦 Feature: Allow ordering linked records in relationships (e.g., many-to-many)

### Use case

**Follow-up to [#9916]**

This feature request was originally submitted in #9916 but was closed as resolved due to a change #10543. However, that update does not address the core issue described here — the ability to define and persist the order of linked records in one-to-many and many-to-many relationships.

The original request remains valid and is pasted below without changes for clarity and continuity.

**Use case**
In many scenarios, maintaining a specific order for linked records is essential. Examples include:

- Social Media Planning: When planning Instagram carousel posts, images need to follow a particular sequence to maintain narrative flow or visual appeal.
- Project Management: Tasks linked to a project should be ordered by priority or workflow.
- Playlists or Galleries: Songs or images in a playlist or gallery must appear in a defined order for a meaningful user experience.

Currently, NocoDB lacks the ability to define and persist the sequence of linked records, making it cumbersome to handle workflows where order is essential.

### Suggested solution

The solution could involve two steps:

1. Adding an "Order" column: Since the intermediary table already exists for many-to-many relationships, it would be relatively straightforward to add a numeric Order column to store the sequence of linked records.
2. Drag-and-drop interface: To simplify user interactions, implement a drag-and-drop feature in the UI for managing the order of linked records. This order should be stored in the Order column and respected in views, exports, and automations. This approach mirrors a similar feature available in Airtable, which makes managing ordered linked records intuitive.

### Additional context

Replicating similar functionality available in tools like Airtable would also make NocoDB more appealing for users migrating from other platforms.

jurczykpawel:
## Summary
- add persistent `nc_order` ordering for many-to-many junction links, including lazy backfill for existing junction tables when links are added or reordered
- add a nested links reorder API endpoint and backend permission/validation path
- add drag handles in the linked-records popup so linked records can be reordered visually

Fixes #11485

## Testing
- `./node_modules/.bin/eslint src/db/BaseModelSqlv2/add-remove-links.ts src/db/BaseModelSqlv2/relation-data-fetcher.ts src/db/BaseModelSqlv2.ts src/services/data-table.service.ts src/controllers/data-table.controller.ts`
- `./node_modules/.bin/eslint components/virtual-cell/components/LinkedItems.vue composables/useLTARStore.ts`
- `git diff --check`
- `./node_modules/.bin/run-p 'build:*'` in `packages/nocodb-sdk`
- `./node_modules/.bin/tsc --noEmit --pretty false` in `packages/nocodb` currently fails on existing repo/environment type resolution issues such as `NcRequest` express fields and missing optional/local modules (`@noco-local-integrations/core`, `moment`, `nc-gui/utils/debug`, etc.)

### DIFF
--- a/packages/nc-gui/components/virtual-cell/components/LinkedItems.vue
+++ b/packages/nc-gui/components/virtual-cell/components/LinkedItems.vue
@@ -1,4 +1,5 @@
 <script lang="ts" setup>
+import Draggable from 'vuedraggable'
 import { type ColumnType, type LinkToAnotherRecordType, isDateOrDateTimeCol } from 'nocodb-sdk'
 import { PermissionEntity, PermissionKey, RelationTypes, isLinksOrLTAR } from 'nocodb-sdk'
 
@@ -89,6 +90,7 @@ const {
   fetchChildrenChunk,
   clearChildrenCache,
   resetChildrenCache,
+  reorderLinkedRecords,
 } = useLTARStoreOrThrow()
 
 const { withLoading } = useLoadingTrigger()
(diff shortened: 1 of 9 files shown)


jurczykpawel:
Closing — this PR was opened prematurely by an automated agent before review. The approach needs rework (runtime ALTER TABLE on the link path breaks transaction atomicity on MySQL, V1 has-many UI/backend mismatch, NULL-order cross-DB sorting, and no tests). Will reopen a revised version if/when ready. Apologies for the noise.

PR #14170, opened 2026-06-29
fix(nocodb): disambiguate paired link column on V2 upgrade by fk_index_name

### LINKED ISSUE #13781: 🐛 Bug: V2 Link upgrade swaps field IDs when multiple HM columns reference the same table
## Problem

When upgrading Link fields from V1 to V2, if a table has **multiple Has Many columns pointing to the same related table**, the upgrade process can swap the field identities — resulting in silent data corruption for all external integrations (webhooks, API consumers, automation workflows).

## Root cause

In \`columns.service.ts\` \`convertLinkToV2()\` (~L6616-6667), the paired BT column lookup uses:

\`\`\`typescript
if (
  opts.fk_parent_column_id === colOptions.fk_parent_column_id &&
  opts.fk_child_column_id === colOptions.fk_child_column_id &&
  opts.type === pairedRelType
)
\`\`\`

When multiple HM fields reference the same table, they share identical \`fk_parent_column_id\` (source table PK) and \`fk_child_column_id\` (related table PK). The lookup matches the **first** BT column it finds, not the correct one for that specific HM field. This causes the second HM field to pair with the wrong BT column, swapping their col_relations after upgrade.

## Reproduction

1. Create table A and table B
2. On table A, create **two** Has Many link fields both pointing to table B (e.g. \`inputs\` and \`outputs\`)
3. Note their field IDs
4. Upgrade both link fields to V2
5. Check the field IDs — they will be swapped or reassigned incorrectly

## Actual behavior

Example: a \`tasks\` table with three link fields to a \`documents\` table:

| Field | Relation | V1 ID | V2 ID |
|-------|----------|-------|-------|
| \`source_doc\` | BT | \`col_aaa\` | \`col_aaa\` (unchanged) |
| \`output_docs\` | HM | \`col_bbb\` | \`col_ddd\` (new ID, swapped) |
| \`extra_materials\` | HM | \`col_ccc\` | \`col_eee\` (new ID, swapped) |

After upgrade, records created via \`output_docs\` were being linked as \`source_doc\` and vice versa — the col_relations were crossed.

## Impact

- **Silent data corruption**: no error is thrown, data just goes to the wrong field
- **All external integrations break**: webhooks, API consumers referencing field IDs get wrong data
- **Irreversible**: V2 upgrade cannot be rolled back, so the damage is permanent unless manually fixed in the database

## Suggested fix

The paired column lookup needs an additional condition to distinguish between multiple HM fields with the same FK structure. Options:

1. Track which BT columns have already been paired during the upgrade loop and skip them
2. Use the column's own ID or the specific col_relation ID as part of the matching criteria
3. Match based on the original V1 FK column name (which is unique per relation even when pointing to the same table)

## Related

- #13762 — V2 filter only matches display value
- #13770 — V2 HM create doesn't auto-link

decoded-cipher:
Fixes #13781.

`convertLinkToV2` matched the counterpart link column only by `(fk_parent_column_id, fk_child_column_id, type)` and took the first hit. When two relations share that key the wrong counterpart is picked, swapping the links' identities on the irreversible V1→V2 upgrade.

Trigger: an external source with multiple FK constraints over the same column pair (e.g. a Postgres schema with two FKs on `child.pid → parent.id`), which yields two links sharing both column ids and differing only by `fk_index_name`. It does not reproduce on an internal base, where each HM link mints its own FK column.

Fixed by resolving the counterpart via a shared `pickPairedLtarColumn` helper that disambiguates by `fk_index_name`; behaviour is unchanged for unambiguous or legacy (missing `fk_index_name`) cases, so existing conversions don't regress.

### DIFF
--- a/packages/nocodb/src/helpers/ltarPairing.ts
+++ b/packages/nocodb/src/helpers/ltarPairing.ts
@@ -1,0 +1,84 @@
+/**
+ * Helpers for identifying the counterpart column of a Link-to-another-record
+ * (LTAR) relation while upgrading V1 links to V2.
+ *
+ * Background (nocodb/nocodb#13781):
+ * When a link is converted to V2 we have to locate the column on the related
+ * table that forms the other half of the relation (the BT for an HM, or the HM
+ * for a BT). Historically this match keyed only on
+ * `(fk_parent_column_id, fk_child_column_id, type)` and returned the first hit.
+ * Those keys are NOT unique when an external source has multiple foreign-key
+ * constraints over the same column pair: NocoDB then creates two links that
+ * share both column ids and differ only by `fk_index_name`, so the first match
+ * could pick the wrong counterpart and the two links' identities ended up
+ * swapped after the (irreversible) upgrade.
+ *
+ * A genuine HM/BT pair is always created with the same FK constraint name in
+ * `fk_index_name` (see `createHmAndBtColumn`), and distinct relations get
+ * distinct constraint names. So when the base match is ambiguous we
+ * disambiguate by `fk_index_name`.
+ */
+
+/** Minimal subset of LTAR relation metadata needed to pair columns. */
+export interface LtarPairColOptions {
+  fk_parent_column_id?: string;
+  fk_child_column_id?: string;
+  fk_index_name?: string;
+  type?: string;
+}
+
+export interface LtarPairCandidate<
+  TColumn = unknown,
+  TColOptions extends LtarPairColOptions = LtarPairColOptions,
+> {
+  column: TColumn;
+  colOptions: TColOptions;
+}
+
+/**
+ * Pick the column that is the true counterpart of the link being converted.
+ *
+ * Behaviour is intentionally unchanged when the base match yields 0 or 1
+ * candidate, or when the source column has no `fk_index_name` (legacy rows) — in
+ * those cases the first base match is returned exactly as before, so existing
+ * conversions cannot regress. Only the previously-broken ambiguous case
+ * (an external source with multiple FK constraints over the same column pair)
+ * now resolves deterministically.
+ */
+export function pickPairedLtarColumn<
+  TColumn,
+  TColOptions extends LtarPairColOptions,
+>(
+  candidates: LtarPairCandidate<TColumn, TColOptions>[],
+  source: LtarPairColOptions,
+  pairedRelType: string,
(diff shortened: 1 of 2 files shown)


pranavxc (commented):
Approach looks correct - disambiguating by `fk_index_name` is the issue's own suggested option 3, and the 0/1-candidate / missing-fk_index_name fallback keeps existing conversions safe (no regression).

I reproduced this to pin down the trigger:

  - Internal base: not reproducible. Two normal HM links to the same table get distinct `fk_child_column_id` (each mints its own FK column), so the base match is already unique - verified live, and true back to 2026.04.0. The issue's repro steps don't trigger it.
  - External source: reproduced live. Importing a Postgres schema with two FK constraints on the same `child.pid` → `parent.id` produces two links sharing `fk_parent_column_id` and `fk_child_column_id`, differing only by `fk_index_name`. Converting the second one, the old first-match picks the wrong counterpart (swap); the new `fk_index_name` match fixes it. ✅

Please correct following: 
1. Tighten the description - the real trigger is an **external source with multiple FK constraints over the same column pair**, not **"multiple HM links to the same table"** (that doesn't reproduce on an internal base). Worth correcting so it's  reproducible. Of if reproducible in internal, please include the proper steps to reproduce.

decoded-cipher:
@pranavxc - Thanks for reviewing. You're right — the trigger is an external source with multiple FK constraints over the same column pair, not internal HM links.

I've updated the PR description and the code comments to reflect that.

packages/nocodb/src/helpers/ltarPairing.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Pick the column that is the true counterpart of the link being converted.
+ *
+ * Behaviour is intentionally unchanged when the base match yields 0 or 1
+ * candidate, or when the source column has no `fk_index_name` (legacy rows) — in
+ * those cases the first base match is returned exactly as before, so existing
+ * conversions cannot regress. Only the previously-broken ambiguous case
+ * (multiple links to the same related table) now resolves deterministically.
```
pranavxc: Here also you can update comment with something like

```
(an external source with multiple FK constraints over the same column pair)
```

decoded-cipher:
@pranavxc - Updated 👍

pranavxc (approved):

PR #14171, opened 2026-06-29
fix(nc-gui): detect browser language on first load

goododeo-spec:
## Problem

On first load the UI always falls back to English even when the browser/system language is set to another language (e.g. Chinese).

The detection logic in `useGlobal/state.ts` matches the browser's preferred languages against `i18n.global.availableLocales`. Locale messages are loaded lazily (`plugins/a.i18n.ts` creates i18n with `locale: 'en'` and loads packs on demand), so at boot `availableLocales` typically contains only `['en']`. A browser set to `zh-CN` therefore never matches and the app always falls back to `en`.

## Fix

Match the browser's preferred languages against the full list of supported locales (the `Language` enum) plus the `LanguageAlias` map, instead of the lazily-populated `availableLocales`:

- alias match first, e.g. `zh-CN`/`zh_CN` → `zh-Hans`, `zh-TW`/`zh_TW` → `zh-Hant`
- then an exact (case-insensitive) locale match
- then a language-prefix match, narrowed by region code when there are multiple candidates

The first (highest priority) browser language that resolves to a supported locale wins. The existing startup flow then lazy-loads and applies that locale.

## Test plan

- [ ] Set the browser language to Chinese, clear `localStorage` (`nocodb-gui-v2`), load the app → UI should start in Chinese.
- [ ] Browser language English → UI starts in English (unchanged).
- [ ] A user who has explicitly selected a language is unaffected (stored preference still wins).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/nc-gui/composables/useGlobal/state.ts
+++ b/packages/nc-gui/composables/useGlobal/state.ts
@@ -14,10 +14,6 @@ export function useGlobalState(storageKey = 'nocodb-gui-v2'): State {
   /** reactive timestamp to check token expiry against */
   const timestamp = useTimestamp({ immediate: true, interval: 100 })
 
-  const {
-    vueApp: { i18n },
-  } = useNuxtApp()
-
   const router = useRouter()
 
   const isSharedBaseOrErdOrView = computed(() => isSharedBaseOrErdOrViewRoute(router.currentRoute.value))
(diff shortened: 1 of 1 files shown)


pranavxc:
Nice fix on the root cause (matching against the full `Language`/alias list instead of the lazily-loaded `availableLocales`). Two issues in the new detection logic though:

**1. English-first users with a secondary language get mis-detected.**

`'en'` is used as both the fallback sentinel and a valid match, so the "first match wins" short-circuit (`if (locale !== 'en') return locale`) never triggers for English. With `navigator.languages = ['en-US', 'zh-CN']`:

- `en-US` → `matchPreferredLocale` returns `'en'` (prefix match) → `locale` stays `'en'`
- because `locale === 'en'`, iteration continues → `zh-CN` → `zh-Hans` wins

So a user whose **top** preference is English (`en-US`/`en-GB`) but who also lists another language gets switched to that other language — a common setup (e.g. `['en-US', 'es']`). Suggest a distinct "no match" sentinel so the first *resolving* language (including `en`) wins:

```js
let resolved
for (const language of preferredLanguages.value) {
  const m = matchPreferredLocale(language)
  if (m) { resolved = m; break }
}
const preferredLanguage = resolved ?? 'en'
```

**2. Alias match is case-sensitive while the exact/prefix matches aren't.**

`const aliasKey = language.replace(/-/g, '_')` preserves case and `languageAliases[aliasKey]` is an exact keyed lookup, but the exact/prefix matches use `toLowerCase()`. Language tags are case-insensitive per spec, so a browser sending `zh-cn` → `zh_cn` misses the alias and falls through to the prefix path. Normalize case for the alias lookup too (lower-case both the key and the alias-map keys).

Worth adding a test for the English-first + secondary-language case (`['en-US', 'zh-CN']` → English), since that's the one that regresses.

PR #14204, opened 2026-07-03
fix(csv-import): prevent period in column name from breaking PostgreSQL insert

### LINKED ISSUE #14200: 🐛 Bug: import excel column name with dot fails

### Steps to reproduce

import of Column name "Ext. Beleg-Nr" fails 

Image

BTW - the heading should read "Import" not "Export"
postgres log: 
2026-07-03 05:29:22.044 UTC [8256] ERROR:  column "Ext" of relation "FO_buchungszeilen" does not exist at character 114

### Desired Behavior

should handle column names correctly and import must not fail

### Project Details

2026.06.2

### Attachments

table definition
[buchungszeilen-1783057815410.sql](https://github.com/user-attachments/files/29623291/buchungszeilen-1783057815410.sql)

test Excel
[20251106 Buchungszeilen 09-2025 copy.xlsx](https://github.com/user-attachments/files/29623238/20251106.Buchungszeilen.09-2025.copy.xlsx)

rakshith1928:
## Change Summary

Fixes #14200  by always replacing periods (.) in database column names during sanitization, even when column name sanitization is disabled via environment variables.

Previously, when NC_DATABASE_COLUMN_NAME_SANITIZE_ENABLED=false (or NC_SANITIZE_COLUMN_NAME=false), column names containing periods (e.g. Ext. Beleg-Nr) were returned unchanged. This caused Knex to interpret the period as an identifier qualifier (table.column), resulting in invalid SQL during import.

This change always replaces . with _ before the environment variable check, ensuring period-containing column names are safe for SQL generation while preserving the existing behavior for other characters when sanitization is disabled.

## Change type

## Changes

Fix 1-> (columnHelpers.ts): Always strip periods from column names, even when NC_DATABASE_COLUMN_NAME_SANITIZE_ENABLED=false. Periods are incompatible with Knex identifier handling.
Fix 2-> (data-import.processor.ts): Defense-in-depth  strip periods from dest column names in import data keys before passing to Knex.
Fix 3-> (pg.extractor.ts): Error display  change regex \w+ → [^"]+ so column names with special chars are fully captured in error messages ( as in issue Ext. Beleg-Nr instead of just Ext)

## Test/ Verification
Compiled clean with tsc --noEmit

Periods in database column names are incompatible with Knex's identifier handling, as . is interpreted as a qualifier separator. Replacing periods before the sanitization bypass prevents invalid SQL generation while keeping the existing opt-out behavior for other column name sanitization rules.

### DIFF
--- a/packages/nocodb/src/helpers/columnHelpers.ts
+++ b/packages/nocodb/src/helpers/columnHelpers.ts
@@ -635,6 +635,9 @@ export async function populateRollupForLTAR({
 }
 
 export const sanitizeColumnName = (name: string, sourceType?: DriverClient) => {
+  // Periods are always replaced — Knex treats '.' as a schema/table qualifier
+  // separator and produces broken SQL for any identifier containing a period.
+  name = name.replace(/\./g, '_');
   if (
     process.env.NC_DATABASE_COLUMN_NAME_SANITIZE_ENABLED === 'false' ||
     process.env.NC_SANITIZE_COLUMN_NAME === 'false'
--- a/packages/nocodb/src/helpers/db-error/pg.extractor.ts
+++ b/packages/nocodb/src/helpers/db-error/pg.extractor.ts
@@ -308,7 +308,7 @@ export class PgDBErrorExtractor implements IClientDbErrorExtractor {
           }
 
           const extractColumnNameMatch = error.message.match(
-            / column "(\w+)" does not exist/i,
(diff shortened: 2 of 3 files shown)


fendy3002:
The issue is with external db having field / table name with unsupported character. Sanitize is useless because it is the state of connected database.

Won't be supported for some period

PR #14272, opened 2026-07-13
feat: add opt-in Gateway API HTTPRoute support to the nocodb chart

somaz94:
## Change Summary

Adds an opt-in Gateway API HTTPRoute to the nocodb chart as an alternative to Ingress for clusters running a Gateway API implementation (Envoy Gateway, Istio, Cilium, etc.). Disabled by default (`httproute.enabled: false`), so existing installs are unaffected. It mirrors the ingress values shape: `parentRefs`, `hostnames`, `annotations`, and a `rules` escape hatch that overrides the default "/ -> service" rule when set.

## Change type

- [x] feat: opt-in Gateway API HTTPRoute template + values

## Test/ Verification

- `helm lint charts/nocodb`: passes
- `helm template` (default): no HTTPRoute rendered
- `helm template` (enabled with parentRefs + hostnames): valid HTTPRoute, default `/ -> service:8080` rule
- `helm template` (custom `rules`): override renders with filters

Chart version bumped 1.0.0 -> 1.1.0 (MINOR, backward-compatible opt-in feature).

The HTTPRoute reuses the chart's existing helpers (`nocodb.fullname`, `nocodb.labels`, `nocodb.annotations`, `nocodb.tplvalues.render`) and follows the same structure as `templates/ingress.yaml`.

### DIFF
--- a/charts/nocodb/Chart.yaml
+++ b/charts/nocodb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: nocodb
 description: NocoDB — A Free & Self-hostable Airtable Alternative. Production-ready chart for external Postgres and Redis.
 type: application
-version: 1.0.0
+version: 1.1.0
 appVersion: "2026.06.1"
 home: https://nocodb.com
 icon: https://github.com/nocodb/nocodb/raw/develop/packages/nc-gui/assets/img/icons/512x512.png
--- a/charts/nocodb/README.md
+++ b/charts/nocodb/README.md
@@ -210,6 +210,11 @@ Create one Secret and reference it from each `*.existingSecret`:
 | `ingress.annotations` | Ingress annotations (cert-manager cluster-issuer here) | `{}` |
 | `ingress.hosts` | Ingress hosts | See values.yaml |
 | `ingress.tls` | Ingress TLS entries | `[]` |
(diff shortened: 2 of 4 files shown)


somaz94:
@mertmit — gentle ping (cc @pranavxc @DarkPhoenix2704). This adds opt-in Gateway API HTTPRoute support to the nocodb chart.

Worth surfacing: no pre-merge CI in this repo can exercise this change. Every `pull_request` workflow is path-filtered to `packages/**` (`ci-cd.yml`, `jest-unit-test.yml`) or to lockfiles (`dependency-review.yml`), and the one workflow that actually lints and renders the chart — `helm-release.yml` — is `workflow_dispatch`-only. So the empty check list isn't a failure; nothing is watching `charts/**`.

Since there's no automated signal to point at, here's a local run against the repo's own CI values (helm v3.18.1):

```
$ helm lint charts/nocodb
1 chart(s) linted, 0 chart(s) failed

$ helm template t charts/nocodb -f charts/nocodb/ci/external-only-values.yaml
renders clean

$ helm template t charts/nocodb -f charts/nocodb/ci/external-only-values.yaml \
    --set httproute.enabled=true --set 'httproute.parentRefs[0].name=my-gw'
kind: HTTPRoute   # backendRefs -> t-nocodb:8080
```

Default is `httproute.enabled: false`, so stock values render nothing — existing installs are unaffected. Happy to add a `ci/httproute-values.yaml` case so it's covered whenever `helm-release.yml` is dispatched.

PR #14309, opened 2026-07-24
perf(nocodb): O(1) workspace-user lookup in user field validation

irontaek:
# Change Summary

Replaces four `baseUsers.find(...)` scans in the user field handler's `parseUserInput` with O(1) `Map` lookups.

# Why

When a User-type cell is written, `parseUserInput` validates each submitted user value against the workspace user list. For every value it scanned the whole list:

```ts
const u = baseUsers.find((u) => u.id === user.id);    // or u.email === ...
```

There are four such call sites (id/email × object-input/string-input branches). Each is O(workspaceUsers), run once per value in the cell, so validating a multi-user cell is O(values × workspaceUsers). On a bulk import into a user field — many rows, many values, a large workspace — this adds up.

# How

Index the workspace users by `id` and by `email` once, then look up in O(1):

```ts
const baseUserById = new Map<string, (typeof baseUsers)[number]>();
const baseUserByEmail = new Map<string, (typeof baseUsers)[number]>();
for (const bu of baseUsers) {
  if (bu.id != null && !baseUserById.has(bu.id)) baseUserById.set(bu.id, bu);
  if (bu.email != null && !baseUserByEmail.has(bu.email)) baseUserByEmail.set(bu.email, bu);
}
// ...
const u = baseUserById.get(user.id);      // was baseUsers.find(u => u.id === user.id)
const u = baseUserByEmail.get(user.email);
```

Behaviour is identical — `find` returned the first match, and the maps are built first-occurrence-wins to preserve that. Workspace ids/emails are unique, so it's a straight lookup swap. I couldn't run the full monorepo suite locally, but the change is local and type-preserving.

### DIFF
--- a/packages/nocodb/src/db/field-handler/handlers/user/user.general.handler.ts
+++ b/packages/nocodb/src/db/field-handler/handlers/user/user.general.handler.ts
@@ -142,6 +142,16 @@ export class UserGeneralHandler extends GenericFieldHandler {
       },
     );
 
+    // Index workspace users by id and email so each per-value lookup below is O(1) instead of a
+    // baseUsers.find scan — the validation runs once per user value in the cell, so the scan was
+    // O(values × workspaceUsers). First occurrence wins, matching find.
+    const baseUserById = new Map<string, (typeof baseUsers)[number]>();
+    const baseUserByEmail = new Map<string, (typeof baseUsers)[number]>();
+    for (const bu of baseUsers) {
+      if (bu.id != null && !baseUserById.has(bu.id)) baseUserById.set(bu.id, bu);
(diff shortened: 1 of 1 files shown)


irontaek:
Friendly ping — anything needed from my side to move this forward?

One note from a self-review pass across my open perf PRs: a reviewer on a similar patch elsewhere rightly pointed out that building an index per call can cost more than the scan it replaces when only one value gets looked up. I checked this patch against that pitfall: the two maps are built once per `validate` call in a single pass over `baseUsers` — the same order of work as one `find` scan — so the worst case (a single-user cell) stays within a small constant of the old cost, while multi-user cells drop from O(values × workspaceUsers) to O(values + workspaceUsers). `baseUsers` is fetched fresh each call, so there's no stable array identity that would allow caching the index across calls.

fendy3002 (approved):

PR #14345, opened 2026-08-04
fix(nc-gui): sync chatwoot support widget color scheme with app theme

### LINKED ISSUE #13148: 🐛 Bug: Chatwoot widget does not follow app dark/light theme toggle

### Steps to reproduce

1. Open nocodb
2. Toggle the app theme to dark.
3. The app mode changes, but the widget stays light.
4. Optional verification in browser console: 
window.__NUXT__.config.public.chatwoot.settings.darkMode // returns "light"
window.$chatwoot.setColorScheme("dark") // widget switches to dark mode

### Desired Behavior

When the application theme is toggled between light and dark modes, the Chatwoot support widget should automatically match the app theme.

### Project Details

NocoDb use as docker: true
Node.js DB version : v22.22.0 ( inside container)
Database used : SQLite3
Database URL :sqlite3 default internal DB
Project creation method: New Project
Database on which spreadsheet is created : N/A
Database version:3.x (default SQLite3)

### Attachments

_No response_

okxint:
## Summary

The chatwoot support widget was permanently stuck on light mode regardless of the user's theme preference.

**Root cause:** `nuxt.config.ts` initialises chatwoot with `darkMode: 'light'` and `useProvideChatwoot.ts` never calls `window.$chatwoot.setColorScheme()` when the theme changes. The `useTheme()` composable already exposes `isDark` and drives `applyTheme()` for every other UI element via `watch(isDark, applyTheme)`, but chatwoot was not wired in.

**Fix:** Destructure `isDark` from `useTheme()` inside `useProvideChatwoot` and watch it. When the theme toggles and the widget is initialised, call `window.$chatwoot.setColorScheme('dark' | 'light')`. The call is guarded by `chatwootReady` and a runtime `ncIsFunction` check so it's a no-op if the SDK hasn't loaded yet.

Fixes #13148

### DIFF
--- a/packages/nc-gui/composables/useProvideChatwoot.ts
+++ b/packages/nc-gui/composables/useProvideChatwoot.ts
@@ -1,6 +1,8 @@
 export const useProvideChatwoot = () => {
   const { setUser, setConversationCustomAttributes, setCustomAttributes } = useChatWoot()
 
+  const { isDark } = useTheme()
+
   const { $api } = useNuxtApp()
 
   const { user, appInfo } = useGlobal()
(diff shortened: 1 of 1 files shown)


fendy3002:
Please sign the CLA here https://github.com/nocodb/nocodb/pull/14345#issuecomment-5176506495

PR #14346, opened 2026-08-04
fix(nocodb): register pg type parser for INTERVAL to prevent [object Object]

### LINKED ISSUE #14306: 🐛 Bug:[self-hosted] Interval data type

### Steps to reproduce

 public TimeSpan? CustomDuration { get; set; } = null; 
 resolves to interval data type in postgres by defualt 

currently nocodb is missing that type or wrongly maps to single line text and the value in nocodb is [object Object]

### Desired Behavior

nocodb should display interval data type value like 03:00:00
hh.mm.ss

### Project Details

Node: **v24.14.0**
Arch: **x64**
Platform: **linux**
Docker: **true**
RootDB: **sqlite3**
PackageVersion: **2026.07.0**

### Attachments

_No response_

okxint:
## Summary

Postgres `interval` columns (e.g. a .NET `TimeSpan` property mapped by EF Core) displayed `[object Object]` in NocoDB instead of the human-readable duration string like `03:00:00`.

**Root cause:** `CustomKnex.ts` registers custom `pg` type parsers for DATE, TIMESTAMP, TIMESTAMPTZ, FLOAT8, and NUMERIC, but omits OID **1186** (`INTERVAL`). The `pg` driver's built-in interval parser returns a structured JS object `{years, months, days, hours, minutes, seconds, milliseconds}` which coerces to `[object Object]` when stored or displayed.

**Fix:** Register a pass-through type parser for OID 1186 (identical to the existing DATE parser) so the raw Postgres wire-format string is returned unchanged — e.g. `03:00:00` or `1 year 2 days 03:04:05`.

```ts
// before (no INTERVAL parser — pg returns a JS object)
const pgTypes = { FLOAT4: 700, FLOAT8: 701, DATE: 1082, ... };

// after
const pgTypes = { FLOAT4: 700, FLOAT8: 701, DATE: 1082, INTERVAL: 1186, ... };
types.setTypeParser(pgTypes.INTERVAL, (val) => val);
```

Fixes #14306

### DIFF
--- a/packages/nocodb/src/db/CustomKnex.ts
+++ b/packages/nocodb/src/db/CustomKnex.ts
@@ -14,6 +14,7 @@ const pgTypes = {
   FLOAT4: 700,
   FLOAT8: 701,
   DATE: 1082,
+  INTERVAL: 1186,
   TIMESTAMP: 1114,
   TIMESTAMPTZ: 1184,
   NUMERIC: 1700,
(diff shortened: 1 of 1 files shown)


fendy3002:
Please sign CLA here https://github.com/nocodb/nocodb/pull/14346#issuecomment-5176607912

PR #14354, opened 2026-08-06
fix: return 404 when deleting a non-existent API token

### LINKED ISSUE #14349: 🐛 Bug: DELETE /api/v1/tokens/{id} returns HTTP 500 instead of 404 for an unknown token

### Steps to reproduce

`DELETE /api/v1/tokens/{id}` returns **HTTP 500** instead of **404** when no token matches the given identifier. `ApiToken.get()` returns `undefined` and the result is dereferenced without a null check.

1. Authenticate as a super admin.
2. `DELETE /api/v1/tokens/some-id-that-does-not-exist`
3. Observe **HTTP 500**, body `{"msg":"Something didn't work as expected. Please try again. If the problem persists, contact support."}`

Server log:

```
TypeError: Cannot read properties of undefined (reading 'description')
    at <...>.apiTokenDelete
    at async <...>.apiTokenDelete
```

#### Where this comes from

`packages/nocodb/src/services/api-tokens.service.ts` (current `develop`):

```ts
async apiTokenDelete(param: { tokenId: string; user: User; req: NcRequest }) {
  const apiToken = await ApiToken.get(param.tokenId);
  if (
    !extractRolesObj(param.user.roles)[OrgUserRoles.SUPER_ADMIN] &&
    apiToken.fk_user_id !== param.user.id
  ) {
    NcError.notFound('Token not found');
  }

  this.appHooksService.emit(AppEvents.API_TOKEN_DELETE, {
    userId: param.user?.id,
    tokenId: apiToken.id,
    tokenTitle: apiToken.description,
    req: param.req,
  });
  ...
```

There is no `if (!apiToken) NcError.notFound(...)` guard after the lookup:

- For a **super admin**, the left operand of the `&&` short-circuits, so the ownership check never runs and `apiToken` is first dereferenced in the `appHooksService.emit(...)` payload — hence the `reading 'description'` TypeError.
- For a **non-super-admin**, `apiToken.fk_user_id` throws for the same reason.

Either way an unknown token id produces a 500. The `NcError.notFound('Token not found')` that the code already contains is only reachable when the token *does* exist but belongs to someone else.

#### How I hit it

This is easy to trigger accidentally now that tokens are show-once. `GET /api/v1/tokens` no longer returns the `token` value (only `token_prefix` / `token_hash`), so a client that used to hold raw token values and delete by them gets a 500 rather than a "no such token" signal. Passing the token's `id` works correctly.

### Desired Behavior

Return **404 "Token not found"** when no token matches the identifier, by guarding the lookup:

```ts
const apiToken = await ApiToken.get(param.tokenId);
if (!apiToken) {
  NcError.notFound('Token not found');
}
```

### Project Details

```
NocoDB used as docker : true
NocoDB version : 2026.07.0
Database used in NC_DB URL : sqlite3
Project was created by clicking : New Project
Database on which spreadsheet is created : sqlite3
OS on which NocoDB is running : linux (Docker, Fly.io, shared-cpu 2x / 1GB)
```

### Attachments

_No response_

bunnysayzz:
## Change Summary

Fixes #14349 — `DELETE /api/v1/tokens/{tokenId}` crashed with a **500 error** when the token did not exist, because the service dereferenced `apiToken.fk_user_id` without a null check.

Now a missing token returns **404 Not Found** (via `NcError.notFound`) instead of crashing, in both `OrgTokensService` (the reported path for the org-scoped endpoint) and `ApiTokensService` (the base-scope endpoint, which had the identical defect).

## Change type

## Test/ Verification

- Added regression tests to both `org-tokens.service.spec.ts` and `api-tokens.service.spec.ts`:
  - missing token → `notFound` (404), no `delete` call (admin and non-admin role paths)
  - existing token owned by the user → deletes successfully
- Verified locally: 7/7 tests pass across both spec files (run under Node 22 with `jest --testPathPattern`; the repo's default `testRegex` does not match `.service.spec.ts` files, same as the pre-existing specs).

The root cause: `ApiToken.get()` returns `undefined` for a missing token, and both services then read `apiToken.fk_user_id` unconditionally, throwing `TypeError` → 500. The guard mirrors the existing `NcError.notFound('Token not found')` convention already used in the ownership check below it.

### DIFF
--- a/packages/nocodb/src/services/api-tokens.service.spec.ts
+++ b/packages/nocodb/src/services/api-tokens.service.spec.ts
@@ -1,13 +1,59 @@
 import { Test } from '@nestjs/testing';
-import { ApiTokensService } from './api-tokens.service';
+import { ApiTokensService } from '~/services/api-tokens.service';
+import { AppHooksService } from '~/services/app-hooks/app-hooks.service';
 import type { TestingModule } from '@nestjs/testing';
+import { ApiToken } from '~/models';
+import { NcError } from '~/helpers/catchError';
+import { validatePayload } from '~/helpers';
+import { OrgUserRoles } from 'nocodb-sdk';
+
+jest.mock('~/models', () => ({
+  ApiToken: {
+    get: jest.fn(),
+    delete: jest.fn(),
+    insert: jest.fn(),
+    list: jest.fn(),
+    listForNonSsoUser: jest.fn(),
+    listWithCreatedBy: jest.fn(),
+    count: jest.fn(),
+  },
+}));
+
+jest.mock('~/helpers/catchError', () => ({
+  NcError: {
+    notFound: jest.fn(),
+    _: { notFound: jest.fn() },
+  },
+}));
+
+jest.mock('~/helpers', () => ({
+  validatePayload: jest.fn(),
+}));
+
+jest.mock('~/services/app-hooks/app-hooks.service', () => ({
+  AppHooksService: jest.fn().mockImplementation(() => ({
+    emit: jest.fn(),
+  })),
+}));
 
 describe('ApiTokensService', () => {
   let service: ApiTokensService;
+  const notFoundSpy = NcError.notFound as unknown as jest.Mock;
 
   beforeEach(async () => {
+    jest.clearAllMocks();
+    // The real NcError.notFound always throws (return type `never`).
+    notFoundSpy.mockImplementation(() => {
+      throw new Error('Token not found');
+    });
     const module: TestingModule = await Test.createTestingModule({
-      providers: [ApiTokensService],
+      providers: [
+        ApiTokensService,
+        {
+          provide: AppHooksService,
+          useValue: { emit: jest.fn() },
+        },
+      ],
     }).compile();
 
     service = module.get<ApiTokensService>(ApiTokensService);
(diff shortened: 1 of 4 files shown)


packages/nocodb/src/db/sql-client/lib/sqlite/SqliteClient.ts
```diff
@@ -1626,39 +1618,14 @@ class SqliteClient extends KnexClient {
...
-
       const fkCheckEnabled = (
         await this.sqlClient.raw('PRAGMA foreign_keys;')
       )?.[0]?.foreign_keys;
 
       if (fkCheckEnabled)
         await this.sqlClient.raw('PRAGMA foreign_keys = OFF;');
 
-      await this.sqlClient.raw('PRAGMA legacy_alter_table = ON;');
+      await this.sqlClient.raw('PRAGMA legacy_alter_table = OFF;');
```
pranavxc: This looks like unrelated change and might break column type change or column delete.

bunnysayzz:
Thanks for the catch @pranavxc — you're right, that SQLite change had nothing to do with this fix. It snuck in because I'd based this branch on top of an older branch of mine by mistake.

I've rebased onto the latest `develop`, so this PR is now just the token 404 fix and its tests (4 files, single commit). The `legacy_alter_table` change is already in `develop` via #14302, so nothing's lost — it was never supposed to ride along here.

Appreciate you flagging it 🙏

pranavxc (approved):

PR #14418, opened 2026-08-16
fix(currency): preserve minus sign when pasting negative values in en-US locale (Fixes #14413)

### LINKED ISSUE #14413: 🐛 Bug: Currency field silently drops the minus sign on paste (default en-US locale)

### Steps to reproduce

- Create a table with a Currency field named Amount. Leave every option at its default (Currency Locale en-US, Code USD, Precision 2).
- In a text editor or a spreadsheet outside NocoDB, select and copy the text -100.50.
- Click the Amount cell in the grid and press <kbd>Ctrl</kbd>+<kbd>V</kbd>.
- Read the cell back — in the grid, and via GET /api/v2/tables/{tableId}/records.

Actual: the cell stores 100.5 and renders $100.50. The sign is gone, with no error or warning. Also affected: -1,234.56 → 1234.56, -0.01 → 0.01, -$100.50 → 100.5. Not affected: typing -100.50 into the cell saves correctly (the editor is a native number input that bypasses the serializer), and a Currency field whose locale is anything other than en-US keeps the sign. That asymmetry is the tell.

### Desired Behavior

Pasting a negative amount into a Currency field should store the negative value, exactly as typing it does and exactly as a de-DE/en-GB Currency field already does.

### Project Details

```
NocoDB used as docker : false (built from source)
NocoDB version : develop @ 445f1043 (2026.08.0 line, packages/nocodb 0.301.3)
Database used in NC_DB URL : sqlite3 (default) — defect is client-side, dialect-independent
Project was created by clicking : New Project
OS on which NocoDB is running : macOS (Darwin 25.6.0)
Node.js version if running as node : v22.22.2
```

### Attachments

_No response_

waterWang:
## Summary

Fixes #14413

When pasting a negative value (e.g., `-100.50`) into a Currency field with the default `en-US` locale, the minus sign was silently dropped because the sanitizer regex `[^0-9.]` only kept digits and the decimal point, discarding the leading `-`.

## Root Cause

In `serializer.ts`, the `serializeCurrencyValue` function has a special fast path for `en-US` locale (line 285):

```typescript
// Before (buggy): strips minus sign
return value?.replace(/[^0-9.]/g, '')?.trim();

// After: keeps minus sign, matching the locale-aware path
return value?.replace(/[^\d.-]/g, '')?.trim();
```

The locale-aware path used by other locales (e.g., `de-DE`) already used `[^\d.-]` which correctly preserves the minus sign. This fix aligns the `en-US` fast path with the same logic.

### DIFF
--- a/packages/nocodb-sdk/src/lib/columnHelper/utils/serializer.ts
+++ b/packages/nocodb-sdk/src/lib/columnHelper/utils/serializer.ts
@@ -282,7 +282,7 @@ export const serializeCurrencyValue = (
         columnMeta.currency_locale === 'en-US' ||
         typeof (formatter as any).formatToParts !== 'function'
       ) {
-        return value?.replace(/[^0-9.]/g, '')?.trim();
+        return value?.replace(/[^\d.-]/g, '')?.trim();
       }
 
       const { group, decimal } = getGroupDecimalSymbolFromLocale(


dstala:
Please sign CLA before we can pick this up for review
https://github.com/nocodb/nocodb/pull/14418#issuecomment-5306963730

fendy3002:
to be fixed at https://github.com/nocodb/nocodb/pull/14431
sorry it's faster that way

PR #14419, opened 2026-08-16
fix(nocodb): retain all button visibility filters on save

### LINKED ISSUE #14337: 🐛 Bug: visibility conditions for button webhook are not adding

- [ ] I confirm there is no existing issue for this bug.

### Steps to reproduce

On grid view when I add a visibility condition for the webhook button it is working but if I add a second filter and click on save then only the second one is retained and active.
For my example I'm applying a filter on a single select and on a LTAR but looks like even 2 single selects are not working.

### Desired Behavior

Both filters should be displayed and active

### Project Details

Nocodb as docker with enterprise licence version 2026.07.0 table created by clicking on new projet with default DB

### Attachments

_No response_

MHJahanbakhsh:
## Summary
- Fixes undefined `incomingUidt` when rebuilding button colOptions on column update — the prior guard never defined the variable, so visibility filters were still wiped and only newly created filters survived save (Fixes #14337).
- Syncs button visibility filters via `filterRef` (same pattern as LTAR/Lookup/Rollup) and avoids overwriting in-progress draft filters on editor remount.
- Includes Button in keep-alive field editors so filter UI state is preserved across field switches in the table explorer.

## Test plan
- [ ] Edit a Button (webhook) column and add one visibility filter; save — filter persists
- [ ] Add a second visibility filter and save — **both** filters remain active
- [ ] Re-open the column editor — both filters are shown
- [ ] Convert the Button column to another type — visibility filters are cleaned up
- [ ] In Fields tab, switch away from and back to a Button field while editing filters — draft filters are retained

### DIFF
--- a/packages/nc-gui/components/smartsheet/column/ButtonOptions/index.vue
+++ b/packages/nc-gui/components/smartsheet/column/ButtonOptions/index.vue
@@ -374,12 +374,31 @@ const filtersCount = ref(0)
 
 if (isEdit.value) {
   const existingFilters = (vModel.value.colOptions as ButtonType)?.filters
-  if (Array.isArray(existingFilters) && existingFilters.length) {
-    vModel.value.filters = existingFilters.map((f: FilterType) => ({ ...f }))
+  // Prefer draft filters already on the form (e.g. restored from an unsaved
+  // edit) over colOptions so remounting the editor does not drop in-progress
+  // visibility conditions.
+  if (!Array.isArray(vModel.value.filters) || !vModel.value.filters.length) {
+    if (Array.isArray(existingFilters) && existingFilters.length) {
+      vModel.value.filters = existingFilters.map((f: FilterType) => ({ ...f }))
+    }
+  }
+  if (Array.isArray(vModel.value.filters) && vModel.value.filters.length) {
     isFilterSectionOpen.value = true
-    filtersCount.value = existingFilters.filter((f: FilterType) => !f.is_group && f.fk_column_id).length
+    filtersCount.value = vModel.value.filters.filter((f: FilterType) => !f.is_group && f.fk_column_id).length
   }
 }
+
+// Keep formState.filters in sync with the filter editor (same pattern as
+// LTAR/Lookup/Rollup). Relying only on v-model can lose entries when the
+// editor remounts or when deep filter mutations do not re-emit the array.
+watch(
+  () => filterRef.value?.filters,
+  (next) => {
+    if (!vModel.value) return
+    vModel.value.filters = next ? [...next] : []
+  },
+  { deep: true },
+)
 </script>
 
 <template>
--- a/packages/nc-gui/components/smartsheet/details/Fields.vue
+++ b/packages/nc-gui/components/smartsheet/details/Fields.vue
@@ -284,7 +284,13 @@ const changingField = ref(false)
 
 // Field types whose editors are kept alive (v-show) across field switches so filter
 // state inside SmartsheetToolbarColumnFilter is never destroyed.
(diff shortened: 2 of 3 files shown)


rameshmane7218 (commented):
Thanks for digging into this. Reviewed and tested locally against a live EE instance.
Summary: the `Fields.vue` change is a real fix and should land; the `ButtonOptions`
watch causes a new data-loss bug and should come out; the `Column.ts` hunk is needed
but diverges from the version this repo will receive on the next sync.

### 1. `ButtonOptions/index.vue` — please drop the `filterRef` watch

`ColumnFilter` already syncs `v-model` for consumers that bind it when `autoSave` is
false, and this editor does bind it (`v-model="vModel.filters"`). The comment there
notes that Rollup/Lookup/LTAR deliberately *omit* `v-model`, which is why those
editors need a parent-side watch — Button doesn't.

More importantly it loses data. The filter component sits behind
`v-if="isFilterSectionOpen"`, so collapsing the "Visibility condition" section
unmounts it, `filterRef` becomes null, and the `: []` branch clears the array:

- without this hunk: add a filter → collapse the section → Save → filter is saved
- with this hunk: same steps → the pending filter is silently discarded

In Rollup/LTAR the equivalent `v-if` is bound to a persisted toggle
(`meta.enableConditions`), so clearing on unmount is correct there. Here the guard is
a presentational disclosure chevron, so the same code becomes data loss.

The `isEdit` draft-preference change is fine and can stay.

### 2. `Column.ts` — the declaration is genuinely missing here, but please match upstream

Good catch that `incomingUidt` is undeclared on `develop` — any Button-column update
currently fails with `ReferenceError: incomingUidt is not defined`, and typecheck
reports TS2304. The declaration exists in the repo this one is synced from, along with
a guard your version omits; please use that form so the two don't diverge:

```ts
const incomingUidt =
  column.uidt || (column as { ui_data_type?: UITypes }).ui_data_type;

const requiredColAvail =
  !!incomingUidt &&
  (!requiredColumnsToRecreate[oldCol.uidt] ||
    requiredColumnsToRecreate[oldCol.uidt].every((k) => column[k]));
```

Two differences that matter: `ui_data_type` is the alias `insertColOption` dispatches
on, so falling back to `oldCol.uidt` instead desyncs the two; and without the
`!!incomingUidt` gate, a partial PATCH that omits `uidt` still deletes colOptions that
`insertColOption` can't rebuild (this is what drops dependent QR/Barcode columns).

### 3. `Fields.vue` — good, one formatting nit

Adding `UITypes.Button` to `KEEP_ALIVE_TYPES` fixes a real gap: an unsaved visibility
filter is discarded when you switch fields in the multi-field editor. Verified both
ways — with the entry the draft survives, without it the editor remounts and the draft
is gone.

The multi-line array fails `prettier/prettier`; it fits on one line:

```ts
const KEEP_ALIVE_TYPES = [UITypes.Links, UITypes.LinkToAnotherRecord, UITypes.Rollup, UITypes.Lookup, UITypes.Button]
```

### 4. Scope of #14337

The "filters wiped on save, only newly created survived" root cause was already fixed
on `develop` — `Column.update2` no longer deletes Button filters unless the column is
converting away from Button. Verified: adding a second condition and saving keeps both,
and reverting that guard locally reproduces the original symptom exactly. So test-plan
items 1–3 pass without this PR; the real contributions here are the undeclared
variable and the keep-alive gap. Worth updating the title/description.

Separately, while testing this flow we found two pre-existing bugs in the multi-field
editor that also eat Button visibility filters — a pending `update` op being dropped
when a newly added field is edited, and Reset not restoring conditions for keep-alive
editors. Both are out of scope here; we'll fix them separately.

rameshmane7218 (approved):
Approving to unblock — the `Fields.vue` keep-alive fix and the `Column.ts` declaration are both needed. Follow-up PR will remove the `filterRef` watch and align `Column.ts` with upstream, plus fix the two pre-existing multi-field-editor bugs noted above.

PR #14428, opened 2026-08-18
fix: prevent false insert-failure toast on successful row add (#13002)

### LINKED ISSUE #13002: Bug: 'Add row failed: Some internal error occurred' but row is created
## Description
When adding a new row via the UI (either creating new or duplicating), the error message 'Add row failed: Some internal error occurred' appears, but the row IS actually created in the database.

## Steps to Reproduce
1. Open a table in NocoDB UI
2. Click 'Add row' or duplicate an existing row
3. Error message appears: 'Add row failed: Some internal error occurred'
4. Check database - row was actually created

## Expected Behavior
Either:
- No error if row was created successfully
- Or actual error if row creation failed

## Environment
- NocoDB on Railway
- Backend: Supabase (PostgreSQL)

## Impact
Confusing UX - users think the operation failed when it succeeded.

wakqasahmed:
## Summary
- `insertRow` in `useInfiniteData.ts` wrapped all post-insert UI bookkeeping (cache-index shifting, row-color/button formula evaluation, aggregate reload) in the same `try`/`catch` as the actual `$api.dbViewRow.create` call. If any of that purely local bookkeeping threw after the API call had already succeeded, the shared `catch` reported it as an insert failure ("Failed to insert row: ...") even though the row had already been persisted server-side.
- Extracts the bookkeeping into a small `runPostInsertBookkeeping()` helper that runs it in its own try/catch and only logs a failure there instead of surfacing it as an insert error, so a genuine "Failed to insert row" toast can now only come from the insert API call itself.
- Adds `packages/nc-gui/vitest.config.ts` exposing the `~` alias Nuxt gives the app at runtime (mirrors `.nuxt/tsconfig.json`), so composables under `test/` can be imported directly by vitest without pulling in the whole Nuxt build pipeline.

Closes #13002

## Test plan
- Added `packages/nc-gui/test/insert-row-post-bookkeeping.test.ts` (vitest), covering:
  - bookkeeping runs normally on success
  - an error thrown mid-bookkeeping is swallowed/logged instead of propagating (the #13002 regression case)
  - side effects up to the throw point still run
- Verified the added tests fail without the fix (reverted `runPostInsertBookkeeping` to call the bookkeeping directly) and pass with it — 6/6 tests green including the existing `populate-insert-object.test.ts`.

### DIFF
--- a/packages/nc-gui/composables/useInfiniteData.ts
+++ b/packages/nc-gui/composables/useInfiniteData.ts
@@ -46,6 +46,22 @@ const isGroupByValueEqual = (a: unknown, b: unknown): boolean => {
   return String(a) === String(b)
 }
 
+/**
+ * Runs post-insert UI bookkeeping (cache writes, row-color/button formula
+ * evaluation, aggregate reload) once a row has already been persisted by the
+ * API. A failure here is never an insert failure — the row already exists in
+ * the database — so it must not surface as one (e.g. a "Failed to insert
+ * row" toast on a row that was in fact created successfully, #13002).
+ * Swallows and logs instead of throwing.
+ */
+export const runPostInsertBookkeeping = (bookkeeping: () => void) => {
+  try {
+    bookkeeping()
+  } catch (error) {
+    console.error('Row was inserted successfully but post-insert UI update failed', error)
+  }
+}
+
 const formatData = (
   list: Record<string, any>[],
   pageInfo?: PaginatedType,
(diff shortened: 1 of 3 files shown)


dstala:
Please sign here before we can pick this up for review
https://github.com/nocodb/nocodb/pull/14428#issuecomment-5322948596

wakqasahmed:
Hi @dstala — the CLA is signed now (the CLAassistant comment above shows "All committers have signed the CLA"). Ready for review whenever you get a chance. Thanks!

wakqasahmed:
Hi @rameshmane7218 @pranavxc — noticed this PR doesn't have a reviewer assigned yet — it's been about 4 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

wakqasahmed:
Hi @pranavxc @dstala — checking back in — no reviewer yet, about 7 days quiet. Happy to make any changes needed.

PR #14566, opened 2026-09-12
chore(nc-gui): untrack generated tour anchor catalog

rameshmane7218:
`tours/anchors.catalog.md` is a build artifact of `modules/tour-anchors.ts`. Nothing reads it back — the union that actually matters, `NcTourAnchorId`, is emitted to `.nuxt/types/nc-tour-anchors.d.ts` via `addTypeTemplate`, and `.nuxt` is already gitignored. The catalog is author-facing docs only.

Tracking it means constant churn: the module rewrites the file at setup and on every `.vue` watcher hit, and each entry records `file:line`. Any edit that shifts a line in an anchored component dirties the working tree and produces a diff with no real change in it.

The file stays on disk and regenerates on `pnpm dev`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rameshmane7218:
Folded into #14565 instead — that PR already touches this file.

